PR #76, opened 2014-08-28
Ability to detach disks prior to destroying a VM

pietern:


pietern:
Functionality available via a different path (detach explicitly prior to destroy).

PR #181, opened 2014-11-21
make storage resource manager

tkak:
I'd like to use StorageResourceManager with govmomi.
So, I made PR to support StorageResourceManager. I would be happy if you could merge.


dougm:
Looks great, thanks @tkak !

tkak:
Thanks!

PR #194, opened 2014-12-27
fix a issue of ignored https_proxy environment variable with https scheme

tkak:
`govmomi.NewClient` doesn't work when it access to vCenter via https proxy. It seems that `https_proxy` environment variable is ignored with RoundTripper.


pietern:
Thank you!

For future commits, please observe these guidelines for writing commit messages: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html.

tkak:
Thank you! I'll be careful for commit messages next time.

Issue #192, opened 2014-12-23
finder.HostSystem doesn't seem to be returning the correct host


gavingray:
I first noticed this when I was using the finder.HostSystem function for my project that's consuming govmomi.  I thought it was me, but now I'm not so sure about that given the following output.

    $ govc host.info -dc=858c7e1d-89f7-11e4-934e-80e6500a64dc -host.ip=192.168.238.139
    Name:            192.168.238.139
    Manufacturer:    VMware, Inc.
    Logical CPUs:    4 CPUs @ 2793MHz
    Processor type:  Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
    Memory:          4095MB
    Boot time:       2014-12-21 19:20:50.556233 +0000 UTC
    $ govc host.info -dc=858c7e1d-89f7-11e4-934e-80e6500a64dc -host=192.168.238.139
    Error: no such host

finder.HostSystem is actually not used all that much in the code.  In fact, after grepping a bit, the only place I found it was in flags/host_system.go.  I'll have to dig in a little more, but I suspect this is something I'll have to fix to address the AddStandaloneHost and Destroy host issue I filed yesterday.  In the event that I'm off base on this I'm happy to be corrected in order to save myself the time and trouble. :-D

gavingray:
PR submitted.  recurser.Recurse was matching the ComputeResource in the hostFolder, not the HostSystem child of the ComputeResource.

dougm:
Fixed in PR #195

PR #195, opened 2015-01-03
Issue #192: HostSystem doesn't seem to be returning the correct host.

gavingray:


dougm:
Great, thanks @gavingray !

PR #209, opened 2015-01-27
Update travis.yml

dougm:
- Update to Go 1.4

- Update go vet path

- Add Makefile to run locally and on travis


pietern:
Looks like govet needs to be silenced for the generated code.

LGTM otherwise, feel free to merge when green.

PR #210, opened 2015-01-27
Work on pool.* commands

pietern:
* Add usage & description to all commands
* `pool.create` against multiple pools
* `pool.change` against multiple pools


dougm:
Awesome, LGTM!

PR #211, opened 2015-01-27
Set insert_key = false

pietern:
Vagrant 1.7 expects to be able to replace the insecure key with a
freshly generated one. This doesn't work for an ESX guest yet.


dougm:
:+1:

PR #212, opened 2015-01-27
Key session file off of entire URL

pietern:
Before this change the same session file would be used if the schema was
changed. If a session was valid against an http URL, and then the schema
changed to https, the same session file would be used, which would still
try and use http.


dougm:
:+1:

PR #214, opened 2015-01-28
Adding EventManager so that events can be queried for

daniellockard:


pietern:
Thanks for the PR! @dougm and I have been chatting about separation of extra functionality into separate packages for a while, and maybe this one would be a prime candidate. In this case, all added functionality dealing with events (maybe something like event subscriptions, automatic gathering of auxiliary data, etc), would be added to a sub-package "event" (or similar), such that we can move towards a fairly minimal root package, with added functionality sprinkled on top in sub packages.

Can you share your specific use for the event manager and what you're building with it? If this is useful in general, it would be nice to be part of this package.

@dougm What do you think?

daniellockard:
Sure!  I'm currently trying to find a way to watch for CustomizationSucceeded / CustomizationFailed / CustomizationStarted events so that I can wait for for customization to be finished before I move on to the next step of my provisioning.  It's my understanding that I should be able to build up a filter that gives me back these sorts of events for a ManagedEntity (VM). I'm not sure exactly how to make it wait for the CustomizationSucceeded events yet but we'll get there.

The project I'm working on right now is automated instance creation using cloning and guest customization.

dougm:
I think helpers that wrap methods.*, such as this one that could be generated from the wsdl are ok to stay in the govmomi package.  But if we have a good package naming convention for them to live in their own packages, that'd be fine with me too.  So in this case, event/manager.go would work well, but what about types, such as the existing GuestProcessManager, guestprocess/manager.go ?  DistributedVirtualPortgroup, etc., are more challenging :)

According to https://golang.org/doc/effective_go.html#package-names:
"By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps."

Helpers that do more than thinly wrap methods.*, such as govmomi.VirtualDeviceList, I agree for sure should be in their own packages.

dougm:
@daniellockard can you rebase against master?  Let's go ahead and merge this and we will look at doing the package moves later.

daniellockard:
@dougm rebase'd

dougm:
Looks like you need to add a blank line to make goimports happy: https://travis-ci.org/vmware/govmomi/builds/52454644

daniellockard:
@dougm sorry about that, should be good now.

dougm:
@daniellockard no problem, sorry we took so long to merge!

PR #221, opened 2015-02-17
Add Login/Logout functions to client struct

pietern:


dougm:
LGTM.  Wonder though if these methods should be attached to a SessionManager wrapper instead?

pietern:
Talked about this, attaching these functions to a wrapper later, when we need another function on it.

PR #222, opened 2015-02-18
Add CreateFolder method

gavrie:
I was looking for a method to create a new folder inside a given folder, similar to `folder.CreateDatacenter` and `folder.CreateVM`. This commit adds a suitable `folder.CreateFolder` method.

### DIFF
--- a/folder.go
+++ b/folder.go
@@ -73,6 +73,20 @@ func (f Folder) CreateDatacenter(datacenter string) (*Datacenter, error) {
 	return nil, err
 }
 
+func (f Folder) CreateFolder(name string) (*Folder, error) {
+	req := types.CreateFolder{
+		This: f.Reference(),
+		Name: name,
+	}
+
+	res, err := methods.CreateFolder(f.c, &req)
+	if res != nil {
+		return NewFolder(f.c, res.Returnval), err
+	}
+
+	return nil, err
+}
+
 func (f Folder) CreateVM(config types.VirtualMachineConfigSpec, pool *ResourcePool, host *HostSystem) (*Task, error) {
 	req := types.CreateVM_Task{
 		This:   f.Reference(),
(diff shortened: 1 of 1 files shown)


pietern:
Hi @gavrie! Thanks for your contribution.

Can you please sign and email our CLA? After that we'll be able to merge this commit. Thanks.

See: https://github.com/vmware/govmomi/wiki

gavrie:
Hi Pieter,

I’ve signed and mailed the CLA.

Regards,

-- Gavrie

folder.go
```diff
@@ -73,6 +73,20 @@ func (f Folder) CreateDatacenter(datacenter string) (*Datacenter, error) {
...
 }
 
+func (f Folder) CreateFolder(name string) (*Folder, error) {
+	req := types.CreateFolder{
+		This: f.Reference(),
+		Name: name,
+	}
+
+	res, err := methods.CreateFolder(f.c, &req)
+	if res != nil {
```
dougm: Hi @gavrie , thanks for the patch and CLA!  Could you change this to:
```go
if err != nil {
   return err
}
```
And at the end of the func: return NewFolder(...), nil
We use this same pattern in the rest of the code, except Folder.CreateDatacenter (which we should fix too).

gavrie:
@dougm, I've updated the patch as you suggested. I initially followed the example of `CreateDatacenter` even though it seemed non-standard to me...

I haven't changed `CreateDatacenter` for now, since the comment in there suggests there is a specific reason for the nonstandard error handling there.

pietern:
@gavrie Great, thank you. The `CreateDatacenter` inconsistency must have fallen through the cracks. Fixing that in a separate PR. Merging your commit. Thanks again :+1:

gavrie:
Sure, glad to help :smile:

Issue #224, opened 2015-02-18
Proposal: Re-authenticate on NotAuthenticated error


hickeng:
Eventually an authenticated connection times out and any attempt to use it causes authentication errors.
Add a reauthenticate hook and have that called by any failing method so that an attempt can be made to reauthenticate before failing the operation prompting the error. This could use cached credentials or a long term session token.

pietern:
We now have keepalive functionality: 
https://github.com/vmware/govmomi/blob/v0.1.0/session/keep_alive.go

This wraps a soap.RoundTripper and waits for idle time to pass before sending a no-op request to keep the session alive. The idle timer starts when a login operation returns successfully and stops when a logout operation returns successfully.

In a similar fashion, a wrapper could be created that catches a NotAuthenticated error and (re-)authenticates the session. This is more tedious however, because a user with non-administrator privileges can receive a NotAuthenticated fault for some operations and not for others. Keepalive bypasses this issue.

lox:
Any guidance on how to actually use that round tripper @pietern?

PR #225, opened 2015-02-19
added SessionIsActive to Client

cgtexmex:
Simple addition to client implementing the SessionIsActive method

### DIFF
--- a/client.go
+++ b/client.go
@@ -34,6 +34,7 @@ var serviceInstance = types.ManagedObjectReference{
 type Client struct {
 	Client         *soap.Client
 	ServiceContent types.ServiceContent
+	Session        SessionManager
 }
 
 func getServiceContent(r soap.RoundTripper) (types.ServiceContent, error) {
@@ -60,10 +61,11 @@ func NewClient(u url.URL, insecure bool) (*Client, error) {
 		Client:         soapClient,
 		ServiceContent: serviceContent,
 	}
-
+	// automatically create a new SessionManager
+	c.Session = NewSessionManager(&c, *c.ServiceContent.SessionManager)
 	// Only login if the URL contains user information.
 	if u.User != nil {
-		err = c.Login(*u.User)
+		err = c.Session.Login(*u.User)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


dougm:
Thanks @cgtexmex , LGTM.  @pietern - is it time for a SessionManager wrapper or are you ok having these methods attached to the Client type?

pietern:
@dougm I agree. It should also implement an interface that can help us taking care of #224.

cgtexmex:
I viewed the client as the govmomi equivalent of the session manager -- I guess it more aligns to the ServiceInstance.  I'd be happy to help with the implementation of a SessionManager wrapper, etc...

dougm:
@cgtexmex great, have a look at the existing govmomi/*_manager.go wrappers.  Each has a corresponding govmomi.Client.*Manager() getter.  So in this case we'd end up with govmomi/session_manager.go and a Client.SessionManager() SessionManager() method.

We have been talking about moving these wrappers into their own packages.  So this could be govmomi/session/manager.go and govmomi.Client.SessionManager() would return type session.Manager - thoughts @pietern ?

pietern:
@dougm Sounds good. The end goal being that we have bits and pieces of reusable functionality that you can pick and choose from and only depend on the vim25 packages.

cgtexmex:
I am working this -- I'll handle it in two PRs -- first will be the basic session package creation with the Manager.  That will contain the SessionIsActive method.  Then I'll have the next PR address Issue #224.

pietern:
@cgtexmex Sounds good. How do you plan to address #224? One idea is through implementing the soap.RoundTripper on top of the session manager, such that it can intercept failed requests and reauthenticate if necessary. Another idea is to have it live on the side but asynchronously run ping requests to keep the session alive. I'd like some discussion around this before proceeding.

@dougm What do you think?

cgtexmex:
@pietern I've not given it a lot of thought as of yet...I'll be sure to ping you and @dougm when I've got a solid direction.

pietern:
@cgtexmex Thanks.

dougm:
If we have a goroutine on the side keeping the session alive, then we would defeat any attempt to Client.Transport.(*http.Transport).CloseIdleConnections() - unless we add a wrapper method that will stop the ping goroutine and then call CloseIdleConnections.  Otherwise, it seems reasonable and rvc takes a similar approach: https://github.com/vmware/rvc/blob/master/lib/rvc/modules/vim.rb#L126-L131

If we can re-authenticate as needed in 1 place within the RoundTripper impl and keep it simple, that would probably be better.

pietern:
Agree that one place to do this very predictibly is the better solution here.

We can make soap.RoundTripper our abstraction and have the session manager implement this interface. Then we'll have an "AuthenticatingRoundTripper" of sorts, that takes care of authentication as needed, plus an explicit authentication function that can be called by govmomi.NewClient.

cgtexmex:
I've create a session manager wrapper and moved the login / logout functionality to that wrapper.  I've also added a SessionManager type to the Client Struct.  This allows the Client to perform the login / logout.  Other SessionManager specific methods (SessionIsActive) would require the user to grab the Session from Client and execute the method.  

If you are happy with this then I'll look at the addressing #224 in a new PR

dougm:
@cgtexmex looks great, thanks!

PR #226, opened 2015-02-20
Change return pattern in CreateDatacenter

pietern:


pietern:
Fixes issue raised in #222.

dougm:
:+1:

PR #229, opened 2015-03-07
Loosen .ovf match in ova.import

dougm:
The .ovf file within a .ova archive may not have the same name.

For example:
ova: VMware-vRealize-Log-Insight-2.5.0-2475126.ova
ovf: VMware-vRealize-Log-Insight-2.5.0-2475126_OVF10.ovf

### DIFF
--- a/govc/importx/archive.go
+++ b/govc/importx/archive.go
@@ -55,7 +55,12 @@ func (t *TapeArchive) Open(name string) (io.ReadCloser, int64, error) {
 			break
 		}
 
-		if h.Name == name {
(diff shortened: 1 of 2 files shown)


pietern:
LGTM. Is it possible to have multiple OVFs in an OVA that match the basename?

dougm:
Not sure.  If we find out that's the case, then will probably scrap the TapeArchive abstraction and make it more .ova specific.

pietern:
OK, sounds good. If that's the case we probably end up interpreting the entire thing anyway.

PR #230, opened 2015-03-09
Add context.Context argument

pietern:
* Functions resulting in I/O now take a first context.Context argument (in `vim25/...`)
* Use `context.TODO()` wherever this argument is required but not yet added in

### DIFF
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -151,7 +151,7 @@ func (c *Client) RoundTrip(reqBody, resBody HasFault) error {
 	tstop := time.Now()
 
 	if d.enabled() {
-		d.logf("done in %6dms", tstop.Sub(tstart)/time.Millisecond)
+		d.logf("%6dms (%T)", tstop.Sub(tstart)/time.Millisecond, resBody)
 	}
 
 	if err != nil {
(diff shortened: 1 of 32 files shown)


vim25/soap/client.go
```diff
@@ -43,48 +41,43 @@ type HasFault interface {
...
 
-	u        url.URL
-	insecure bool
-	_        uint32 // for alignment: https://code.google.com/p/go/issues/detail?id=6404
-
-	cn  uint64         // Client counter
-	rn  uint64         // Request counter
-	log io.WriteCloser // Request log
+	u url.URL
+	k bool // Named after curl's -k flag
```
dougm: Still need a 32-bit field next to the bool for alignment.  Or, I think you can move the bool to the end of the struct (after all the other 64-bit fields / aligned structs).

dougm:
Very nice, LGTM.  Looks like ci / Makefile needs to be tweaked to install deps.

vim25/soap/client.go
pietern: I believe this was only needed for the atomic increments on the client counter and request counter fields (since these are 64 bit). These are now located in a separate debug container, which is aligned itself, so AFAICT this should be fine.

dougm: ah right, in that case :+1:

PR #231, opened 2015-03-10
Misc

pietern:
This PR is a bag of stuff I came across while moving functionality out of the root package into independent subpackages and moving all code to use a context.Context to make it easy to control client side timeouts on arbitrary sequences of requests. See: http://godoc.org/golang.org/x/net/context

Including:
* Move the event manager to a package (cc @daniellockard)
* Move the session manager to a package (cc @cgtexmex @hickeng)
* We were using `url.URL` as value instead of pointer value (unnatural if you look at the stdlib)
* Embed `*soap.Client` in root client
* Add keep alive `soap.RoundTripper` to `session` package (cc @hickeng)
* Move functionality to wait for a task to complete to `task` package
* Move managed object wrapper structs to `object` package
* Move property collector wrapper and functions to `property` package
* Wrapper functions in object package use context.Context
* Client in root package uses context.Context


dougm:
Awesome @pietern , looks great!  btw, feel free to remove '_gen' from the filenames, we can revisit that if we get a decent wrapper generator working.

pietern:
@mkuzmin As a heads up, once this PR is merged it will break downstream code.

Wiring context.Context into the entire library was prompted by both your WaitForIP timeout problems and some code I had written that needed timeout management. Now that everything is wired up, every in flight chain of requests can be cancelled outside-in at any time. Through context.Context it is trivial to add timeouts to sequences of requests.

PR #234, opened 2015-03-16
Change 'go install' to 'go get' if govmomi is not present

kars7e:
If one doesn't have govmomi package installed, 'go install' will not work, and 'go get' should be used instead.


pietern:
I had to call out the installation of prebuilt govc binaries in the README as well, so folded in the "go get" change in that commit as well. See #237.

Thanks for the heads up.

PR #235, opened 2015-03-18
Version prep

pietern:
This marks 0.1.0.

The build script produces a number of govc binaries.

Once this is merged I'm putting those up for download on the release page.


dougm:
:+1:

PR #244, opened 2015-04-02
HTTP errors

pietern:
The following prompted these commits:

1. Not specifying the /sdk path in the URL makes the server return a 400 Bad Request
2. The HTTP status code was not checked in the SOAP client and it instead expected every response to contain an XML body. This is not the case when you're not talking to the SDK to begin with, in which case ESX returns a 400 Bad Request. Since the client was expecting a body, but didn't receive one, the error returned was io.EOF (caused by trying to read from an empty body). To fix this, we now capture unexpected HTTP status codes and return an error, instead of trying to read the body. The only statuses I have seen returned from the SDK endpoint are 200 OK and 500 Internal Server Error.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -36,7 +36,7 @@ func TestNewClient(t *testing.T) {
 
 	c, err := NewClient(context.Background(), u, true)
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


dougm:
:+1:

PR #245, opened 2015-04-03
Added logic to create/read/update cluster

mevansam:
Added code to be able to find, create, read and update vSphere clusters. The code builds upon the existing ClusterComputeResource class which was a simple skeleton around the vSphere cluster definition.


mevansam:
I noticed that although Travis failed when I ran ```make check test``` against the main upstream repo it still failed.

```
$ make check test             
checking go imports...
/bin/sh: goimports: command not found
make: *** [goimports] Error 1
```

pietern:
Hi Mevan,

The builds fail because you don't use gofmt on your code. It is good practice to gofmt all your code so there is only one style to stick to. See https://golang.org/cmd/gofmt/.

A couple of comments:
* First commit ( 4f5fcc6 ) looks great
* Second commit ( ea14348 ) adds a very basic getter on a property of the managed object. I think this is a slippery slope, as you could define a getter for every property and it becomes rather unwieldy to maintain that. Also, it is very easy for a consumer of the API to get this field by using the Properties function themselves (allowing to grab whatever fields they would need).
* f6d3c04 -- we can't merge this: vim25/types/types.go is automatically generated from the vSphere WSDL and manual modifications mean we cannot generate it anymore. The documentation for ResourceAllocationInfo shows that the fields are really optional, so they can be omitted. Are you sure the error is caused by the zero values not being serialized? Zero values for these fields have actual meaning. Also see http://pubs.vmware.com/vsphere-55/index.jsp#com.vmware.wssdk.apiref.doc/vim.ResourceAllocationInfo.html .

Can you submit the CLA? See: https://github.com/vmware/govmomi/wiki

Thanks,
Pieter

pietern:
Regarding the bool field ExpandableReservation, see #247.

The other fields should be set to -1 if they are unlimited.

mevansam:
Hi Pieter

Thanks for the feedback. I will make the corrections and submit a new PR and email the CLA as well.

Regards
-Mevan

pietern:
@mevansam Can you rebase those commits on this PR and force push here? I rather have a single PR where the work is done than have one per iteration of the code. Thanks.

PR #250, opened 2015-04-18
Add code to ignore environmentBrowser value in MissingSet method

### LINKED ISSUE #242: Get "SystemError" at vim25.mo.LoadRetrievePropertiesResponse()
I can not get a suitable virtual machine list with finder.VirtualMachineList() in case that there are some VM templates at VMFolder. A result that o.MissingSet at objectContentToType() returns "environmentBrowser" value at VM template object may be the reason for the error.

https://github.com/vmware/govmomi/blob/master/vim25/mo/retrieve.go#L35

tkak:
`environmentBrowser` value is unset in VirtualMachine Object of VM templates. This means that `objectContentToType` always returns error as following xml data, when I try to find target VM template with `finder.VirtualMachine`. This fixes #242.

```
...
<missingSet>
  <path>environmentBrowser</path>
  <fault>
    <fault xsi:type="SystemError">
      <reason>unexpected error reading property</reason>
    </fault>
    <localizedMessage/>
  </fault>
</missingSet>
...
```

### DIFF
--- a/vim25/mo/retrieve.go
+++ b/vim25/mo/retrieve.go
@@ -33,7 +33,11 @@ import (
 func objectContentToType(o types.ObjectContent) (*reflect.Value, error) {
 	// Expect no properties in the missing set
 	for _, p := range o.MissingSet {
-		return nil, soap.WrapVimFault(p.Fault.Fault)
+		if p.Path == "environmentBrowser" {
+			continue
+		} else {
+			return nil, soap.WrapVimFault(p.Fault.Fault)
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


pietern:
@tkak Thanks for the PR. I'm writing a functional test to verify this issue exists when talking to vCenter and to be able to verify this fixes it. I don't like this fix though, as it also ignores `environmentBrowser` from the missing set when it is explicitly requested as a property on a non-VM managed object, in which case we _do_ want to return an error. However, maybe this is the only short term fix, in which case I'm fine with merging it for now.

pietern:
Fixed in #253.

PR #252, opened 2015-04-25
Fix error when using SDRS datastores

### LINKED ISSUE #227: support for StoragePod
currently, it seems that govc is unable to perform any actions on datastores within StoragePods (datastore clusters). For example:

    govc datastore.ls -ds='/MyDatacenter/datastore/MyStoragePod/MyDatastore'

returns with `Error: cannot traverse type StoragePod`

tkak:
This fixes #227.


pietern:
Great, thank you!

PR #253, opened 2015-04-27
Fix issue 242

### LINKED ISSUE #242: Get "SystemError" at vim25.mo.LoadRetrievePropertiesResponse()

pietern:
This fixes #242.

Thanks to @tkak to investigation and initial patch.

Instead of only merging the fix, I wanted to add an integration test as well so that we can catch regressions. The `functional` package is the place to add more integration tests as needed.


dougm:
LGTM.  Great to have some more functional tests/helpers that don't depend on govc/bats.

test/functional/helper.go
```diff
@@ -0,0 +1,130 @@
...
+		c:   test.NewAuthenticatedClient(t),
+		fns: make([]func(), 0),
+	}
+
+	h.f = find.NewFinder(h.c, true)
+
+	return h
+}
+
+func (h *Helper) Defer(fn func()) {
```
dougm: Don't see this used anywhere, do we need it and Teardown() vs built-in defer?

pietern: I added it so that resources that are added on-demand from the helper code can be cleaned up as well. Without a teardown function it is not possible to capture those in the scope of the test function. I didn't end up using this in the single test case but think it would be good to leave them in for future stuff. For example: if we want to create a resource pool on demand for every test, we should just do that from the helper, but still destroy it after every test run.

tkak:
@pietern Thank you for the modification!

pietern:
@tkak Thank you for figuring out the source of this bug :)

Issue #251, opened 2015-04-24
govc seems inconsistant


manningcarl:
I'm trying to get govc to deploy to vmware vsphere 6. I have had minimal success getting govc to connect correctly, at first i thought i had a syntax error in the command however the dump below shows the same command recieved success only once a few seconds apart. For nearly all of my commands i receive a NoPermission error and i cannot reproduce with any consistancy.

login as: root
root@10.21.10.114's password:
Last login: Fri Apr 24 11:24:29 2015 from 10.20.14.178
[root@localhost ~]# export GOVC_URL='cmanning\@vsphere.local:Carl\-Manning1@10.21.15.100'
[root@localhost ~]#  govc about -debug --u=cmanning\@vsphere\.local:Carl-Manning1@10.21.15.100 --k=true
Name:         VMware vCenter Server
Vendor:       VMware, Inc.
Version:      6.0.0
Build:        2559267
OS type:      linux-x64
API type:     VirtualCenter
API version:  6.0
Product ID:   vpx
UUID:         8f3f2c90-e66b-46dc-a132-601e1d809887
[root@localhost ~]#  govc about -debug --u=cmanning\@vsphere\.local:Carl-Manning1@10.21.15.100 --k=true
Error: NoPermission
[root@localhost ~]# export GOVC_URL='cmanning\@vsphere.local:Carl\-Manning1@10.21.15.100'
[root@localhost ~]#  govc about -debug --u=cmanning\@vsphere\.local:Carl-Manning1@10.21.15.100 --k=true
Error: NoPermission

pietern:
Can't reproduce with the exact same vCenter build. Running about many times in a row and don't see the NoPermission error. Do you use the embedded identity server or a separate SSO appliance?

```
$ govc about
Name:         VMware vCenter Server
Vendor:       VMware, Inc.
Version:      6.0.0
Build:        2559267
OS type:      linux-x64
API type:     VirtualCenter
API version:  6.0
Product ID:   vpx
UUID:         412b2a77-3ce7-486b-a323-3e3e0beff124
```

jamiemoore:
I seem to see the same issue with 5.5.  Only seems to work once for an account. Then you receive the Error: NoPermission.

Name:         VMware vCenter Server
Vendor:       VMware, Inc.
Version:      5.5.0
Build:        2183111
OS type:      linux-x64
API type:     VirtualCenter
API version:  5.5
Product ID:   vpx

manningcarl:
[...]

PR #256, opened 2015-05-05
Add flag to toggle persisting session to disk

pietern:
This removes the need for a shell script fix when restored sessions don't work (#251).


dougm:
:+1:

PR #257, opened 2015-05-06
Added logic for cluster and host management

mevansam:
Added logic to create/delete clusters and add hosts either as standalone or to a cluster.


pietern:
I folded the formatting fixes into the original commits and rebased to remove the merges in #265.

Thanks for your work!

Issue #259, opened 2015-05-19
how retrieve datastore summary ?


nlamirault:
hi
it is possible to retrieve the datastore summary ? Like this [1]
Thanks

[1] :  https://github.com/vmware/pyvmomi/blob/master/docs/vim/Datastore/Summary.rst

pietern:
Sure, you can get a handle to a datastore by going through the datacenter object and getting the children of the datastore folder. This is made a little easier by the `list` package.

Alternatively, you can use govc to get this information:

```
$ govc ls -json /ha-datacenter/datastore/datastore1 | jq '.elements[0].Object.Summary'
{
  "MaintenanceMode": "",
  "Type": "VMFS",
  "MultipleHostAccess": null,
  "Accessible": true,
  "DynamicType": "",
  "DynamicProperty": null,
  "Datastore": {
    "Value": "54f68da4-27c2b11a-9669-0022192dc408",
    "Type": "Datastore"
  },
  "Name": "datastore1",
  "Url": "/vmfs/volumes/54f68da4-27c2b11a-9669-0022192dc408",
  "Capacity": 137975824384,
  "FreeSpace": 136956608512,
  "Uncommitted": 0
}
```

nlamirault:
i try a function to list all datastores : 

```go
func (v *VSphere) Datastores(dc *object.Datacenter) ([]*object.Datastore, error) {
	finder := find.NewFinder(v.Connection.Client, true)
	finder.SetDatacenter(dc)
	dss, err := finder.DatastoreList(context.TODO(), "*")
	if err != nil {
		return nil, fmt.Errorf("Can't retrieve datastores: %v", err)
	}
	log.Debugf("vSphere datastores: %s", dss)
	return dss, nil
}
```
How can i use the "list" package to display the summary informations ?
Thanks

pietern:
See the referenced pull request. Sample output:

```
$ examples/datastores/datastores -url='https://username:password@host/sdk' -insecure
Name:       Type:  Capacity:  Free:
nfs         NFS    491.4GB    342.3GB
datastore1  VMFS   128.5GB    124.1GB
datastore2  VMFS   544.2GB    540.0GB
```

nlamirault:
Thanks

PR #260, opened 2015-05-20
Add example that lists datastores

pietern:
Example for #259.


dougm:
:+1: would be nice to have this in govc too

PR #262, opened 2015-05-21
Rename persist flag to persist-session

pietern:
It conflicted with the "persist" flag for the vm.disk.attach command.

Should have been caught by tests. It's time to backfill on tests for vm.disk.*.


dougm:
:+1:

PR #264, opened 2015-05-22
Change finder functions to no longer take varargs

pietern:
This means that these functions can safely return a not found error for patterns that didn't match any object. It is up to the caller to figure out what to do in case of such an error.

For the govc commands, many will continue executing for the remaining arguments, but exit with a non-zero status to indicate failure for one of the arguments. This is not the case right now, but with this change becomes easier to do so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@
 
 ...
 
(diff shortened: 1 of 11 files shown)


dougm:
:+1:

Issue #267, opened 2015-06-15
Retry requests that return a temporary error


pietern:
Optionally retry requests that return a temporary error. This can be embedded in the vim25.Client. Retry delay can be made configurable through a function that returns `(retry bool, delay time.Duration)`, so that multiple retry strategies can be mounted.

PR #268, opened 2015-06-15
Retry on temporary network errors

pietern:
This addresses #267.


dougm:
:+1: very nice!

pietern:
Retrying twice instead of once (as before).

PR #274, opened 2015-06-24
Modify archive.go bug

EricYT:
The h.Name is like ./foo.ovf, so foo*.ovf can not match the name.

### DIFF
--- a/govc/importx/archive.go
+++ b/govc/importx/archive.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"path"
 )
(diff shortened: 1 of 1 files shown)


pietern:
Thanks for the patch.

Could you please sign our [CLA](/vmware/govmomi/wiki) and email a signed copy?

@dougm Could you take a look here? You're more familiar with the OVF importing code than I am.

dougm:
LGTM

pietern:
Thanks for submitting the CLA, @EricYT.

Merging the patch.

PR #278, opened 2015-07-06
Credentials via environment variables

### LINKED ISSUE #273: govc : Can't use # in password
By default, vcsa requires one special character, so I use #

however, this cause the govc to fail to login:

. govc.env
./govc about
Error: Post https://administrator:Pass123/sdk#@192.168.1.28: dial tcp: unknown port tcp/Pass123

the password here is Pass123

export GOVC_URL="kikitux.net\administrator:Password123#@192.168.1.28"
export GOVC_INSECURE=1

pietern:
Support the `GOVC_USERNAME` and `GOVC_PASSWORD` environment variables.

This enables users to specify usernames and passwords with special characters that prohibit them from being specified as part of the `GOVC_URL` variable.

Fixes #273.

### DIFF
--- a/govc/flags/client.go
+++ b/govc/flags/client.go
@@ -37,6 +37,8 @@ import (
 
 const (
 	envURL           = "GOVC_URL"
+	envUsername      = "GOVC_USERNAME"
+	envPassword      = "GOVC_PASSWORD"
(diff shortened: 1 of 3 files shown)


dougm:
:+1:

PR #284, opened 2015-07-16
Add some useful methods to VirtualMachine and some information to govc output

gavrie:
The attached commits add a few useful methods to the VirtualMachine object:
- `String()` methods for pretty printing
- `Relocate` a VM to another host
- Retrieve the current host and resource pools for a VM

In addition, the `govc vm.info` output now shows the current host that a VM is running on.

### DIFF
--- a/govc/vm/info.go
+++ b/govc/vm/info.go
@@ -26,6 +26,7 @@ import (
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/property"
 	"github.com/vmware/govmomi/vim25/mo"
 	"golang.org/x/net/context"
@@ -102,13 +103,16 @@ func (cmd *info) Run(f *flag.FlagSet) error {
 				continue
 			}
 
-			host, err := vm.HostSystem(ctx)
-			if err != nil {
-				return err
-			}
-			hostName, err := host.Name(ctx)
-			if err != nil {
-				return err
+			var hostName string
+			hostRef := mvm.Summary.Runtime.Host
+			if hostRef == nil {
+				hostName = "<unavailable>"
+			} else {
+				host := object.NewHostSystem(c, *hostRef)
+				hostName, err = host.Name(ctx)
+				if err != nil {
+					return err
+				}
 			}
 
 			res.VmInfos = append(res.VmInfos, vmInfo{mvm, hostName})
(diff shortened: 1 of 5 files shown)


govc/vm/info.go
```diff
@@ -100,7 +102,16 @@ func (cmd *info) Run(f *flag.FlagSet) error {
 				continue
 			}
 
-			res.VirtualMachines = append(res.VirtualMachines, mvm)
+			host, err := vm.HostSystem(ctx)
```
pietern: This will do an extra call to find the host system through the summary property of the virtual machine managed object. I think it is possible to extract this from the properties that already retrieved in the `RetrieveOne` call. Then all that's needed is a `object.NewHostSystem` to get the corresponding object.

object/host_system.go
```diff
@@ -32,12 +32,31 @@ type HostSystem struct {
 	InventoryPath string
 }
 
+func (h HostSystem) String() string {
```
pietern: I rather not have helpers that look like plain getters but are actually doing network calls under the hood. This makes it possible for your application to stall without you having any control over it (as is the case when you use the `context.Context` struct where you can explicitly cancel those network calls if they take too long).

Since they are calling out to the `Name` function anyway, could you remove them (we can chat about this in the PR)?

pietern:
Thanks for your work @gavrie, looks great! I inlined some comments.

govc/vm/info.go
gavrie: Sure, I'll check if that's possible.

object/host_system.go
gavrie: OK, that makes sense. But I still think it's useful to be able to print an object and get something human-readable. How about caching the result of calling `Name()` and returning that as a result of `String()`, with the default being something like `<unknown>`?

pietern: While the name is human readable, it might be ambiguous if there is not enough context (i.e. multiple objects with the same name in different folders). The reference -- which is already available on these objects -- is an unambiguous reference to these objects. It's less human friendly, though still contains both the object type and an identifier. What do you think of using the reference instead? That way it is even possible to only implement `String()` once on the `Common` struct.

gavrie: My latest commit implements `String()` as you suggested, with the addition of also printing the `InventoryPath` if it is set. This does not require network traffic and adds useful data.

govc/vm/info.go
gavrie: Done.

gavrie:
@pietern, I've implemented your suggestions. Please let me know if you have any further comments on my PR.

pietern:
Thanks @gavrie, looks great! Could you change the `String()` implementation of the resource pool object as well? After that this one is good to merge.

gavrie:
Done. Shall I squash the commits before the merge, or do you want to do that?

pietern:
@gavrie Thanks for your work!. I squashed the changes into the relevant commits.

PR #286, opened 2015-07-16
Update to vim25/6.0 API

dougm:
- Regenerate from 6.0 wsdl

- Refactor govc pool commands now for the ResourceAllocationInfo
  interface


pietern:
Good stuff :+1:

dougm:
cc @hickeng

PR #291, opened 2015-07-27
Add support for clusters, hosts and VDS portgroups to govc ls -l

speijnik:
Signed-off-by: Stephan Peijnik <speijnik@anexia-it.com>

### DIFF
--- a/govc/ls/command.go
+++ b/govc/ls/command.go
@@ -108,6 +108,10 @@ func (l listResult) Write(w io.Writer) error {
 			if _, err = fmt.Fprintf(w, "%s (Network)\n", e.Path); err != nil {
 				return err
 			}
+		case mo.DistributedVirtualPortgroup:
+			if _, err = fmt.Fprintf(w, "%s (DistributedVirtualPortgroup)\n", e.Path); err != nil {
+				return err
+			}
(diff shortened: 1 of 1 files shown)


speijnik:
As a side note:

Long output of ls does not show objects for which no "type-case" is present at all.
Wasn't sure if that was intentional or not, but a "default:" case may not hurt there...

speijnik:
Just had another look, this could be further simplified, see https://github.com/anexia-it/govmomi/commit/0ca14af987db2791bb12e2a28546217e63881df6

The rationale behind that approach is that an object reference contains the object type anyways. This means that we can get rid of the switch/case altogether and just use the string from the reference.

pietern:
Thanks for your contribution. Can you please sign and submit our CLA?

Regarding the second commit: looks good. Could you change it to output the trailing / for folder types, as it currently behaves? Thanks!

caglar10ur:
I believe we can close this pull request as https://github.com/vmware/govmomi/commit/40c26fc6eb941aa6853530d4c6d1f2917a92b485 deprecates this.

pietern:
That sounds about right, thanks for noticing @caglar10ur.

PR #292, opened 2015-08-05
Make govet stop complaining

fdawg4l:
The govet Makefile target was complaining about the indentation and import order in the following files.  This change is to make it stop complaining.


pietern:
Thanks!

PR #293, opened 2015-08-05
Update travis.yml for new infra

dougm:
See: http://docs.travis-ci.com/user/migrating-from-legacy

Looks like our travis build hasn't run for ~2 months, not sure if this is why, but we should make this change in any case.


pietern:
:+1:

PR #297, opened 2015-08-06
Add OVF unmarshalling

pietern:
This models the properties we would need to extract from an OVF to parameterize OVF deployment.

Not yet hooked up to the import functions in govc.


dougm:
LGTM :+1:

PR #298, opened 2015-08-07
Provide ability to override url username and password

bruceadowns:
This is modeled after the govc behavior and is needed when the username and/or password contain special characters that makes it unable to specify them in an URL.


pietern:
Great, thank you!

PR #299, opened 2015-08-11
include VirtualApp in ls -l output

caglar10ur:


dougm:
Thanks!  btw @pietern couldn't we just use single printf with e.Object.Reference().Type instead of the switch there?

caglar10ur:
I assumed the type switch is there to whitelist known types otherwise I see no reason not to use e.Object.Reference().Type

pietern:
Agreed, we should do that. Still we would have an if for the folder type since that is displayed with a trailing / but other than that it should be good.

PR #300, opened 2015-08-11
Add CustomFieldsManager wrapper and cli commands

dougm:


govc/fields/rm.go
```diff
@@ -0,0 +1,69 @@
...
+func init() {
+	cli.Register("fields.rm", &rm{})
+}
+
+func (cmd *rm) Register(f *flag.FlagSet) {}
+
+func (cmd *rm) Process() error { return nil }
+
+func (cmd *rm) Usage() string {
+	return "NAME"
```
pietern: KEY...

pietern: No, it would be `NAME...`. I'm confused between the keys and names. It seems FindKey will find the integer key by the string name of that key, but that the integer key is not used anywhere else in the govc code.

govc/fields/rename.go
```diff
@@ -0,0 +1,69 @@
...
+func init() {
+	cli.Register("fields.rename", &rename{})
+}
+
+func (cmd *rename) Register(f *flag.FlagSet) {}
+
+func (cmd *rename) Process() error { return nil }
+
+func (cmd *rename) Usage() string {
+	return "KEY NAME"
```
pietern: Should this be `NAME NEW_NAME` or something? The key argument is actually the name, not the integer key, following the code below (going into FindKey with a string argument).

govc/fields/set.go
```diff
@@ -0,0 +1,95 @@
...
+func init() {
+	cli.Register("fields.set", &set{})
+}
+
+func (cmd *set) Register(f *flag.FlagSet) {}
+
+func (cmd *set) Process() error { return nil }
+
+func (cmd *set) Usage() string {
+	return "KEY VALUE PATH..."
```
pietern: If the changes above are also done then this should also be `NAME` instead of `KEY`.

pietern:
There are a couple of naming nits around `NAME` vs `KEY`. I don't know what the right solution is here, since there is a bunch of ambiguity. What about calling all key _names_ `key` and all key _IDs_ `id`?

dougm:
Currently using the same param names ('key', 'name', 'value') as the API:
```
func (m CustomFieldsManager) Add(..., name string, ...)
func (m CustomFieldsManager) Remove(.., key int)
func (m CustomFieldsManager) Rename(.., key int, name string)
func (m CustomFieldsManager) Set(..., key int, value string)
```

The cli supports using the int key or key by-name, the FindKey helper falls back to:
```
	k, err := strconv.Atoi(key)
	if err == nil {
		// assume literal int key
		return k, nil
	}
```

We could also just remove the key by-name helper and always use int key.. the user can lookup the int key using:
```
 govc fields.ls | grep key-name | awk '{print $1}'
```

Thoughts?

pietern:
OK sounds good, I'm +1 on sticking to KEY where appropriate and resolving it as int or string.

PR #309, opened 2015-08-18
Added Name method to datacenter.go

Viper23:
Datacenter is missing the Name() function that is present in other wrappers such as VirtualMachine and Datastore.  The inconsistency can be confusing, and getting the datacenter name is a common enough operation to deserve its own method.

### DIFF
--- a/object/datacenter.go
+++ b/object/datacenter.go
@@ -41,6 +41,17 @@ func NewDatacenter(c *vim25.Client, ref types.ManagedObjectReference) *Datacente
 	}
 }
 
+func (d Datacenter) Name(ctx context.Context) (string, error) {
+	var dc mo.Datacenter
+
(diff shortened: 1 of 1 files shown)


dougm:
@Viper23 thanks for the patch and apologies for the delayed response, we are working on a better system to track issues and PRs.

In this case, with commit b34f346 I think we can collapse the various Name() methods to the object.Common type.  Something like the following (untested):

```go
func (c Common) Name(ctx context.Context) (string, error) {
	var o mo.ManagedEntity

	err := c.Properties(ctx, c.Reference(), []string{"name"}, &o)
	if err != nil {
		return "", err
	}

	return o.Name, nil
}
```

dougm:
Taking care of this in #521

PR #315, opened 2015-09-01
Misc govc changes

dougm:
Add -hints option to host.esxcli, enables other frontends to parse the simple output format and render table data in their own way

Reduce number of round trips for vm.info property collection


pietern:
Looks good!

@gavrie This reverts your commit and renames the root property in the JSON output back to "VirtualMachines". This wasn't tested in your change, so I assume you didn't use the JSON output for this command? Checking to make sure this doesn't break anything on your end.

PR #316, opened 2015-09-02
Misc clean up

bruceadowns:
* change enumeration to be more idiomatic to golang


pietern:
Hah! I don't know if it's more readable, but oh well...

dougm:
Would be great to move Humanize() from examples to a helper, we can use it in datastore.ls -l at least and likely elsewhere too.

bruceadowns:
> Would be great to move Humanize() from examples to a helper

Agreed. @dougm Where would you put it?

PR #317, opened 2015-09-02
Added common method of humanizing byte strings

bruceadowns:
* added units package
* added ByteSize type as float64
* added String() receiver for ByteSize
* modified datastore.ls and datastore example to use new method


dougm:
Looks good, can we fold govc/vm/disk/byte_value.go into the units package too?

dougm:
Thanks @bruceadowns LGTM.  I'll let @pietern review/merge, since he wrote the code you are re/moving.

units/size.go
```diff
@@ -0,0 +1,101 @@
...
+		return errors.New("invalid byte value")
+	}
+
+	f, err := strconv.ParseFloat(m[1], 64)
+	if err != nil {
+		return err
+	}
+	*b = ByteSize(f)
+
+	//switch strings.ToUpper(m[2]) {
```
pietern: Comment can be deleted, or you can remove the lower case cases below.

units/size.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"regexp"
+	"strconv"
+)
+
+type ByteSize float64
```
pietern: It was an int64 in the `vm/disk` implementation. Can you use that here as well? I don't want to lose any precision if it is passed as an option argument when creating a disk.

units/size.go
bruceadowns: Excellent eyes! Yes, I forgot to remove the commented out line. I will fix.

units/size.go
bruceadowns: Initially I was confused by this comment because the user may actually be more precise.

i.e.
`govc vm.disk.create -size=10.5g ...`
vs
`govc vm.disk.create -size=10g ...`

But I believe you are referring the floating point arithmetic of int64(size.0 / 1024.0) possibly losing precision. Correct? If so, then I agree it makes sense to use int64.

pietern: Yep, you're right. You can still parse the first case as a float and then do the math until you arrive at an integer or have to round off (in case you get fractional bytes...). As long as the backing type is an integer.

bruceadowns:
Whoops, forgot about my unit tests. Luckily travis caught it!

dougm:
Thanks @bruceadowns !

PR #318, opened 2015-09-03
Add helper method to check if VMware Tools is running in the guest OS.

l1jiang:


object/virtual_machine.go
```diff
@@ -353,6 +353,18 @@ func (v VirtualMachine) Answer(ctx context.Context, id, answer string) error {
 	return nil
 }
 
+// ToolsRunningStatus returns the current running status of VMware Tools in the guest OS.
```
pietern: Name in comment doesn't match the name of the function.

pietern:
One minor comment, looks good otherwise.

pietern:
Cool, thank you!

PR #319, opened 2015-09-04
Add helper method to wait for virtual machine power state to be off

l1jiang:


pietern:
Thanks for the PR.

Can you change this to `WaitForPowerState` and make the power state to wait for an argument? This way the function is also usable to wait out of band for power on, suspend, etc.

l1jiang:
Renamed method to `WaitForPowerState` and added argument for power state.

pietern:
Perfect, thanks.

PR #321, opened 2015-09-08
Add govc vm.info resources option

dougm:
New '-r' option outputs the same data as the vSphere Client's
"Resources" section.

Make the "General" data optional via '-g' option, but enabled by
default.


pietern:
Cool! LGTM.

PR #323, opened 2015-09-09
Clean up of vcsa creation script

bruceadowns:
* added check for greadlink and gmktemp
* added note to install coreutils on Darwin


dougm:
Nice, thanks @bruceadowns

Issue #190, opened 2014-12-18
finder.VirtualMachine and VC


hickeng:
When switching from a single ESX host to VC, the finder.VirtualMachine() function doesn't always find existing VMs. I'm currently using SearchIndex().FindChild() as a work around with "ha-datacenter/vm/" as the root folder.

When connected to VC, the finder mechanism seems to only be able to find VMs that appear under "Discovered virtual machine" which is rarely a comprehensive list.

mkuzmin:
any progress on this?
Or maybe there is some other method to find a VM by name without knowing its full path?

dougm:
The path given to finder.VirtualMachine is relative to the datacenter vmFolder.  You can use "folderName/vmName" or "*/vmName".  But we should probably add an option to recurse folders so we can find by name without a folder path.

hickeng:
I've just been using FindChildByName in the interim. Not quite as concise and you need to process the reference returned,  but viable.
Please excuse me if I've got the function name slightly wrong, phone isn't good at browsing source.

George Hicken - mobile
ghicken@vmware.com


-------- Original message --------
From: Doug MacEachern
Date:20/01/2015 09:37 (GMT-08:00)
To: vmware/govmomi
Cc: George Hicken
Subject: Re: [govmomi] finder.VirtualMachine and VC (#190)


The path given to finder.VirtualMachine is relative to the datacenter vmFolder. You can use "folderName/vmName" or "*/vmName". But we should probably add an option to recurse folders so we can find by name without a folder path.

-

dougm:
I had looked into that and tested it out, but doesn't look like it recurses into sub folders:

http://pubs.vmware.com/vsphere-55/topic/com.vmware.wssdk.apiref.doc/vim.SearchIndex.html#findChild

"This only searches the immediate children of a managed entity"

mkuzmin:
VM name is an unique property, and enumerating all the folders on the a client side doesn't look as effective solution. `SearchIndex` doesn't provide such lookup methods, but there is alternate API. Here is how I currently do that in Groovy:
`def vm = new InventoryNavigator(client.getRootFolder()).searchManagedEntity("VirtualMachine", "name")`
It would be great to have the same API here.

dougm:
Agreed, I'll try to look into this over the weekend or early next week.

mkuzmin:
Any update on this?

aashishparikh:
Hey guys.. Any updates?

PR #328, opened 2015-09-10
Update finder.go / func VirtualMachineList

sszgr:
In the functions of VirtualMachineList, e.Object.(type) may be mo.Folder, and exist VirtualMachine under the Folder.
Add function findVirtualMachineListForFolder, which is change to foreach all Folder. But I didn't get VirtualMachine.InventoryPath.
Sorry for my bad code, hope someone could help me to improve it.

### DIFF
--- a/find/finder.go
+++ b/find/finder.go
@@ -534,6 +534,28 @@ func (f *Finder) DefaultResourcePool(ctx context.Context) (*object.ResourcePool,
 	return rp, nil
 }
 
+//foreach all Folder
+func (f *Finder) findVirtualMachineListForFolder(ctx context.Context, fobj object.Reference, vms *[]*object.VirtualMachine) {
+	if string(fobj.Reference().Type) != "Folder" {
+		return
+	}
+	folder := object.NewFolder(f.client, fobj.Reference())
+	childs, err := folder.Children(ctx)
+	if err != nil {
+		return
+	}
+	for _, chie := range childs {
+		switch chie.Reference().Type {
+		case "VirtualMachine":
+			vm := object.NewVirtualMachine(f.client, chie.Reference())
+			//vm.InventoryPath?
+			*vms = append(*vms, vm)
+		case "Folder":
+			f.findVirtualMachineListForFolder(ctx, chie, vms)
+		}
+
+	}
+}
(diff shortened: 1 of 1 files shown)


pietern:
Hi @sszgr, thank you for your contribution. If I read your code correctly, you're trying to recurse into a folder and list the virtual machines there if you see a folder? I.e. you want to walk the entire tree and return all VMs in all folders?

The function currently lets you specify the path you want to get a listing for. This means that if there is a folder you want to recurse into, you can specify so in the path. If you want to recurse into every folder, you can specify the path using a wildcard, like so: `/mydc/vms/*`. This will expand to every entry in this folder, including the folders, and will list the contents for every single entry (being VMs and folders).

For building something that recurses I would rather have an interface like this: http://godoc.org/path/filepath#Walk and implement it for any managed object rather than just VMs.

By the way, I see you opened and closed a number of pull requests. Please update an existing pull request instead of creating a new one for every change. Also see: https://help.github.com/articles/changing-a-commit-message/

sszgr:
Hi @pietern, thank you for your reply!
I found that ```f.find(ctx, f.vmFolder, false, "/myDC/vm/*")``` had just list the folders under current folder without recursion. and ```f.find(ctx, f.vmFolder, true, "/myDC/vm/*")``` just list more folders under the child folders which are under current folder without recursion.
As well as in ```func(f*Finder) VirtualMachineList()``` had also not traversal folder, it just case ```mo.VirtualMachine```.
Maybe I get the wrong comprehension, I thought that ```func VirtualMachineList``` should list all VirtualMachine under the path by using recursion. I would feel pity if it can just only list current VirtualMachine  without traversal, ignore this.
By the way, I will follow your suggest, read git help carefully.
Thank you.

dougm:
Issue #190

sszgr:
Thank you very much.

austinlparker:
@sszgr if you're still around, I could really use this PR. if you're no longer interested in it I'll resubmit your change.

sszgr:
sorry for reply  your email so late , thank you very much.

sszgr:
@vmwclabot thanks.

avnish30jn:
@pietern `/mydc/vms/*` this doesn't recurse into the folder.

PR #330, opened 2015-09-11
add not start thread option

runner-mei:
I'm writing a monitoring program, do not want to start the thread while execute a request.

### DIFF
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -51,6 +51,7 @@ type Client struct {
 	k bool // Named after curl's -k flag
 	d *debugContainer
 	t *http.Transport
+	NotThread bool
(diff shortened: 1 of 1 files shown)


pietern:
Thanks for your contribution.

The reason for having the separate goroutine is to be able to cancel the in flight request when the context gets cancelled (manually, or timed out automatically). It's possible to not use a context and in that case this function will fail. Can you change your code to check whether or not the context argument is nil or not? That way we don't need to add another field to the client struct.

Before merging your contribution, can you please sign and submit our CLA? See [the wiki](https://github.com/vmware/govmomi/wiki). Thanks.

runner-mei:
repush it   #331

PR #331, opened 2015-09-12
don't start goroutine while context is nil

runner-mei:


runner-mei:
I already send cla, and why not merge

pietern:
@runner-mei Got the confirmation. Merging the PR. Thanks for your contribution! :)

PR #336, opened 2015-09-15
Added ovf.Property output to import.spec

bruceadowns:
* ovf.Property will be used to display full property information


pietern:
Thanks @bruceadowns!

PR #338, opened 2015-09-16
Relax the convention around importing an ova

bruceadowns:
* allow for an ova file to not have a file extension
* look for *.ovf in the ova tar archive

### DIFF
--- a/govc/importx/ova.go
+++ b/govc/importx/ova.go
@@ -18,8 +18,6 @@ package importx
 
 import (
 	"flag"
-	"path"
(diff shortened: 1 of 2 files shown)


pietern:
Can you make this change in `importx/ova.go` as well? Calling `import.ova` will still look for an OVF with the same name as the OVA.

bruceadowns:
re: importx/ova.go
yup, good catch.

pietern:
Thanks @bruceadowns !

PR #339, opened 2015-09-17
Introduce govc vapp.{info|destroy|power}

caglar10ur:


dougm:
Looks great!  Just one small request, can you add an entry to govc/CHANGELOG.md ?

PR #340, opened 2015-09-17
Handle the import.spec case where no spec file is provided

bruceadowns:
* Originally this was handled implicitly via a blank extension. We now allow for a blank file extension, therefore the 'no file name provided' case needs to be explicit.


pietern:
What happens if no path is specified here? Does `cmd.Spec` panic?

bruceadowns:
Without the fix, it does the following:

$ govc import.spec
Error: open : no such file or directory

pietern:
OK, sounds good.

PR #341, opened 2015-09-18
change for looking up a VM using instanceUUID

abathla:
Added bool flag instanceUUID to find a VM with instanceUUID if set to true.


pietern:
@dougm What do you think?

dougm:
:+1: Guessing this wrapper method started as a copy-n-paste of the FindByIp wrapper and simply missed adding the additional param.

pietern:
@abathla Any reason in particular to use a `*bool` instead of a value?

abathla:
@pietern no, just that type definition of FindByUuidRequest had *bool in it.

abathla:
@pietern If I change the definition to type, then I make it mandatory for all the calls to pass it, since its a optional value, that didn't seem right to me. At the same time passing false instead of null doesn't hurt much if change it to type.

pietern:
Sounds good, thanks for the clarification!

PR #342, opened 2015-09-18
Added verbose option to the import.spec feature

bruceadowns:
* allows for a trimmed down spec that is more managable
* defaults to false


pietern:
Thanks @bruceadowns!

PR #351, opened 2015-10-05
Fix vcsa ssh config

dougm:
Commit bfabd01 switched from perl to sed - MaxSessions regex was only
valid in perl, which broke vagrant.

cc @bruceadowns


pietern:
LGTM! :)

bruceadowns:
In retrospect, the sed equivalent to this:
perl -pi -e 's/^MaxSessions \d+//' /etc/ssh/sshd_config

would have been:
sed -i -e 's/^MaxSessions [[:digit:]]\+$//' /etc/ssh/sshd_config
or
sed -i -e 's/^MaxSessions [0-9]\+$//' /etc/ssh/sshd_config

Since sed does not support \d.

PR #353, opened 2015-10-07
Use service ticket for datastore file access

dougm:
- Add object.Datastore methods to upload/download datastore files using
  a service ticket rather than user/pass authentication

- Required for http datastore access when authenticating with client
  certificate

- Faster for upload/download of large files to talk to an ESX box
  directly than via vCenter

### DIFF
--- a/govc/datastore/download.go
+++ b/govc/datastore/download.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"flag"
 
+	"golang.org/x/net/context"
+
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
 	"github.com/vmware/govmomi/vim25/soap"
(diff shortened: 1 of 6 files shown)


pietern:
Cool, LGTM.

Can we assume the port is going to be the same on the host as it is on the client's URL (which might be connected to vCenter). The ESX host can be firewalled, or using a different port. In that case, can we fall back on uploading/downloading directly through vCenter (if necessary, of course).

dougm:
Thanks for the review @pietern - I'll get rid of the client port assumption and switch to HostSystem.Summary.Config.Port.  And I can add an option to the govc datastore.{upload,download} commands to use vCenter directly.

dougm:
Regarding the port, HostSystem.Summary.Config.Port is '902' when connected directly to ESX, so that isn't useful in this case.

The client port assumption has been removed, let's assume for now that VC/ESX sdk listeners are on the default 443 port.  This was originally added for the Vagrant/esxbox case, but that is now resolved via the Datastore.httpTicket ApiType check.

If there's an issue with routing, i.e. client has a route to VC but not the chosen ESXi host, let's revisit when/if we run into that.

pietern:
Cool, go for it! :) :+1:

PR #354, opened 2015-10-13
Added a method to create snapshot of a virtual machine

gthombare:


object/virtual_machine.go
```diff
@@ -357,6 +357,24 @@ func (v VirtualMachine) Answer(ctx context.Context, id, answer string) error {
 	return nil
 }
 
+// creates a new snapshot of a virtual machine.
+func (v VirtualMachine) Snapshot(ctx context.Context, name string, description string, memoryFlag bool, quiesceFlag bool) (*Task, error) {
+	req := types.CreateSnapshot_Task{
```
dougm: Looks good, a few minor requests:
Let's use the full method name 'CreateSnapshot', since 'snapshot' is a property name of the VirtualMachine.

Start the doc with the method name to avoid golint warnings: // CreateSnapshot creates a new snapshot of a virtual machine.

Remove 'Flag' from the param names, the other wrappers use the same param names as the struct field names.

dougm:
Thanks @gthombare - I didn't get a notification from github that you had updated this, otherwise I would have merged it sooner.

gthombare:
@dougm Thanks for the review and merging my cl

PR #360, opened 2015-10-23
KeepAlive support with certificate based login

hickeng:
Seeing NotAuthenticated errors again when using certificate based auth


pietern:
Straightforward, thanks for the PR :)

PR #361, opened 2015-10-25
Add HostFirewallSystem wrapper

dougm:
- Add govc firewall.ruleset.find command

- Add esxcli.GetFirewallInfo method


pietern:
Nice! :+1:

PR #362, opened 2015-10-26
Fix KeepAlive

dougm:
See commit 9ded9c1a


pietern:
Hmm.. of course.. should have caught this.. :+1:

dougm:
I missed it too, I think we are still used to C style switch :)

PR #363, opened 2015-10-26
Add DistributedVirtualPortgroup support to vm.info

dougm:
The resource summary (-r) includes the VM network(s), we need to use the
appropriate mo type for DistributedVirtualPortgroup


pietern:
On a roll today ;) :100: :+1:

PR #368, opened 2015-11-03
Cluster related enhancements

dougm:


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ### (unreleased)
 
+* Add object.HostSystem methods: EnterMaintenanceMode, EnterMaintenanceMode, Disconnect, Reconnect
```
pietern: 2x EnterMaintenanceMode

govc/cluster/change.go
```diff
@@ -0,0 +1,96 @@
...
+	// DRS
+	f.Var(flags.NewOptionalBool(&cmd.DrsConfig.Enabled), "drs-enabled", "Enable DRS")
+
+	drsModes := []string{
+		string(types.DrsBehaviorManual),
+		string(types.DrsBehaviorPartiallyAutomated),
+		string(types.DrsBehaviorFullyAutomated),
+	}
+	f.StringVar((*string)(&cmd.DrsConfig.DefaultVmBehavior), "drs-mode", "",
+		"DRS behavior for virtual machines"+strings.Join(drsModes, ", "))
```
pietern: Add ": " or something before strings.Join?

govc/cluster/change.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	for _, path := range f.Args() {
+		clusters, err := finder.ClusterComputeResourceList(ctx, path)
+		if err != nil {
+			return err
+		}
+
+		for _, cluster := range clusters {
+			fmt.Println(cluster.InventoryPath)
```
pietern: Debug or intentional?

govc/host/destroy.go
```diff
@@ -0,0 +1,87 @@
...
+	*flags.HostSystemFlag
+}
+
+func init() {
+	cli.Register("host.destroy", &destroy{})
+}
+
+func (cmd *destroy) Register(f *flag.FlagSet) {}
+
+func (cmd *destroy) Process() error { return nil }
```
pietern: So this removes a host from a cluster? Could you add a little usage doc?

pietern:
Some minor comments, LGTM otherwise

dougm:
Thanks for the review @pietern - addressed all comments.  btw I'm tempted to rename host.destroy to host.remove, but stuck with 'destroy' as we use that name for govc calls to Destroy_Task.  Let me know if you'd prefer host.remove instead.

pietern:
@dougm Agree `host.remove` is better, since the host won't be destroyed at all ;)

Also, it's symmetric with `host.add`.

dougm:
@pietern host.remove it is

pietern:
LGTM! :+1:

PR #369, opened 2015-11-03
Add bash function to save/load GOVC environments

pietern:
I got tired of going through my history to run previously run `export GOVC_***` commands.


dougm:
cool! :+1:

PR #373, opened 2015-11-10
Add DVS and vSAN support

dougm:
More cluster related enhancements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### (unreleased)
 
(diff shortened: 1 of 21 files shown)


pietern:
Some nits:

1) Some files include a 2014 copyright header. Let's merge this though, and I can add a small thing to update existing licenses, and add a step to the check target.
2) host.vnic.info could return a struct instead, so we can use the json output for it as well.

Looks great!

dougm:
Thanks @pietern - agreed, I meant to add json output support, but that can be done after.

PR #374, opened 2015-11-11
Update copyright years in code headers

pietern:
Add `scripts/license.sh` to update file headers based on the Git log.


dougm:
:+1: love the license.sh script!

PR #376, opened 2015-11-18
Add ClusterComputeResource.Hosts() support

dougm:
Prior to this change, the ClusterComputeResource.Hosts() method would panic.


pietern:
Nice catch! :+1:

PR #378, opened 2015-11-19
Check host state in Datastore.AttachedHosts

dougm:
This method is used for datastore uploads, avoid choose a host that is
not in a known powered on and connected state.


pietern:
Nice :+1:

PR #380, opened 2015-11-21
Option to disable API version check

dougm:
- Setting GOVC_MIN_API_VERSION=- will skip the version check

- Skip version check for development builds


pietern:
LGTM :+1:

PR #381, opened 2015-11-21
figure out the datacenter name from the InventoryPath

caglar10ur:


object/datastore.go
```diff
@@ -99,13 +99,17 @@ func (d Datastore) Browser(ctx context.Context) (*HostDatastoreBrowser, error) {
 // ServiceTicket obtains a ticket via AcquireGenericServiceTicket and returns it an http.Cookie with the url.URL
 // that can be used along with the ticket cookie to access the given path.
 func (d Datastore) ServiceTicket(ctx context.Context, path string, method string) (*url.URL, *http.Cookie, error) {
+	dcPath := "ha-datacenter"
+	if d.InventoryPath != "" {
+		dcPath = strings.Split(d.InventoryPath, "/")[1]
```
dougm: This will only work if you are connected directly to ESX, otherwise the VC datacenter can have any name.  Maybe for now we could use and environment variable to override, "GOVMOMI_ESX_DATACENTER" that defaults to "ha-datacenter".  In any case, we should find out why "ha-datacenter" was renamed to "datacenter" in the dev version, and if it's going to stay way we can find a better solution.

pietern:
In the client we always know the version of the server through ServiceInfo. I think we could use that to figure out what the default name is. @caglar10ur Is the inventory path only different using the datastore APIs or across the board? If the latter, this should be changed in the general case.

dougm:
Looks like we can actually just remove dcPath from the query params and ESX will use the default value, in 6.0 at least.  @caglar10ur can you test that against the vmkernel-main branch too?

caglar10ur:
Yep, removing the dcPath for fine here

caglar10ur:
<closing this, will send another pr>

PR #385, opened 2015-11-23
Let the ESXi to figure out datastore name

caglar10ur:
Removes the dcPath as suggested by Doug @ #381

Signed-off-by: S.Çağlar Onur <conur@vmware.com>


dougm:
:+1:

PR #390, opened 2015-12-01
Remove host reference from HostFirewallSystem

dougm:
iirc, started with a method attached to object.HostFirewallSystem to get the esxcli firewall info, that didn't work out due to cyclic import.


pietern:
Cool, so all the `HostFooSystem` instances don't need a ref to the host they belong to?

:+1:

dougm:
Correct, glad to remove this 1 special case we didn't actually need!

PR #392, opened 2015-12-02
Add host storage commands

pietern:
These commands cannot yet change the host's devices in
any way, only show HBAs, LUNs, and partition tables.

### DIFF
--- a/object/host_storage_system.go
+++ b/object/host_storage_system.go
@@ -1,0 +1,80 @@
+/*
+Copyright (c) 2015 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


object/host_storage_system.go
```diff
@@ -0,0 +1,80 @@
...
+	Common
+}
+
+func NewHostStorageSystem(c *vim25.Client, ref types.ManagedObjectReference) *HostStorageSystem {
+	return &HostStorageSystem{
+		Common: NewCommon(c, ref),
+	}
+}
+
+func (s HostStorageSystem) RetrieveDiskPartitionInfo(ctx context.Context, devicePath string) (types.HostDiskPartitionInfo, error) {
```
dougm: Can we return pointer (*types.HostDiskPartitionInfo) here and the method below as we do in other wrapper methods?

dougm:
Looks great!  Just the one question on method return type.

pietern:
OK, good call, I'll change it.

dougm:
Awesome, :+1:

PR #398, opened 2015-12-09
Explicitly instantiate and register flags

pietern:
Some code in `govc/cli` would use reflection to walk the command struct, find all flags, and register them with the flag set. This proved to be inflexible in the case where a single flag should be reused multiple times under multiple names. Special casing the reflection logic seemed like the wrong thing to do and would make it more complex and implicit than it already was. This change moves all commands and flags in govc to explicitly register and process.

Caution is taken through a new common struct in `govc/flags` to ensure neither `Register` nor `Process` is run more than once per flag (through `sync.Once`).


pietern:
The `Parameterize datastore in VM tests` commit is piggybacking since I needed it to run the govc tests against a machine that didn't have a `datastore1` datastore :)

dougm:
+1, LGTM and thanks for taking on this big change!

pietern:
Thanks @dougm

PR #399, opened 2015-12-09
Specify the new entity's name upon import

pietern:
Can be specified by setting it as a property in the options file, or by passing it as a command line argument.


dougm:
:+1:

PR #400, opened 2015-12-09
Let the lister recurse into a ComputeHost

pietern:
This enables ls to look at the objects managed by a particular host.


dougm:
:+1: nice!

PR #402, opened 2015-12-11
Move govc url.Parse wrapper to soap.ParseURL

dougm:
By exposing this wrapper, programs other than govc can easily use the
same url defaults as govc (scheme, path).


pietern:
Cool :+1:

PR #403, opened 2015-12-11
Script to capture and decrypt hostd SOAP traffic

dougm:


pietern:
Black magic... :8ball: :shipit:

PR #406, opened 2015-12-15
Use OptionalBool for ExpandableReservation

dougm:


pietern:
Could you add a test as well where the expendable flag is used? With the dereferencing going on in that flag I'd like to make sure this doesn't regress.

dougm:
@pietern good call, done.

pietern:
Great :+1:

PR #408, opened 2015-12-18
Add support for custom session keep alive handler

dougm:
- Switch to CurrentTime in the default handler (smaller payload)
  And useful for custom handlers as it requires authentication, whereas
  ServiceContent does not

- Add session.Manager TerminateSession method


pietern:
Looks good to me! :+1:

PR #411, opened 2016-01-10
clear golint warnings

casualjim:
Clears a bunch of golint warnings

This also adds a golint make task. You can check that one to find out where we're violating the naming conventions on our exported interface.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 all: check test
 
-check: goimports govet
+check: goimports govet golint
 
 goimports:
 	@echo checking go imports...
@@ -11,6 +11,17 @@ goimports:
 govet:
 	@echo checking go vet...
 	@go tool vet -structtags=false -methods=false .
+
+golint:
+	@echo checking go lint ...
+	@go get -v github.com/golang/lint/golint
+	@for file in $$(find . -name '*.go' | grep -v 'vendor\|govc\|vim25\|test\|authorization_manager.go\|customization_spec_manager.go\|datacenter.go\|host_network_system.go\|http_nfc_lease.go\|search_index.go\|virtual_app.go'); do \
+		golint $${file}; \
+		if [ -n "$$(golint $${file})" ]; then \
+			exit 1; \
+		fi; \
+	done
+
 
 test:
 	go get
--- a/client.go
+++ b/client.go
@@ -14,8 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/*
-This package is the root package of the govmomi library.
+/*Package govmomi implements the vmware monitoring interface.
 
 The library is structured as follows:
 
@@ -68,6 +67,7 @@ import (
 	"golang.org/x/net/context"
 )
 
+// Client represents a govmomi client
 type Client struct {
 	*vim25.Client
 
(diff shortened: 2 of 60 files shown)


dougm:
Thanks @casualjim !

How about if we do the following, a 'golint' target that applies all .go files, ignoring just comment warnings, something like:
```
golint | sh -c '! grep -v comment'
```

Then we have a separate 'golint-comment' target where we specify the smaller set of packages/files that should be 100% lint clean.  And we can take the time to incrementally add/improve the doc.  I'd prefer to avoid adding comments just to keep golint quiet, but a copy-n-paste from the sdk and/or link would be fine.  What do you think?  cc @pietern

casualjim:
I rewrote my committer email here.

Also there are other kind of warnings that get triggered too, all of them are in the object package though:

```shellsession
git:(golint-warnings) ✗ !? » golint ./object
object/authorization_manager.go:45:32: method ById should be ByID
object/customization_spec_manager.go:150:36: method CustomizationSpecItemToXml should be CustomizationSpecItemToXML
object/customization_spec_manager.go:165:36: method XmlToCustomizationSpecItem should be XMLToCustomizationSpecItem
object/datacenter.go:31:2: struct field VmFolder should be VMFolder
object/host_network_system.go:206:28: method UpdateConsoleIpRouteConfig should be UpdateConsoleIPRouteConfig
object/host_network_system.go:221:28: method UpdateDnsConfig should be UpdateDNSConfig
object/host_network_system.go:236:28: method UpdateIpRouteConfig should be UpdateIPRouteConfig
object/host_network_system.go:251:28: method UpdateIpRouteTableConfig should be UpdateIPRouteTableConfig
object/http_nfc_lease.go:32:6: type HttpNfcLease should be HTTPNfcLease
object/http_nfc_lease.go:37:6: func NewHttpNfcLease should be NewHTTPNfcLease
object/http_nfc_lease.go:44:23: method HttpNfcLeaseAbort should be HTTPNfcLeaseAbort
object/http_nfc_lease.go:59:23: method HttpNfcLeaseComplete should be HTTPNfcLeaseComplete
object/http_nfc_lease.go:73:23: method HttpNfcLeaseGetManifest should be HTTPNfcLeaseGetManifest
object/http_nfc_lease.go:87:23: method HttpNfcLeaseProgress should be HTTPNfcLeaseProgress
object/search_index.go:60:22: method FindByDnsName should be FindByDNSName
object/search_index.go:101:22: method FindByIp should be FindByIP
object/search_index.go:124:22: method FindByUuid should be FindByUUID
object/search_index.go:124:98: method parameter instanceUuid should be instanceUUID
object/virtual_app.go:62:21: don't use underscores in Go names; method CreateChildVM_Task should be CreateChildVMTask
object/virtual_app.go:93:21: don't use underscores in Go names; method PowerOnVApp_Task should be PowerOnVAppTask
object/virtual_app.go:107:21: don't use underscores in Go names; method PowerOffVApp_Task should be PowerOffVAppTask
object/virtual_app.go:123:21: don't use underscores in Go names; method SuspendVApp_Task should be SuspendVAppTask
```

dougm:
Ah right, we are using the same case as the SDK.  I'd be ok with fixing the case to make golint happy and to avoid more of these in the future.  What do you think @pietern ?  We can make that a separate change, but would like to avoid the big grep -v list in the Makefile staying around too long.

pietern:
@dougm I agree, let's make it a separate change, and do a version bump as well. Then the whole tree is golint clean, we add it to the build, and we keep it that way :)

dougm:
Sounds good to me.  Looks like you just need to rebase @casualjim

casualjim:
k will rebase and will submit a second PR to fix the casing issues

pietern:
There are also a couple of instances where "ovf", "vsphere", "vmware" should be properly cased.

I still do think many of this stuff is overhead, we might as well just add comments where comments are due. The majority of the comments here don't explain anything what the function name already explains. It's just to keep golint happy.

PR #412, opened 2016-01-11
Creating Platform Specific govc ls Commands.

excilsploft:
I use VMWare integrated openstack, this has the effect of creating paths in VCenter that have lots of spaces and parentheses (usually with gnarly uuids in between) by escaping the the spaces and parentheses, I am able to grep for a path listing and put it directly to pvcopy and descend the path by simply pasting (or addressing the target without tediously hand editing escapes in)


pietern:
Hi,

I don't think adding platform specific implementations is the right approach for your problem. In fact, with the shell you are able to do this without any modifications:

```
$ govc ls vm | sed -e "s/'/'\\\\''/g; 1s/^/'/; \$s/\$/'/"
'/ha-datacenter/vm/vm name with spaces'
```

Now that is arguably more tedious than manually adding escaping, but you can alias is, or even a wrapper script for pbcopy that does this for you. This would look something like this:

```
#!/bin/sh -e
sed -e "s/'/'\\\\''/g; 1s/^/'/; \$s/\$/'/" | pbcopy
```

And can be used like this:

```
$ govcx ls vm | grep whatever | xpbcopy
```

PR #413, opened 2016-01-13
Move stat function to object.Datastore

pietern:
Change the errors it returns to include the file or directory that the user attempted to stat.


dougm:
nice :+1:

PR #414, opened 2016-01-13
Print os.Args[0] in error messages

pietern:
Conforms to most other Unix tools. Makes it easier to spot govc specific errors in script output.


dougm:
:+1:

PR #420, opened 2016-01-18
Fix bats tests

dougm:
Error message output changed in commits 72bdb45 and 0a4c978


pietern:
Argh, yeah we should fix up some real CI here...  :+1:

PR #421, opened 2016-01-19
Support vapp in pool.info command

dougm:
A VirtualApp extends ResourcePool, add '-a' option to pool.info to
support listing vapps.  Remove vapp.info command as it is functionally
the same as pool.info, we can use the vapp.info namespace for something
vapp specific.

Fix pool.info memory status (convert bytes to MB)


dougm:
cc @caglar10ur

caglar10ur:
:shipit:

dougm:
Thanks @caglar10ur .  Can you review @pietern ?

pietern:
LGTM :+1:

PR #425, opened 2016-01-26
change datastore.upload and datastore.download to work with VSphere 5.1 

### LINKED ISSUE #424: govc datastore.download/datastore.upload
```
[25-01-16 17:32:05] Downloading... OK
govc: ServerFaultCode: A specified parameter was not correct.
url
```
I get the above error when using `govc` with a 5.1 datastore. 
It appears that `ds.DownloadFile(context.TODO(), args[0], args[1], &p)` in datastore/download.go and `ds.UploadFile(context.TODO(), args[0], args[1], &p)` in datastore/upload.go don't work with 5.1 datastores. 

I'll be submitting a pull request that resolves the issue shortly.

groob:
Fixes #424 for me.

### DIFF
--- a/govc/datastore/download.go
+++ b/govc/datastore/download.go
@@ -57,7 +57,12 @@ func (cmd *download) Run(ctx context.Context, f *flag.FlagSet) error {
 		return errors.New("invalid arguments")
 	}
 
-	ds, err := cmd.Datastore()
+	dsurl, err := cmd.DatastoreURL(args[0])
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


dougm:
Thanks for the patch @groob - We had switched to using Datastore.{Upload,Download}File in commit 6d4932a to support client certificate authentication.  I'd like to take a closer look at why this fails against 5.1.

We also need you to submit a CLA to accept contributions: https://github.com/vmware/govmomi/wiki
Thanks again.

PR #427, opened 2016-01-26
Use gvt to get dependencies

fdawg4l:
The go vendor experiment (GO15VENDOREXPERIMENT) adds a vendor/ dir which out of tree imports can be added.  Gvt can be used to specify the import at a specific revision.  Use gvt to get the context package at the given release.  This will avoid build breakages if the context package changes underneath you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 sudo: false
 language: go
-go: 1.4
+go: 1.5
 
 before_install:
   - go get golang.org/x/tools/cmd/vet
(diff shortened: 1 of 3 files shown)


pietern:
What happens if gvt changes? ;)

Seriously though, net/context is the only non-stdlib dependency, and is maintained by the Go team itself. I think that we can reasonably assume that its public API doesn't change. I don't think we should take a dependency on an external tool just to facilitate importing that package at a specific rev.

fdawg4l:
Fair enough, @pietern!  Agree on closing.

PR #429, opened 2016-01-27
Introduce AttachedClusterHosts

caglar10ur:
A datastore could be attached to different hosts in different
clusters.


dougm:
Nice :+1:

PR #431, opened 2016-01-28
introduce Datastore.Type()

caglar10ur:
which returns the type of file system volume.

### DIFF
--- a/object/datastore.go
+++ b/object/datastore.go
@@ -339,3 +339,13 @@ func (d Datastore) Stat(ctx context.Context, file string) (types.BaseFileInfo, e
 	return res.File[0], nil
 
 }
+
+// Type returns the type of file system volume.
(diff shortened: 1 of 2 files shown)


object/datastore.go
```diff
@@ -339,3 +339,13 @@ func (d Datastore) Stat(ctx context.Context, file string) (types.BaseFileInfo, e
 	return res.File[0], nil
 
 }
+
+// Type returns the type of file system volume.
+func (d Datastore) Type(ctx context.Context) (types.HostFileSystemVolumeFileSystemType, error) {
+	var mds mo.Datastore
+
+	if err := d.Properties(ctx, d.Reference(), []string{"summary"}, &mds); err != nil {
```
dougm: Should be able to use "summary.type" here instead of "summary".  Small payload in this case, but good habit to limit the property collector.

caglar10ur:
Merging it as Doug verbally give his green light :)

PR #435, opened 2016-01-30
govc.el: minor fixes for distribution as a package

purcell:
- "Emacs" in the package description line is redundant
- Declare a dependency on Emacs 24.3, for setq-local and tabulated-list-mode
- Don't use an illegal key prefix: C-c [letter] is reserved for
  user customisation, as per https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html

In connection with https://github.com/milkypostman/melpa/pull/3546

### DIFF
--- a/govc/emacs/govc.el
+++ b/govc/emacs/govc.el
@@ -1,10 +1,10 @@
-;;; govc.el --- Emacs interface to govc for managing VMware ESXi and vCenter
+;;; govc.el --- Interface to govc for managing VMware ESXi and vCenter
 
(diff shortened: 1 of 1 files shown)


dougm:
Thanks @purcell !  I had been made aware of the prefix conventions, but thought optional global-minor-modes counted as user customisation - since projectile defaults to "C-c p" and docker.el defaults to "C-c d".  But just now found comments in bbatsov/projectile#150 that this was an intentional violation.

PR #437, opened 2016-02-02
Handle import statements same as include

caglar10ur:


pietern:
Out of curiosity: why do you need the import tag to be handled? Is there a difference in generated code?

caglar10ur:
I'm working on adding some internal API bits to the govmomi and they use import for those wsdl files, that's how I ended up finding this :) 

It doesn't do anything for public code though...

dougm:
:+1: we can argue about the internal APIs in another thread :)

PR #439, opened 2016-02-03
Update missing property whitelist

dougm:
Ignore missing alarmActionsEnabled missing property, possible with a
vApp child VM.


pietern:
LGTM

PR #440, opened 2016-02-04
Remove hardcoded urn:vim25 value from vim_wsdl.rb

caglar10ur:
Add internalvim25 and internalreflect and grab urn from the wsdl.

### DIFF
--- a/gen/vim_wsdl.rb
+++ b/gen/vim_wsdl.rb
@@ -141,12 +141,14 @@ class Simple
 
   def vim_type?
     ns, _ = self.type.split(":", 2)
-    ns == "vim25"
+    ns == "vim25" or ns == "internalvim25" or ns == "internalreflect"
   end
 
   def vim_type(t = self.type)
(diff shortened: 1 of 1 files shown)


gen/vim_wsdl.rb
```diff
@@ -141,12 +141,14 @@ def var_name
...
   def vim_type?
     ns, _ = self.type.split(":", 2)
-    ns == "vim25"
+    ns == "vim25" or ns == "internalvim25" or ns == "internalreflect"
   end
 
   def vim_type(t = self.type)
     ns, t = t.split(":", 2)
-    raise if ns != "vim25"
+    if ns != "vim25" and ns != "internalvim25" and ns != "internalreflect"
```
dougm: Maybe put these in 1 place, a var or a method like so:
```ruby
def valid_ns?(ns)
  ["vim25", "internalvim25", "internalreflect"].include?(ns)
end
```

PR #441, opened 2016-02-05
Remove duplication code with cmd.persist check

arbrix:


govc/vm/disk/attach.go
```diff
@@ -90,11 +90,7 @@ func (cmd *attach) Run(ctx context.Context, f *flag.FlagSet) error {
 	backing := disk.Backing.(*types.VirtualDiskFlatVer2BackingInfo)
 
 	if cmd.link {
-		if cmd.persist {
-			backing.DiskMode = string(types.VirtualDiskModeIndependent_persistent)
-		} else {
-			backing.DiskMode = string(types.VirtualDiskModeIndependent_persistent)
```
dougm: Thanks for the PR @arbrix 
I think this should actually be VirtualDiskModeIndependent_nonpersistent here, rather than always setting to persistent.

caglar10ur:
ping @arbrix

govc/vm/disk/attach.go
dougm: done in #579

PR #444, opened 2016-02-05
Fix issue with optional UnitNumber

grrtrr:
The UnitNumber is optional in the wsdl, hence the 'omitempty' directive
causes issues, as evidenced by 2 FIXMEs in the code. This addresses
those FIXMEs, being able to properly distinguish between absent optional
UnitNumber and a properly present zero-valued UnitNumber.


grrtrr:
Sorry, forgot the test cases. Closing this, will resubmit when tests pass.

PR #446, opened 2016-02-05
Fix issue with optional UnitNumber

grrtrr:
The UnitNumber is optional in the wsdl, hence the 'omitempty' directive
causes issues, as evidenced by 2 FIXMEs in the code. This addresses
those FIXMEs, being able to properly distinguish between absent optional
UnitNumber and a properly present zero-valued UnitNumber.

### DIFF
--- a/govc/device/info.go
+++ b/govc/device/info.go
@@ -125,7 +125,11 @@ func (r *infoResult) Write(w io.Writer) error {
 		} else {
 			if c := r.list.FindByKey(d.ControllerKey); c != nil {
 				fmt.Fprintf(tw, "  Controller:\t%s\n", r.Devices.Name(c))
-				fmt.Fprintf(tw, "  Unit number:\t%d\n", d.UnitNumber)
+				if d.UnitNumber != nil {
+					fmt.Fprintf(tw, "  Unit number:\t%d\n", *d.UnitNumber)
+				} else {
+					fmt.Fprintf(tw, "  Unit number:\t<nil>\n")
+				}
 			}
 		}
 
--- a/govc/importx/ovf.go
+++ b/govc/importx/ovf.go
@@ -231,16 +231,6 @@ func (cmd *ovfx) Import(fpath string) (*types.ManagedObjectReference, error) {
 		}
 	}
 
-	// TODO: ImportSpec may have unitNumber==0, but this field is optional in the wsdl
-	// and hence omitempty in the struct tag; but unitNumber is required for certain devices.
-	s := &spec.ImportSpec.(*types.VirtualMachineImportSpec).ConfigSpec
-	for _, d := range s.DeviceChange {
-		n := &d.GetVirtualDeviceConfigSpec().Device.GetVirtualDevice().UnitNumber
-		if *n == 0 {
-			*n = -1
-		}
-	}
-
 	var host *object.HostSystem
 	if cmd.SearchFlag.IsSet() {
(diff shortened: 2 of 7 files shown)


grrtrr:
I have signed the licence agreement. Please let me know if you would like me to close this PR and resubmit?

dougm:
@grrtrr Thank for the patch and apologies for the delayed response.  Looks great at first glance.  One issue we have is vim25/types/types.go being generated. I can't recall if we have a way to deal with such a case, such that we won't have to manually change that field type in the future.  I will take a closer look tonight or tomorrow.

grrtrr:
@dougm Thank you for the heads-up; I will also have a look if I can do this locally.

dougm:
@grrtrr when you have a chance, can you rebase against vmware/master ?  I'm thinking we can merge your change as-is and revisit the generated part next time we touch the generator.

grrtrr:
@dougm - closing this one, will re-submit with rebased master. Thank you.

grrtrr:
Forgot how smart github is -- it updated the commit from the just-rebased grrtrr/govmomi master branch. Hence keeping this open for reconsideration.

dougm:
Thanks @grrtrr .  Trying out your change, some of the bats tests are failing (see govc/test).  Without running the full test suite, can reproduce with:
```
source govc/test/test_helper.bash
vm=$(new_ttylinux_vm)
govc: Invalid configuration for device '1'.
```

Looks like this issue is related to vm.create with a -disk param:
```
govc vm.create -disk ttylinux-pc_i486-16.1-disk1.vmdk -on=false foo
govc: Invalid configuration for device '1'.
```

I can have a closer look later, if you have any ideas that'd be great.

grrtrr:
@dougm So far I have had no luck on my end, need to customize the bash helper script to our environment; currently all I can get out of the above command is `govc: A component of the virtual machine is not accessible on the host.`

dougm:
@grrtrr You should be able to reproduce the issue without the helper, just by using `vm.create -disk any-existing.vmdk foo`

grrtrr:
@dougm - I was able to reproduce the issue. Problem was caused by `object/virtual_device_list.go:AssignController()` not assigning `-1` if the `UnitNumber` was 0. The `govc vm.create` command now works.

Before I resubmit the changes, I have one question where the change is not totally clear to me: in `govc/importx/vmd.go`, the `configSpec.AddDisk()` set the `UnitNumber` to -1 before this patch; now it would be `nil`. Can you see whether this would require a -1? I'd expect no, but am not fully sure.

dougm:
Great, glad you found the fix.  I don't think importx needs UnitNumber set to -1, I can test it when you resubmit.  Thanks.

dougm:
Thanks @grrtrr , I tried it out and tests are happy again.  I'm still debating what to do with respect to the generated code, but that likely won't block merging this.  I also meant to ask earlier, did you run into an issue directly that this change fixes?  Or did you see the FIXME comments and decided to address them?

grrtrr:
Yes, this fix was caused by an issue: in doing `.ovf`/`.ova` import, we kept on running into the same problem caused by zero `UnitNumber` values. We tracked the cause down due to an optional `UnitNumber` value, discovering the `FIXME` notes on the way. Have been testing the fix in our codebase for some time; it fixed the problem with the optional `UnitNumber` for us.

dougm:
Thanks again @grrtrr - going to merge this and we can revisit the generated code when the time comes.

PR #448, opened 2016-02-17
Add switch to specify protocol version in SOAPAction header

andrewtchin:
Added `-vim-namespace` and `-vim-version` flags to govc. Also set by corresponding env vars documented in README.md
```./govc -vim-namespace urn:vim25 -vim-version 6.0 -k -persist-session=false```

Also updated `govc/test/images/update.sh` since I was getting bad filenames from bintray due to akamai.

Addresses https://github.com/vmware/govmomi/issues/384

### DIFF
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -45,6 +46,9 @@ type RoundTripper interface {
 	RoundTrip(ctx context.Context, req, res HasFault) error
 }
 
+var DefaultVimNamespace = "urn:vim25"
+var DefaultVimVersion = "6.0"
+
 type Client struct {
 	http.Client
 
@@ -53,6 +57,9 @@ type Client struct {
 	d *debugContainer
 	t *http.Transport
 	p *url.URL
+
+	Namespace string // Vim namespace
+	Version   string // Vim version
 }
 
 var schemeMatch = regexp.MustCompile(`^\w+://`)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: vim25/soap/client.go [hunk 1]
```diff
@@ -45,14 +45,19 @@ type RoundTripper interface {
...
-	k bool // Named after curl's -k flag
-	d *debugContainer
-	t *http.Transport
-	p *url.URL
+	u   *url.URL
+	k   bool // Named after curl's -k flag
+	d   *debugContainer
+	t   *http.Transport
+	p   *url.URL
+	Ns  string // Vim namespace
```
dougm: Let's name these fields 'VimNamespace' and 'VimVersion'.  Also put an empty line before the VimNamespace field, this will maintain the alignment of the private fields.

vim25/soap/client.go
```diff
@@ -255,7 +266,11 @@ func (c *Client) RoundTrip(ctx context.Context, reqBody, resBody HasFault) error
 	}
 
 	req.Header.Set(`Content-Type`, `text/xml; charset="utf-8"`)
-	req.Header.Set(`SOAPAction`, `urn:vim25/6.0`)
+	var soapAction bytes.Buffer
+	soapAction.WriteString(c.Ns)
+	soapAction.WriteString("/")
+	soapAction.WriteString(c.Ver)
+	req.Header.Set(`SOAPAction`, soapAction.String())
```
dougm: No need for bytes.Buffer here, you could do fmt.Sprintf("%s/%s", ...) or even path.Join(c.VimNameSpace, c.VimVersion)

andrewtchin:
Updated PR to address Doug's comments

dougm:
LGTM

PR #449, opened 2016-02-18
Fix vm.change's ExtraConfig values being truncated at equal signs

saracen:
`govc vm.change -vm=test -e "guestinfo.hello=world=foo"` would result in `guestinfo.hello` having the value **`world`**, rather than **`world=foo`**.


dougm:
:+1: thanks

frapposelli:
:tada: first automatic cla PR merged :tada:

PR #456, opened 2016-03-09
Added new govc vm.disk.create -mode argument for selecting one the VirtualDiskMode types

eyz:
I hope this is helpful to others who require specific disk modes in their environments.

Please note that this commit includes three disk modes that exist in enum.go as of 779ae0a131adb4b36090fc712cb49c3bb775edbb, but are not represented in the version 6.0.0 build 2594327 web client. They are as follows:

- VirtualDiskModeNonpersistent (https://github.com/vmware/govmomi/blob/779ae0a131adb4b36090fc712cb49c3bb775edbb/vim25/types/enum.go#L3043)

- VirtualDiskModeUndoable (https://github.com/vmware/govmomi/blob/779ae0a131adb4b36090fc712cb49c3bb775edbb/vim25/types/enum.go#L3044) 

- VirtualDiskModeAppend (https://github.com/vmware/govmomi/blob/779ae0a131adb4b36090fc712cb49c3bb775edbb/vim25/types/enum.go#L3047)

Please let me know if you would rather I remove these options before you consider merging.

Thank you

### DIFF
--- a/govc/vm/disk/create.go
+++ b/govc/vm/disk/create.go
@@ -37,6 +37,7 @@ type create struct {
 	Bytes      units.ByteSize
 	Thick      bool
 	Eager      bool
+	DiskMode   string
 }
 
 func init() {
@@ -61,6 +62,7 @@ func (cmd *create) Register(ctx context.Context, f *flag.FlagSet) {
 	f.Var(&cmd.Bytes, "size", "Size of new disk")
 	f.BoolVar(&cmd.Thick, "thick", false, "Thick provision new disk")
 	f.BoolVar(&cmd.Eager, "eager", false, "Eagerly scrub new disk")
+	f.StringVar(&cmd.DiskMode, "mode", "dependent", "One of: dependent (also known as persistent), persistent, nonpersistent, undoable, independent_persistent, independent_nonpersistent, append")
 }
 
 func (cmd *create) Process(ctx context.Context) error {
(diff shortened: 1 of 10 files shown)


govc/vm/disk/create.go
```diff
@@ -61,6 +62,7 @@ func (cmd *create) Register(ctx context.Context, f *flag.FlagSet) {
 	f.Var(&cmd.Bytes, "size", "Size of new disk")
 	f.BoolVar(&cmd.Thick, "thick", false, "Thick provision new disk")
 	f.BoolVar(&cmd.Eager, "eager", false, "Eagerly scrub new disk")
+	f.StringVar(&cmd.DiskMode, "mode", "dependent", "One of: dependent (also known as persistent), persistent, nonpersistent, undoable, independent_persistent, independent_nonpersistent, append")
```
dougm: :+1: on adding the option.  Could you change the usage similar to the other choice options?  Some of the examples:
```
govc/datastore/create.go:119:	f.Var(&cmd.Type, "type", fmt.Sprintf("Datastore type (%s)", strings.Join(allTypes, "|")))
govc/datastore/create.go:126:		fmt.Sprintf("Access mode for the mount point (%s)", strings.Join(modes, "|")))
govc/device/scsi/add.go:53:		fmt.Sprintf("SCSI controller type (%s)", strings.Join(ctypes, "|")))
govc/dvs/portgroup/add.go:57:		fmt.Sprintf("Portgroup type (%s)", strings.Join(ptypes, "|")))
```

govc/vm/disk/create.go
```diff
@@ -113,10 +115,31 @@ func (cmd *create) Run(ctx context.Context, f *flag.FlagSet) error {
...
+
 	if cmd.Thick {
-		backing := disk.Backing.(*types.VirtualDiskFlatVer2BackingInfo)
 		backing.ThinProvisioned = types.NewBool(false)
-		backing.EagerlyScrub = types.NewBool(cmd.Eager)
+	}
+
+	backing.EagerlyScrub = types.NewBool(cmd.Eager)
+
+	switch cmd.DiskMode {
```
dougm: "dependent" isn't listed in the docs: http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.vm.device.VirtualDiskOption.DiskMode.html

Should be ok just to leave it out of the usage above.  Then you can also use the same []string{...} for usage and validation.

govc/vm/disk/create.go
eyz: Sure. With the pipe in between on the usage output, or specifically joining them from an array of options for the usage?

dougm: Pipe just between the options, for example:
```go
f.StringVar(&cmd.DVPortgroupConfigSpec.Type, "type", ptypes[0],
		fmt.Sprintf("Portgroup type (%s)", strings.Join(ptypes, "|")))
```

govc/vm/disk/create.go
eyz: I noticed "dependent" in the UI becomes persistent. I will remove dependent from the new string array

govc/vm/disk/create.go
eyz: OK, thanks for the code review

eyz:
I botched my local repo, so please close and I'll re-submit.

dougm:
You just needed to:
```
git fetch -a
git rebase vmware/master
```
Where 'vmware' is the name of the remote for github.com/vmware/govmomi (yours might be named something else, 'origin' or 'upstream' for example)

PR #458, opened 2016-03-12
Fix VirtualDeviceList.CreateFloppy

dougm:
device.floppy.add related tests have been failing since bfe414f


dougm:
@jzt this fixes a few of the test failures you were seeing.

PR #461, opened 2016-03-14
First pass at running govc tests with Drone

jzt:
There are a few tests that are failing or hanging when running against an ESX in vCA, and they are currently skipped. All other tests are passing successfully.

TODO:
- Add and debug the remaining tests
- Use a golang:1.6-based container that includes dependencies


dougm:
Looks good @jzt .  Which tests are still failing?  If there are some tests that don't work in this env, we can look at changing them and/or disabling via an environment variable.  I can help to debug the failing tests.

dougm:
Also, once we have the bats tests sorted out, we can also set GOVMOMI_TEST_URL to the same value as $GOVC_TEST_URL - that'll enable some Go tests that depend on ESX.

jzt:
@dougm I know of one that looked networking-related was failing, but I couldn't tell whether or not it was failing because of the ESX or because of the test itself. It seems like the networking side of vCA has caused some issues for other ESX hosts in the same datacenter, so it's something I need to look into. I am adding more tests right now and will give you a better idea of which ones are causing problems.

jzt:
@dougm Here's are the failures I have so far:

```
not ok 39 import.ovf with name as argument
# (from function `flunk' in file govc/test/test_helper.bash, line 156,
#  from function `assert_success' in file govc/test/test_helper.bash, line 161,
#  in test file govc/test/import.bats, line 56)
#   `assert_success' failed
# command failed with exit status 1: [14-03-16 21:04:05] Warning: Line 62: Invalid value 'nat' for element 'Connection'
```
```
# govc: ServerFaultCode: The operation is not allowed in the current state.
not ok 40 license.add
# (from function `assert_equal' in file govc/test/test_helper.bash, line 179,
#  in test file govc/test/license.bats, line 27)
#   `assert_equal "License is not valid for this product" $(get_key 00000-00000-00000-00000-00001 <<<${output} | get_property diagnostic)' failed
# expected: License is not valid for this product
# actual:   License
```
```
not ok 42 license.list
# (from function `assert_equal' in file govc/test/test_helper.bash, line 179,
#  in test file govc/test/license.bats, line 45)
#   `assert_equal "Evaluation Mode" $(get_key 00000-00000-00000-00000-00000 <<<$output | jq -r ".Name")' failed
# expected: Evaluation Mode
# actual:   Evaluation
```
```
not ok 48 ls network
# (in test file govc/test/ls.bats, line 44)
#   `[ ${#lines[@]} -eq 1 ]' failed
```

It hung after (skipped) test 64:
```
ok 64 # skip (requires vcsim) vm.create -pool cluster
```

Not sure why as I don't get any output but monitoring the ESX shows that no new tasks are being run for upwards of 15 minutes.

Here are the skipped tests, just for reference:

```
ok 6 # skip (requires vcsim) create and destroy datacenters
ok 7 # skip (requires vcsim) destroy datacenter using glob
ok 8 # skip (requires vcsim) destroy datacenter that doesn't exist
ok 9 # skip (requires vcsim) create datacenter that already exists
ok 29 # skip (requires vcsim) extension
ok 30 # skip (requires vcsim) fields
ok 33 # skip (requires vcsim) host info vc
ok 49 # skip (requires vcsim) ls multi ds
ok 50 # skip (requires vcsim) network dvs backing
ok 51 # skip (requires vcsim) network change backing
ok 54 # skip (requires vcsim) network flag required
ok 64 # skip (requires vcsim) vm.create -pool cluster
```

I'm investigating the hang, I'll let you know what I find.

dougm:
The hanging tests would be waiting for a VM ip address:
```
ok 65 vm.ip
ok 66 vm.ip -esxcli
```

We could add a 'skip' for the failing/hanging tests for now, based on an existing drone env var or adding a new one to .drone.yml.  Then we can revisit later.

jzt:
Skipping failed tests for now, as well as the two causing the hang.

andrewtchin:
LGTM

govc/test/import.bats
```diff
@@ -50,6 +50,7 @@ load test_helper
 }
 
 @test "import.ovf with name as argument" {
+  skip
```
dougm: We don't want to always skip these.  Let's add something like the vcsim_env env helper, call it 'skip_if_vca' - the skip_if_vca function will check for an env var to see if running on vca (fine to set via drone.yml too).  And add a message, something like: skip "disabled on vCA"

jzt:
@dougm: Now skipping tests properly when running in vCA environment.

dougm:
:100:

PR #466, opened 2016-03-15
Double quote network name to prevent space in name from failing

cgtexmex:
Added double quotes to networking tests -- if there's a more precise way to do this please share.

Fixes 465

### DIFF
--- a/govc/test/ls.bats
+++ b/govc/test/ls.bats
@@ -39,15 +39,15 @@ load test_helper
   [ ${#lines[@]} -ge 1 ]
 
   local path=${lines[0]}
-  run govc ls $path
+  run govc ls "$path"
   assert_success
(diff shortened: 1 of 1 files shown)


govc/test/ls.bats
```diff
@@ -39,15 +39,15 @@ load test_helper
   [ ${#lines[@]} -ge 1 ]
 
   local path=${lines[0]}
-  run govc ls $path
+  run govc ls "$path"
   assert_success
   [ ${#lines[@]} -eq 1 ]
 
-  run govc ls network/$(basename $path)
+  run govc ls "network/$(basename "$path")""
```
jzt: Extraneous trailing `"`? (I bet your editor did that) :)

dougm:
:+1:

PR #468, opened 2016-03-16
Implemented the ablitiy to tail the vSphere event stream

cgtexmex:
- Tail and Force flag added to govc events command

### DIFF
--- a/event/helper.go
+++ b/event/helper.go
@@ -1,0 +1,116 @@
+/*
+Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package event
+
+import (
+	"fmt"
+
+	"golang.org/x/net/context"
+
+	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/view"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func multipleObjectEvents(ctx context.Context, m Manager, objects []types.ManagedObjectReference, pageSize int, tail bool, force bool, prop []string, f func([]types.BaseEvent) error) error {
+	// create an EventHistoryCollector for each object
+	var collectors []types.ManagedObjectReference
+	for _, o := range objects {
+		filter := types.EventFilterSpec{
+			Entity: &types.EventFilterSpecByEntity{
+				Entity:    o,
+				Recursion: types.EventFilterSpecRecursionOptionAll,
+			},
+		}
+
+		collector, err := m.CreateCollectorForEvents(ctx, filter)
+		if err != nil {
+			return fmt.Errorf("[%#v] %s", o, err)
+		}
+		defer collector.Destroy(ctx)
+
+		err = collector.SetPageSize(ctx, pageSize)
+		if err != nil {
+			return err
+		}
+
+		collectors = append(collectors, collector.Reference())
+	}
+
+	// create and populate a ListView
+	viewMgr := view.NewManager(m.Client())
+	listView, err := viewMgr.CreateListView(ctx, collectors)
+	if err != nil {
+		return err
+	}
+	count := 0
+	// Retrieve the property from the objects in the ListView
+	return property.WaitForView(ctx, property.DefaultCollector(m.Client()), listView.Reference(), collectors[0], prop, func(pc []types.PropertyChange) bool {
+		for _, u := range pc {
+			if u.Name != prop[0] {
+				continue
+			}
+			if u.Val == nil {
+				continue
+			}
+			f(u.Val.(types.ArrayOfEvent).Event)
+		}
+		count++
+		if count == len(collectors) && !tail {
+			return true
+		}
+		return false
+	})
+
+}
+
+func singleObjectEvents(ctx context.Context, m Manager, object types.ManagedObjectReference, pageSize int, tail bool, force bool, prop []string, f func([]types.BaseEvent) error) error {
+	filter := types.EventFilterSpec{
+		Entity: &types.EventFilterSpecByEntity{
+			Entity:    object,
+			Recursion: types.EventFilterSpecRecursionOptionAll,
+		},
+	}
+
+	collector, err := m.CreateCollectorForEvents(ctx, filter)
+	if err != nil {
+		return fmt.Errorf("[%#v] %s", object, err)
+	}
+	defer collector.Destroy(ctx)
+
+	err = collector.SetPageSize(ctx, pageSize)
+	if err != nil {
+		return err
+	}
+
+	return property.Wait(ctx, property.DefaultCollector(m.Client()), collector.Reference(), prop, func(pc []types.PropertyChange) bool {
+		for _, u := range pc {
+			if u.Name != prop[0] {
+				continue
+			}
+			if u.Val == nil {
+				continue
+			}
+			f(u.Val.(types.ArrayOfEvent).Event)
+
+		}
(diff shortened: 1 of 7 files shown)


event/helper.go
```diff
@@ -0,0 +1,121 @@
...
+		for _, u := range pc {
+			count++
+			if u.Name != prop[0] {
+				continue
+			}
+			if u.Val == nil {
+				continue
+			}
+			f(u.Val.(types.ArrayOfEvent).Event)
+			if !tail {
```
dougm: Looks like this check should be after the for loop, no?

event/helper.go
```diff
@@ -0,0 +1,121 @@
...
+		return fmt.Errorf("[%#v] %s", object, err)
+	}
+	defer collector.Destroy(ctx)
+
+	err = collector.SetPageSize(ctx, pageSize)
+	if err != nil {
+		return err
+	}
+
+	count := 0
```
dougm: Doesn't look like count is needed in this func.

govc/events/command.go
```diff
@@ -45,6 +48,8 @@ func (cmd *events) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.DatacenterFlag.Register(ctx, f)
 
 	f.IntVar(&cmd.Max, "n", 25, "Output the last N events")
+	f.BoolVar(&cmd.Tail, "t", false, "Tail events")
```
dougm: I still like -f better, since 'events' is already similar to the 'tail' command.  We use the same -n option as tail.  Thoughts?

govc/events/command.go
```diff
@@ -54,6 +59,28 @@ func (cmd *events) Process(ctx context.Context) error {
 	return nil
 }
 
+func (cmd *events) printEvents(ctx context.Context, page []types.BaseEvent, m *event.Manager) error {
```
dougm: Might be useful to move this to something like:
```go
func (m *event.Manager) Write(ctx context.Context, page []types.BaseEvent, w io.Writer) error
```
Then the same helper can be used outside of govc to write events to any io.Writer (Stdout, File, etc).

dougm:
Looks and works great @cgtexmex

event/helper.go
cgtexmex: If you're saying "should there be an error check after the loop?" then yes agree...

govc/events/command.go
cgtexmex: I'm fine either way -- the reason I went with -t is because -f IMO is more closely associated with file -- so if we wanted push events to a file -- yeah we could pipe that, but that's where my mind was...your call tough guy

govc/events/command.go
cgtexmex: agreed - how about we cover that in a new issue?

event/helper.go
cgtexmex: Agreed

event/helper.go
cgtexmex: FWIW no additional check just returning the result of property.Wait -- shall either be nil or an error...

dougm: No, I meant it looks like !tail check exits the loop early, before you're iterated over all items in 'pc'.

govc/events/command.go
dougm: ok, let's go with '-f' then.  I don't expect we'd add a file output option for this command, if we did then it would be '-o' (go build, curl and strace come to mind as examples using -o).

event/helper.go
cgtexmex: I never saw that in testing (exiting the loop early), but after reviewing the code there as an issue waiting to happen...specifically there's no need to count the objects in the multiple method...we're either tailing or we aren't...anyhoo...moved the check outside of the loop and simplified a bit

govc/events/command.go
cgtexmex: -f this....done ;-)

dougm:
Looks good Clint.  Not sure why Drone CI job is failing, there's no console output.  I tried restarting it too.  Maybe @casualjim can take a look?

dougm:
Ran the test locally, this one fails:
```
not ok 28 events vm
# (in test file ./events.bats, line 48)
#   `[ ${#lines[@]} -gt $nevents ]' failed
```

Looks like multiple objects is not working, for example, only seeing events for the first vm with this command:
```
govc events "vm/*" "host/*"                                                                                                                                                                                                                                     
[Tue Feb 16 09:35:45 2016] [info] Creating govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F on host esxbox.localdomain in ha-datacenter
[Tue Feb 16 09:35:45 2016] [info] Created virtual machine govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F on esxbox.localdomain in ha-datacenter
```

cgtexmex:
Hmmm....that one should have failed all along then.  Apologies that I missed it.  The code assumes that multiple objects will be the of the same type -- VM, Host, etc...It's the last comment right before the WaitForView function.  I had planned on adding that support later.

I can either take the time and fix that now or add it as another issue later.

dougm:
Sure, fine with me if multiple objects need to be the same type.  If we can get the existing test to pass (it is just using vm types), that'd be good for now.

dougm:
Here's what I'm seeing.. 

```
# 2 vms in my inventory
% govc ls vm
/ha-datacenter/vm/govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F
/ha-datacenter/vm/govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F-2

# only events for the first one are reported, no events for the vm ending in '-2':
% govc events $(govc ls vm)
[Tue Feb 16 09:35:45 2016] [info] Creating govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F on host esxbox.localdomain in ha-datacenter
[Tue Feb 16 09:35:45 2016] [info] Created virtual machine govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F on esxbox.localdomain in ha-datacenter

# the above command should output 4 events, 2 for each vm.  We can see those 4 events when giving govc events a single vm:
% govc events /ha-datacenter/vm/govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F
[Tue Feb 16 09:35:45 2016] [info] Creating govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F on host esxbox.localdomain in ha-datacenter
[Tue Feb 16 09:35:45 2016] [info] Created virtual machine govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F on esxbox.localdomain in ha-datacenter

% govc events /ha-datacenter/vm/govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F-2
[Tue Feb 16 09:36:02 2016] [info] Creating govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F-2 on host esxbox.localdomain in ha-datacenter
[Tue Feb 16 09:36:02 2016] [info] Created virtual machine govc-test-E758DF3C-4D1F-4FAB-BA94-43655831F90F-2 on esxbox.localdomain in ha-datacenter
```

dougm:
Used the following to create those vms:
```
% . govc/test/test_helper.bash
% vm=$(new_id) 
% govc vm.create -on=false $vm
% govc vm.create -on=false "${vm}-2"
```

cgtexmex:
I was able to recreate the issue locally and it was a bug in my code.  Go figure...

In a late change yesterday I removed the count from the multiple object (WaitForView) as I didn't think it was needed...it is needed.  It's passing locally let's see if drone cooperates

view/manager.go
```diff
@@ -0,0 +1,54 @@
...
+package view
+
+import (
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/methods"
+	"github.com/vmware/govmomi/vim25/types"
+	"golang.org/x/net/context"
+)
+
+type Manager struct {
```
dougm: Missed this the first time, we should be using object.Common here (there are probably still a few types that haven't been converted to use object.Common)
```go
type Manager struct {
	object.Common
}

func NewManager(c *vim25.Client) *Manager {
	m := Manager{
		object.NewCommon(c, *c.ServiceContent.ViewManager),
	}

	return &m
}
```

cgtexmex: converted the new view/manager and refactored the existing event/manager

dougm:
Awesome, LGTM

PR #469, opened 2016-03-16
Added new -dump output flag for pretty printing underlying objects using davecgh/go-spew

eyz:
I have found this pretty-printing library (davecgh/go-spew) very useful for discovering underlying data structures when working on new features in govc, and present this new output -dump flag code if there is any interest in inclusion.

Since it's an external library and not included with Go, I suspect there may be a policy on what is allowed to be included in govc. If there is a document on such a policy then I would be interested in reading it for future reference.

Please feel free to close this pull request if there is no interest, or this cannot be merged.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -7,6 +7,7 @@ build:
   commands:
     - go get golang.org/x/tools/cmd/vet
     - go get golang.org/x/tools/cmd/goimports
+    - go get github.com/davecgh/go-spew/spew
     - go get
     - make all
     - govc/build.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ go: 1.4
 before_install:
   - go get golang.org/x/tools/cmd/vet
   - go get golang.org/x/tools/cmd/goimports
+  - go get github.com/davecgh/go-spew/spew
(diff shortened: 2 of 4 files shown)


eyz:
It looks like the build doesn't include outside libraries, per -

govc/flags/output.go:32:2: cannot find package "github.com/davecgh/go-spew/spew" in any of:
	/home/travis/.gimme/versions/go1.4.linux.amd64/src/github.com/davecgh/go-spew/spew (from $GOROOT)
	/home/travis/gopath/src/github.com/davecgh/go-spew/spew (from $GOPATH)

I am closing this. Let me know if you would like it re-opened.

dougm:
We only have 1 external dependency at the moment (golang.org/x/net/context), that should get installed by 'go get' (part of the Makefile test target).  A -dump option would be useful.  I've used github.com/kr/pretty in the past (don't think go-spew was around then), but just did one-offs as needed.

eyz:
OK, re-opening. It didn't pass the build last time, so not sure what needs to be added to pull in the davecgh/go-spew

dougm:
Tried it out, very cool, thanks @eyz

PR #474, opened 2016-03-19
Handle Creation of vm in a specific folder

cblomart:
We arrange vm in folders in vCenter, this allows to create vm in folders by adding an optional folder option to the vm.create.


cblomart:
Anything wrong the $GOVC_TEST_URL from the drone build???

dougm:
@cblomart thanks, looks great.  Can you rebase against vmware/master to pick up change 6e991fc ?  This should make CI happy again.

cblomart:
Any idea from where does the errors in Drone comes?? Seems like the files from the created test virtual machine are not present.

dougm:
There is some more Drone CI related work going on (by @jzt ), not sure yet if this is related.

jzt:
@cblomart At what point do this test VM and its associated files get created? I was under the impression that everything that gets created during testing gets removed when the tests complete. There were ~37 orphaned VMs on the ESX after a series of govc test failures, so I manually deleted them. I have a feeling that the files that are now missing were included (the apparent source of pain for these recent failed builds). I'm a little surprised though, as I did not expect that there were any permanent files on the ESX required for successful testing.

How can I help fix this?

cblomart:
from the tests:
```
[...]
not ok 23 network vm list
# (from function `teardown' in file govc/test/test_helper.bash, line 40,
#  from function `bats_teardown_trap' in file /usr/local/libexec/bats-exec-test, line 230,
#  from function `main' in test file /usr/local/libexec/bats-exec-test, line 1)
#   `#!/usr/bin/env bash' failed with status 123
# Powering on {VirtualMachine 895}... OK
# Powering on {VirtualMachine 896}... OK
# govc: vm '/ha-datacenter/vm/vmfs/volumes/56e33959-0d4d2de8-ca00-0050561d658a/govc-test-03bc809e-0263-4f1a-94ef-df06604a134c/govc-test-03bc809e-0263-4f1a-94ef-df06604a134c.vmx' not found
# govc: Error caused by file /vmfs/volumes/56e33959-0d4d2de8-ca00-0050561d658a/govc-test-0e0cf046-4c1f-4408-9e74-3b37410d6d5e
# govc: Error caused by file /vmfs/volumes/56e33959-0d4d2de8-ca00-0050561d658a/govc-test-e0f84cc5-8f68-404e-8984-2d9e76d32963
[...]
```

logs are not so clear to me...

I see an error from the teardown function teardown which seems to destroy everything.
then a few poweron
then errors from files... naturaly the exact error is not mentioned... except for the first one where it simply doesn't find the vmx.

My interpretations of this as i am not used to bats is that the first few lines concernt the error seens by bats so teardown line 40:
```
teardown() {
  govc ls vm | grep govc-test- | $xargs -r govc vm.destroy
  govc datastore.ls | grep govc-test- | awk '{print ($NF)}' | $xargs -n1 -r govc datastore.rm
}
```
so the error is in the removal of found files...
Are vms properly stopped before removal???
I just pushed an update in that sense.

I discover thing a bit... so you have an esx for the tests. Debuging tests in this case is a bit difficult (do we have any way to know what is lingering in this esx?)

cblomart:
And now:
govc: Post https://192.168.109.3/sdk: dial tcp 192.168.109.3:443: getsockopt: no route to host

jzt:
@cblomart hmm.. from your last comment, it looks like you are using an old secrets file (which is referencing the old IP of the ESX). A rebase with master should fix that, but I'm still looking into the problems relating to missing files.

cblomart:
I rebased yesterday or so O_o

jzt:
@cblomart Regarding test debugging - Unfortunately seeing what is lingering on the ESX is not currently possible. However, the lingering VM problem is not common - it has actually not been a problem except for the single time that I had to go in and clean things up.

Regarding the rebase, looking at your change set, the `.drone.sec` file is definitely not the same one in master, so I would recommend reverting it and retrying the build.

cblomart:
sure it is not the same one as in master now??? i might have made a mistake somwhere (thanks for your patiente)

jzt:
@cblomart Sorry for the late reply. It definitely appears different than master's version. It shows up in the diff under files changed in this PR.

jzt:
@cblomart @eyz @dougm @cgtexmex I figured it out. I had previously regenerated the secrets file (2 days ago) to fix the build after changes were made to `.drone.yml`. I mistakenly thought that the IP address of the ESX changed before that, and was under the assumption that `.drone.sec` was up-to-date as well. I am putting up a PR now with updated an updated `.drone.sec`, which should fix the build going forward (at least with regard to being able to connect to the ESX).

Sorry for the confusion/inconvenience.

cblomart:
Ok seems better for the tests.

Just a problem with the pool destroy test:
```
not ok 60 pool.destroy
# (in test file govc/test/pool.bats, line 115)
#   `[ $result -eq 0 ]' failed
```
Looking at the bats file:
```
  # should not be any existing test pools
  result=$(govc ls "host/*/Resources/govc-test-*" | wc -l)
  [ $result -eq 0 ]
```
So there are leftover pools from somewhere...

PS: this way of testing forbids parallel build/test, is it intended?

cblomart:
Reading the docs concerning ressource pool deletion:
http://pubs.vmware.com/vsphere-60/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc%2Fvim.ResourcePool.html
 
  - In the description:

"When a ResourcePool is destroyed, all the virtual machines are reassigned to its parent pool."

  - the DestroyChildren method:

"Removes all child resource pools recursively. All virtual machines and vApps associated with the child resource pools get associated with this resource pool."

  - the Destroy_Task (inherited from ManagedObject):

"Destroys this object, **deleting its contents** and removing it from its parent folder (if any)."

The way i read it the recursive parameter of pool is not needed. These actions are differents:
  - destroy: destroy the current pool and places all its dependent vms and vapps in the parent pool.
  - destroychildren: remove all child resource pools and places the dependent vms and apps in the current pool

==> the action is always recursive.

I tested on a spare ESXi (5.5. u2; and yes, sorry i test on windows in this case)
```
C:\Users\cblomart>govc ls host/*/Resources/*

C:\Users\cblomart>govc pool.create */Resources/govc-test-1

C:\Users\cblomart>govc ls host/*/Resources/*
/ha-datacenter/host/*********/Resources/govc-test-1

C:\Users\cblomart>govc pool.create */Resources/govc-test-1/a

C:\Users\cblomart>govc pool.create */Resources/govc-test-1/b

C:\Users\cblomart>govc ls host/*/Resources/*/*
/ha-datacenter/host/*********/Resources/govc-test-1/a
/ha-datacenter/host/*********/Resources/govc-test-1/b
C:\Users\cblomart>govc pool.destroy */Resources/govc-test-1

C:\Users\cblomart>govc ls host/*/Resources/*

````





I will look at the pool tests to try and simplify the test in this

dougm:
@cblomart tested locally, looks good.  I did some cleanup of the git history, testing against Drone now.

dougm:
@cblomart thanks for bearing with us through the CI issues, I've merged your work in #480

PR #478, opened 2016-03-26
Update test ESX IP in Drone secrets file

jzt:
This fixes the "no route to host" problems when testing govc
with Drone.


dougm:
1 test failed, but looks like it may just be a pool that didn't get cleaned up.

PR #471, opened 2016-03-17
Added dvs.portgroup.info command to query for distributed portgroup ports using criteria from arguments

eyz:
Arguments:

- -active (Filter by port active or inactive status)
- -connected (Filter by port connected or disconnected status)
- -count (Number of matches to return; 0 = unlimited)
- -dvpg (Distributed Virtual Portgroup inventory path)
- -dvs (Distributed Virtual Switch inventory path; required)
- -inside (Filter by port inside or outside status)
- -vlanId (Filter by VLAN ID; 0 = unfiltered)

PR #481, opened 2016-03-29
Add dvs.portgroup.info command

dougm:
Cleaned up git history of PR #471 
Includes a few other fixes, using finder instead of search index, making golint happy, etc.

### DIFF
--- a/govc/CHANGELOG.md
+++ b/govc/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### 0.5.0 (unreleased)
 
+* Add dvs.portgroup.info command
+
 * Add `-folder` flag to vm.create command
 
 * Add `-dump` flag to OutputFlag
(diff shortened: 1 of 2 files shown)


eyz:
Thank you for the code review. I have incorporated your changes into my branch. If you would like to give me credit on the pull request, then I can another pull request from my repo, eyz/govmomi@5ea9ce2daf216366a33c9b9b29c046fce6fa1789. I had originally used the Finder as well, but found the SearchIndex to be more direct.

dougm:
@eyz regarding the finder vs search index, you can still using the full inventory path with finder.  And in general if there is a required argument, we use flagset.Args rather than flagset.*Var.  Thanks for reviewing and testing my review, it is easier that way sometimes - but credit in git history will always go to the submitter, thanks again!

eyz:
That makes sense about the required argument. Thanks for explaining the pattern.

PR #484, opened 2016-03-29
Generated code updates and fixes

dougm:


dougm:
@caglar10ur can you review?

govc/flags/int32.go
```diff
@@ -0,0 +1,43 @@
...
+	*i = int32Value(v)
+	return err
+}
+
+func (i *int32Value) Get() interface{} {
+	return int32(*i)
+}
+
+func (i *int32Value) String() string {
+	return fmt.Sprint(*i)
```
caglar10ur: [minor] Std lib seems to be using `return fmt.Sprintf("%v", *i)` - https://golang.org/src/flag/flag.go#L123

dougm: Ah, I was looking in a different part of stdlib: https://github.com/golang/go/blob/master/src/cmd/internal/obj/flag.go#L91

PR #489, opened 2016-04-02
device.remove: add keep option

ignatov:
The option allow you unmount the disk from the VM and store disk files in data store.


dougm:
Thanks @ignatov

PR #493, opened 2016-04-05
Leave AddressType empty in EthernetCardTypes

### LINKED ISSUE #492: invalid MAC addressType after vm.network.change
Hello,

Context : Vsphere6, DVS

Reproductible : yes

When using  govc vm.network.change on vm created by govc, it changes the type from
"Assigned" to "Generated".

The MAC address is not changed so you pass your tests, but I you power on/off the VM,
it's MAC address changed to be within the range of "Generated" MAC address.

Example for the same VM :

After creation

---
MAC Address:      00:50:56:xx:yy:zz
  Address type:     assigned
---

After network.change on DVS

---
MAC Address:      00:50:56:xx:yy:zz
  Address type:     generated
---

After power on :

---
MAC Address:      00:0c:29:uu:vv:ww
  Address type:     generated
---

I didn't dig if it's govc or govmomi related, tell me if I could help.

Regards,

dougm:
VirtualEthernetCard.AddressType is an optional field, no need to set it
to generated in this helper.

Fixes #492

### DIFF
--- a/object/virtual_device_list.go
+++ b/object/virtual_device_list.go
@@ -63,7 +63,6 @@ func EthernetCardTypes() VirtualDeviceList {
 		&types.VirtualVmxnet3{},
 	}).Select(func(device types.BaseVirtualDevice) bool {
 		c := device.(types.BaseVirtualEthernetCard).GetVirtualEthernetCard()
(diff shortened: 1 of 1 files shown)


lsioctl:
Hi, It works like a charm, thanks for the quick and effective fix ! Regards.

PR #506, opened 2016-05-05
Added Remove All Snapshot task to Virtual Machine

austinlparker:
This commit adds a call to the RemoveAllSnapshots task for a given VM, similar to what can be accomplished in pyvmomi.


dougm:
👍 thanks @austinlparker

PR #507, opened 2016-05-06
Traverse folders in vmlist

austinlparker:
This is a resubmit of #328 as I could really use that functionality in master and the original contributor seems to not be around any more. I was able to test locally that it did indeed work, although it takes some time to to traverse the folders (for instance, on a vCenter instance with about 600 VMs and quite a few folders, it took 3 minutes to traverse the folder structure and then inspect 3 VMs snapshot status)

### DIFF
--- a/find/finder.go
+++ b/find/finder.go
@@ -703,6 +703,8 @@ func (f *Finder) VirtualMachineList(ctx context.Context, path string) ([]*object
 			vm := object.NewVirtualMachine(f.client, o.Reference())
 			vm.InventoryPath = e.Path
 			vms = append(vms, vm)
+		case mo.Folder:
+			f.findVirtualMachineListForFolder(ctx, e.Object, &vms)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


austinlparker:
Actually, turns out my filter is just garbage - it only took around 40s to traverse all the folders and build the VirtualMachineList

dougm:
@austinlparker thanks, I keep meaning to revisit this.  It should be possible to compose the PropertyCollector query such that recursion happens on the server (VC) side.  When this code was originally written we were only hitting ESX directly.  I've been spending some time with the PropertyCollector recently, so I'll try to get this sorted out soon.

austinlparker:
@dougm Do you have any sort of ETA? I can use my fork in the meantime - I tested this out again today when the app was sitting behind the firewall and it ran much quicker. When I was testing last week, I was running over VPN through a spotty cell connection. It's certainly not much slower than what I'm using in pyvmomi.

dougm:
@austinlparker hopefully later this week or early next

ErezHorev:
+1
Need this as well

austinlparker:
@dougm Any update here?

dougm:
@austinlparker apologies for the long delay.  This is very high on my list, but other things keep getting in the way.  I still want to look at supporting recursion for types other than VirtualMachine, have an option to enable/disable, tests, etc.

austinlparker:
Thanks for letting me know! If there's anything I can specifically do to help, mention me and I can take a crack at implementing some stuff to help out.

dougm:
Still haven't gotten to this yet, but getting closer.

iamthemuffinman:
@dougm I could also use this. Have you made any progress?

dougm:
Apologies again for the long delay.  Please review #663 - the change applies the same behavior to all resource types, rather than just the VirtualMachine type in this PR.

PR #508, opened 2016-05-10
SOAP decoding fixes

dougm:
These are for decoding responses from the in-progress govmomi simulator.  The proper fixes will be more involved, but these small changes have no impact on the non-simulator paths.


caglar10ur:
LGTM :sheep: :it:

PR #509, opened 2016-05-10
Get complete file details in Datastore.Stat

dougm:
This helper was originally optimized within govc to just check if a file
exists.  It isn't much more overhead to get the additional details for a
single file, add the full set for convenience.


caglar10ur:
LGTM :sheep: :it:

PR #514, opened 2016-05-27
Support mkdir/rm of namespace on vsan

fdawg4l:
VSAN uses uuids to refer to dirents.  Being able to use them directly vs
using their display names is helpful.

### DIFF
--- a/govc/datastore/mkdir.go
+++ b/govc/datastore/mkdir.go
@@ -19,6 +19,7 @@ package datastore
 import (
 	"errors"
 	"flag"
+	"fmt"
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
@@ -32,6 +33,7 @@ type mkdir struct {
 	*flags.DatastoreFlag
 
 	createParents bool
+	isNamespace   bool
 }
 
 func init() {
@@ -43,6 +45,7 @@ func (cmd *mkdir) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.DatastoreFlag.Register(ctx, f)
 
 	f.BoolVar(&cmd.createParents, "p", false, "Create intermediate directories as needed")
+	f.BoolVar(&cmd.isNamespace, "namespace", false, "Return uuid of namespace created on vsan datastore")
 }
 
 func (cmd *mkdir) Process(ctx context.Context) error {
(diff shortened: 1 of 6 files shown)


govc/datastore/mkdir.go
```diff
@@ -67,26 +70,46 @@ func (cmd *mkdir) Run(ctx context.Context, f *flag.FlagSet) error {
...
-		return err
-	}
+		if ds, err = cmd.Datastore(); err != nil {
+			return err
+		}
+
+		path := args[0]
+
+		nm := object.NewDatastoreNamespaceManager(c)
+		if uuid, err = nm.CreateDirectory(context.TODO(), ds, path, ""); err != nil {
```
dougm: Use the Context param passed Process() instead of TODO

govc/datastore/mkdir.go
```diff
@@ -67,26 +70,46 @@ func (cmd *mkdir) Run(ctx context.Context, f *flag.FlagSet) error {
...
+			return err
+		}
+
+		fmt.Printf("  UUID:\t%s\n", uuid)
```
dougm: Since uuid is the only output, let's just make it:
```go
fmt.Println(uuid)
```

govc/datastore/rm.go
```diff
@@ -67,24 +70,39 @@ func (cmd *rm) Run(ctx context.Context, f *flag.FlagSet) error {
...
+		path := args[0]
 
-	m := object.NewFileManager(c)
-	task, err := m.DeleteDatastoreFile(context.TODO(), path, dc)
-	if err != nil {
-		return err
+		nm := object.NewDatastoreNamespaceManager(c)
+		err = nm.DeleteDirectory(context.TODO(), dc, path)
+		if types.IsFileNotFound(err) {
+			fmt.Println("yes")
```
dougm: some debug left in?

govc/test/datastore.bats
```diff
@@ -73,3 +73,34 @@ upload_file() {
...
+
+
+@test "datastore.mkdir" {
+
+  # Not supported datastore type is a failure
+  run govc datastore.mkdir -uuid "notfound"
+  assert_failure
+  assert_matches "govc: ServerFaultCode: Cannot complete file creation operation." "${output}"
+
+  run govc datastore.mkdir "dir"
```
dougm: Use `$(new_id)` instead of "dir".

object/namespace_manager.go
```diff
@@ -0,0 +1,105 @@
...
+		DatastorePath: datastorePath,
+	}
+
+	if dc != nil {
+		ref := dc.Reference()
+		req.Datacenter = &ref
+	}
+
+	_, err := methods.DeleteDirectory(ctx, nm.c, req)
+	// If this is a FileFault or a FileNotFoundFault, more relevant fault
```
dougm: As we discussed in person, just return err here and let's leave it to the caller to dig out the useful info.  We can add helper(s) outside of this method for use in govc and elsewhere.

fdawg4l:
Sounds good.  Will take care of all of these.  The `DatastoreNamepsaceManger` creates and removes namespaces (represented by top level directories) only.  I'll update the flag to `-namespace` in `govc` rather than `-uuid`.  I didn't realize children of the namspaces are just regular directories.  Only top most namespaces have special handling.

dougm:
lgtm - bonus points if you add changelog entries

PR #544, opened 2016-07-13
Fixes panic in govc events

### LINKED ISSUE #543: panic in govc events
When a task event doesn't have a defined Entity object the routine will panic due to a nil reference.

A simple nil check needs to be added before this [call](https://github.com/vmware/govmomi/blob/master/govc/events/command.go#L75)

cgtexmex:
Fixes #543


caglar10ur:
LGTM :sheep: :it:

dougm:
lgtm

PR #546, opened 2016-07-15
Populate network mapping from ovf envelope

egorovv:
On OVF import bind network adapters to the networks specified in the
OVF envelope if they exist on the target.


dougm:
thanks @egorovv

PR #547, opened 2016-07-18
vendor golang.org/x/net/context

caglar10ur:


dougm:
Thanks @caglar10ur .  Looks like we'd still need to go install goimports for CI to be happy?

caglar10ur:
amended last commit to do ```go install ./vendor/golang.org/x/tools/cmd/goimports```

dougm:
@caglar10ur on second thought, after seeing the 10k line golang.org/x/tools/imports/zstdlib.go - how about if we just vendor the packages needed to compile (context and go-spew).  I think it'd still be ok to use 'go get' for goimports in CI.

caglar10ur:
Dropped goimports

PR #555, opened 2016-08-04
govc - popualte 'Path' fiels in xxx.info output

egorovv:
Common InventoryPath field is not initialized in the new objects
returned by SearchIndex - for now do it in the SearchFlag


dougm:
:+1:

PR #558, opened 2016-08-10
Fix deletion of powered off vApp

caglar10ur:
Right now trying to delete a vApp with no VMs or a powered off vApp ends
up with

The attempted operation cannot be performed in the current state (Powered off).

errors.


dougm:
lgtm

PR #559, opened 2016-08-16
Bump to 1.7 and start using new context pkg

caglar10ur:
This is here for a RFC, still need to re-generate .drone.sec


dougm:
lgtm

We are currently unplugged from drone, would be fine to update .drone.sec later and get that build going again.

caglar10ur:
@dougm ok did the cosmetic change too. I think this is ready to merge

dougm:
:sheep: :it:

PR #570, opened 2016-08-27
Add device.usb.add command

### LINKED ISSUE #569: govm vm.change to add a USB controller
I can't find a way to add a USB controller to a new VM, either with `vm.create` or `vm.change`.

The source suggests this isn't implemented yet?

```
$ git grep -c -E 'usb|USB|Usb' 
vim25/types/enum.go:28
vim25/types/types.go:41
```

... only the auto-generated code mentions USB.

I can create OS X VMs with vm.create, but then I can't move the mouse around in them, since they lack USB.

If I create VMs with the web UI, then I can add a USB controller and move the mouse & type and such.

dougm:
Supports adding USB controllers to VMs.

Fixes #569

### DIFF
--- a/govc/device/usb/add.go
+++ b/govc/device/usb/add.go
@@ -1,0 +1,105 @@
+/*
+Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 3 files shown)


dougm:
Haven't tested yet other than the change to device.bats, but something along these lines should do the trick @bradfitz

bradfitz:
Thanks! Looks perfect. I'll try this out today.

caglar10ur:
LGTM

bradfitz:
And I confirm it worked for me.

I now have new problems, but that's good. :-)

Thanks again.

PR #576, opened 2016-08-31
Add VirtualMachine.Unregister func

jzt:
This is a simple addition to object/virtual_machine.go to to allow for unregistering a VM.


dougm:
👍

PR #582, opened 2016-09-02
Checking result of reflect.TypeOf is not nil before continuing

dagnello:
Adding defensive checks against nil result from reflect.TypeOf before dereferencing the returned type.


dougm:
:+1:

PR #586, opened 2016-09-06
Add vm.migrate command

dougm:


dougm:
cc @mhagen-vmware

mhagen-vmware:
Awesome, thanks @dougm !

govc/test/vm.bats
```diff
@@ -464,6 +464,19 @@ load test_helper
   assert_failure
 }
 
+@test "vm.migrate" {
+  vcsim_env
+  vm=$(new_empty_vm)
+
+  # migrate from H0 to H1
+  run govc vm.migrate -host DC0_C0/DC0_C0_H1 "$vm"
```
mhagen-vmware: this might be a bit unclear, should we change this parameter to -targetHost or something as all the other commands reference the source host when using that parameter?

dougm: Every command using flags.HostSystemFlag gets the hardcoded name "host".  Other examples where host is a "target" include vm.clone, import.ova, vm.markasvm.  The SDK fields/params are also named "host" these cases.
See MigrateVM_Task for example: http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.VirtualMachine.html#migrate

PR #587, opened 2016-09-07
Add DatastoreFile helpers

dougm:
Supports tail and follow of a datastore file over HTTP.


dougm:
reviewed with @hickeng in person

PR #593, opened 2016-09-16
Make finder.DatacenterList recurse top-level folders

dannc:
finder.go's function DatacenterList(...) would only look for a
Datacenter directly under the rootFolder when using the wildcard
path ("*").  When there is instead a user-defined Folder (or Folders) in
which the Datacanter object is contained, we need to look through those
top-level folders to find the Datacenter objects.

Notably, the finder.DefaultDatacenter function was impacted as it would
only search for Datacenters under the top-level folder.

This patch fixes that issue by recursing down through any top-level
folders if present.

### DIFF
--- a/find/finder.go
+++ b/find/finder.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"path"
+	"strings"
 
 	"github.com/vmware/govmomi/list"
 	"github.com/vmware/govmomi/object"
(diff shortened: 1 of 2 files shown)


dannc:
Anyone have a chance to review this yet?

dougm:
Thanks @dannc , haven't had a change to review yet.  I plan to spend some time on this and the other folder related issues in the next week or so.

dougm:
Apologies for the long delay.  Please review #663 - the change applies the same behavior to all resource types, rather than just the Datastore type in this PR.

PR #598, opened 2016-09-23
Treat DatastoreFile follower Close as "stop"

dougm:
Rather than close right away, wake up and return io.EOF after 1 final
Read().  The main point of follow.DatastoreFile.Close is to break out of
the blocking Read() call.  We still do that, but also give the
caller any other data that may have come in while we were waiting on the
poll interval ticker.


fdawg4l:
LGTM

PR #618, opened 2016-10-25
Add object destroy and rename commands

dougm:
- Remove datacenter.destroy command (use object.destroy instead)

- Remove folder.destroy command (use object.destroy instead)

- Rename folder.move_into -> object.mv


govc/flags/datacenter.go
```diff
@@ -150,6 +150,10 @@ func (flag *DatacenterFlag) ManagedObjects(ctx context.Context, args []string) (
 			return nil, err
 		}
 
+		if len(elements) == 0 {
+			return nil, fmt.Errorf("object '%s' not found", arg)
```
caglar10ur: %q maybe?

dougm: I did think about that :)  Would have used find.NotFoundError here, but the fields are not exported.. decided to just copy-n-paste keeping the same format.

PR #623, opened 2016-11-02
Avoid possible NPE in VirtualMachine.Device method

dougm:


object/virtual_machine.go
```diff
@@ -302,11 +302,21 @@ func (v VirtualMachine) WaitForNetIP(ctx context.Context, v4 bool) (map[string][
...
 		return nil, err
 	}
 
+	// Quoting the SDK doc:
+	//   The virtual machine configuration is not guaranteed to be available.
+	//   For example, the configuration information would be unavailable if the server
+	//   is unable to access the virtual machine files on disk, and is often also unavailable
+	//   during the initial phases of virtual machine creation.
+	if o.Config == nil {
+		return nil, fmt.Errorf("%s Config is not available, connectionState=%s",
```
emlin: Doug, is it possible to return an InvalidState error here?

dougm: I had considered that, but it is a bit awkward to manually construct the fault structures that normally come over the wire, and the soap/vim wrapping.  Since we don't do that elsewhere, decided to keep it a simple error.  And since the Device() method is itself a simple few line helper, we can just implement our own in vic to handle any special cases.

PR #624, opened 2016-11-04
Allow DatastoreFile follower to drain current body

dougm:
If the caller's buffer size is smaller than remaining request body, the
caller would miss data.

Return io.EOF if the file has gone away when we wake up in the poll
loop.

See PR #598


caglar10ur:
Lgtmy

PR #626, opened 2016-11-08
Add -product-version flag to dvs.create

dougm:
If DVS is created without this flag, only hosts with version >= the VC
version can be added to the DVS by default.


caglar10ur:
LGTM

rajanashok:
LGTM

PR #631, opened 2016-11-29
Add basic VirtualNVMEController support

caglar10ur:
Requires vim-version set to 6.5 either via vim-version flag or via
GOVC_VIM_VERSION environment variable. Eg;

```
govc vm.create -vim-version=6.5 -on=false -ds=SSD -net "VM Network" -m 2048 -c 2 -g "vmkernel65Guest" -net.adapter=vmxnet3 -disk.controller nvme esxi
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # changelog
 
+### unreleased (N/A
+
+* Update to vim25/6.5 API
+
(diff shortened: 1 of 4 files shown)


dougm:
nice, lgtm

Changelog(s) updates for this and the 6.5 re-gen would be great too :)

caglar10ur:
Refreshed with Changelog entries

PR #634, opened 2016-11-30
Attach context to HTTP requests for cancellations

henrikhodne:
`http.Transport.CancelRequest` is deprecated, and `http.Request.WithContext` should be used instead.

The motivation for this change is that there appears to be a connection leak somewhere in govmomi where it doesn't close response bodies, leaving connections and goroutines leaked. We're running a long-running server that interacts with vSphere fairly regularly, and whenever we have requests that get their contexts cancelled, the goroutine count goes up and the number of open TCP connections to our vSphere API goes up. After digging through the code, I found that this is the only place where HTTP requests are made, and as far as I can see if `soap.Client.do` returns a response, then `soap.Client.Do` will `defer res.Body.Close()`. So, then my guess is that `http.Client.Do` returns a response, but `soap.Client.do` doesn't return it, which as far as I can see is only possible when the `c.t.CancelRequest` branch is called.

My hope is that this change will keep the cancellation functionality while also removing a possibility for a connection to leak if the request _is_ cancelled.

### DIFF
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -392,29 +392,7 @@ func (c *Client) do(ctx context.Context, req *http.Request) (*http.Response, err
 		return c.Client.Do(req)
 	}
 
-	var resc = make(chan *http.Response, 1)
-	var errc = make(chan error, 1)
-
-	// Perform request from separate routine.
-	go func() {
-		res, err := c.Client.Do(req)
-		if err != nil {
-			errc <- err
-		} else {
-			resc <- res
-		}
-	}()
-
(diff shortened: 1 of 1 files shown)


henrikhodne:
I've deployed this branch in our service and so far things look promising, but the resource leak can sometimes be a bit slow, I should know with more certainty later today if this solves the leak or not.

henrikhodne:
After running this today with high load and lots of context cancellations I'm not seeing any goroutine leaks anymore, so looks like this fixed it.

dougm:
Thanks @henrikhodne , looks great.

PR #635, opened 2016-12-01
Disable use of service ticket for datastore HTTP access by default

### LINKED ISSUE #594: Datastore.URL(...) returns incorrect url when Datacenter is in a folder
I have a vsphere environment where our Datacenter is under a Folder (not the root.Folder).  In my case, the Datacenter is at "Colorado Springs/Resource Pools".

When using the docker-machine vmwarevsphere driver to create a machine, the driver fails when trying to upload the boot2docker.iso file to the Datastore.  I tracked this down to the Datastore.URL(...) method returning a url with a query parameter:

```
 dcPath=Resource+Pools
```

rather than:

```
dcPath=Colorado+Springs%2FResource+Pools
```

I modified the govmomi examples/datastores/main.go to print out the url returned just to be sure that it wasn't in the vmwarevsphere docker-machine driver:

```
./datastores -url https://vsphereserver/sdk -insecure y
URL for Datastore 'Dynamic': https://vsphereserver/folder/%2A?dcPath=Resource+Pools&dsName=Dynamic
URL for Datastore 'Platform_dev': https://vsphereserver/folder/%2A?dcPath=Resource+Pools&dsName=Platform_dev
URL for Datastore 'Production': https://vsphereserver/folder/%2A?dcPath=Resource+Pools&dsName=Production
Name:         Type:  Capacity:  Free:
Dynamic       NFS    6.3TB      6.3TB  
Production    NFS    7.3TB      7.3TB  
Platform_dev  NFS    4.0TB      4.0TB
```

The dcPath component of the URL should be prepending the parent folder(s) to the Datacenter name in order for the upload to work correctly.

dougm:
Can still be enabled by setting GOVMOMI_USE_SERVICE_TICKET=true
as a GOVC_URL query param or environment variable.

- Only use service tickets when connected to VC

- Add Datastore.HostContext helper for cases where datastore files
  can only be read by the host where the VM is running

- Prefer host management IP to host inventory name

- Add -host option to govc datastore.download and datastore.tail to optionally
  specify the host owning the lock for a VM file open for writing (e.g. vmware.log)

- Collapse some common code code for constructing the URL

- Use the full inventory path for the dcPath parameter (Fixes #594)


govc/datastore/download.go
```diff
@@ -73,6 +81,18 @@ func (cmd *download) Run(ctx context.Context, f *flag.FlagSet) error {
 		return err
 	}
 
+	var via string
```
anchal-agrawal: Small readability note: can we move `via` closer to where it's used?

govc/datastore/download.go
```diff
@@ -39,12 +41,18 @@ func init() {
 func (cmd *download) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.DatastoreFlag, ctx = flags.NewDatastoreFlag(ctx)
 	cmd.DatastoreFlag.Register(ctx, f)
+
+	cmd.HostSystemFlag, ctx = flags.NewHostSystemFlag(ctx)
```
vburenin: I wonder why do we pass context and get it back?

govc/datastore/download.go
```diff
@@ -87,8 +107,8 @@ func (cmd *download) Run(ctx context.Context, f *flag.FlagSet) error {
 		return err
 	}
 
-	if cmd.OutputFlag.TTY {
-		logger := cmd.ProgressLogger("Downloading... ")
+	if cmd.DatastoreFlag.OutputFlag.TTY {
+		logger := cmd.DatastoreFlag.ProgressLogger(fmt.Sprintf("Downloading%s... ", via))
```
vburenin: Shouldn't it be something like ": " before %s?

object/datastore.go
```diff
@@ -75,29 +77,24 @@ func (d Datastore) Path(path string) string {
...
-		return nil, err
-	}
-
-	var mds mo.Datastore
-	if err := d.Properties(ctx, d.Reference(), []string{"name"}, &mds); err != nil {
-		return nil, err
-	}
-
+// NewURL constructs a url.URL with the given file path for datastore access over HTTP.
+func (d Datastore) NewURL(path string) *url.URL {
```
vburenin: Why D is not the reference?

object/datastore.go
```diff
@@ -167,43 +165,67 @@ func (d Datastore) useServiceTicketHostName(name string) bool {
...
-		}.Encode(),
-	}
+	u := d.NewURL(path)
 
-	if !d.useServiceTicket() {
-		return u, nil, nil
-	}
+	var host *HostSystem
+
+	h := ctx.Value(datastoreServiceTicketHostKey)
```
vburenin: you could just cast h in place without checking for nil OK: https://play.golang.org/p/T6cKj0hHM3

govc/datastore/download.go
dougm: will do

govc/datastore/download.go
dougm: Default message is just: "Downloading..."
If host is specified, looks like: "Downloading via esx1-host.example.com..."

object/datastore.go
dougm: In general my preference/habit is to use pointer receivers, but the existing wrappers use value.

object/datastore.go
dougm: yes that'll clean this up a bit, I'll do that.

vburenin:
I have no concerns about the correctness of this code. Things looks good to me. There are some cosmetic improvements could be made, but that is it.
LGTM

cgtexmex:
lgtm

hmahmood:
lgtm

find/finder.go
```diff
@@ -312,6 +318,7 @@ func (f *Finder) DatastoreList(ctx context.Context, path string) ([]*object.Data
 		if ref.Type == "Datastore" {
 			ds := object.NewDatastore(f.client, ref)
 			ds.InventoryPath = e.Path
+			ds.DatacenterPath = f.dc.InventoryPath
```
hickeng: is dc guaranteed to be non-nil at this point?

object/datastore.go
```diff
@@ -58,6 +58,8 @@ func (e DatastoreNoSuchFileError) Error() string {
 
 type Datastore struct {
 	Common
+
+	DatacenterPath string
```
hickeng: Given everything lives in the context of a datacenter, should this move into Common eventually?

object/datastore.go
```diff
@@ -167,43 +165,67 @@ func (d Datastore) useServiceTicketHostName(name string) bool {
...
-	// If connected to VC, the ticket request must be for an ESX host.
-	if d.c.IsVC() {
 		hosts, err := d.AttachedHosts(ctx)
 		if err != nil {
 			return nil, nil, err
 		}
 
 		if len(hosts) == 0 {
-			return nil, nil, fmt.Errorf("no hosts attached to datastore %#v", d.Reference())
+			return u, nil, nil
```
hickeng: comment about fallback to VC

object/datastore.go
dougm: Started on that route initially, but opted to keep the diff a bit smaller and ideally there won't be other types that need this path.  Datastore is the only type currently that uses HTTP without the vmodl.

hickeng:
LGTM

find/finder.go
dnaeon: @dougm In one of my projects I'm using absolute path to the datastore, e.g. `/MyDatacenter/datastore/vm-storage01` and use a `Finder.Datastore()` to get the datastore object, but now I get panics because of this.

What would be the recommended way without having to call `SetDatacenter()`, because that would be another parameter I need to add to my project and that's already lots of changes I need to do to adjust the code to the recent change?

Thanks,
Marin

dougm: @dnaeon in that case you don't need Finder, as SearchIndex.FindByInventoryPath does the job with absolute paths.  But I'd be fine with adding the nil check for now.  Are you using any of the Datastore upload/download methods?  Those need the Datacenter ref against VC, unless you set GOVMOMI_USE_SERVICE_TICKET=true

PR #640, opened 2016-12-09
Avoid Finder panic when SetDatacenter is not called

dougm:
PR #635 assumed that Finder would not be used without having called SetDatacenter.
However, when the Datastore methods are used with absolute paths and without SetDatacenter,
would cause a panic.

### DIFF
--- a/find/finder.go
+++ b/find/finder.go
@@ -110,6 +110,35 @@ func (f *Finder) datacenter() (*object.Datacenter, error) {
 	return f.dc, nil
 }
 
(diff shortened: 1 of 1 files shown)


dougm:
cc @dnaeon

dnaeon:
@dougm, thanks for the quick fix! Just tested it and works fine! Hopefully it will be merged soon.

dougm:
@dnaeon great, thanks for testing, merged!

PR #641, opened 2016-12-09
Add storage support to vm.migrate

### LINKED ISSUE #639: Doing a "storage vmotion" without doing a clone?
From looking through the vSphere API docs and the `govmomi` source code, it seems like the only way I can do something like a "storage vmotion" is with `vm.clone`, due to it supporting https://github.com/vmware/govmomi/blob/master/govc/vm/clone.go#L299-L303

However, a `vm.clone` takes a fair bit longer than if I use the **Migrate wizard...** to change the datastore of the VM via storage vmotion, 23 minutes vs 14 minutes.

Is it possible to use `govc` to initiate the same action that done with the **Migrate...** wizard, when choosing **Change datastore**, like what's shown in http://greg.porter.name/wordpress/?p=1692? 

image

dougm:
Fixes #639

### DIFF
--- a/govc/vm/migrate.go
+++ b/govc/vm/migrate.go
@@ -19,19 +19,22 @@ package vm
 import (
 	"context"
 	"flag"
+	"fmt"
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
 type migrate struct {
 	*flags.ResourcePoolFlag
 	*flags.HostSystemFlag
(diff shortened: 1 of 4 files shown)


govc/vm/migrate.go
```diff
@@ -68,10 +76,28 @@ func (cmd *migrate) Usage() string {
...
+
+	logger := cmd.DatastoreFlag.ProgressLogger(fmt.Sprintf("migrating %s... ", vm.Reference()))
+	_, err = task.WaitForResult(ctx, logger)
+	if err != nil {
+		return err
+	}
+
+	logger.Wait()
+
+	return nil
```
solarce: This progress indicator is really nice! 🎉

PR #645, opened 2016-12-13
Add govc-executable-name customization option

tzz:
@dougm I hope this is useful. I needed it because my gopath is (intentionally) not part of my `exec-path`.

### DIFF
--- a/govc/emacs/govc.el
+++ b/govc/emacs/govc.el
@@ -55,6 +55,11 @@
   "Prefix for `govc-mode'."
   :group 'govc)
 
+(defcustom govc-command "govc"
+  "Executable path to the govc utility."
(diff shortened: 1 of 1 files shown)


dougm:
@tzz looks good, happy to add this.  Wondering though, should we name this variable `govc-command` as that seems to be the convention (`diff-command`, `grep-command`, etc).  We could also rename the existing `govc-command` function to `govc-format` to avoid any confusion.

tzz:
That's exactly why I avoided `govc-command` ha ha. I did as you suggested, renaming the function `govc-command` to `govc-format-command` and creating the new defcustom. Change force-pushed.

dougm:
thanks @tzz !

PR #648, opened 2016-12-13
govc.el: auth-source integration

### LINKED ISSUE #647: govc.el should integrate with auth-source
Instead of putting the usernames and passwords in the `govc-urls`, users should be able to use the auth-source library to put them in `~/.authinfo.gpg` etc.

If this idea is aceptable, I can implement it for your review. Let me know @dougm and anyone else interested.

Thanks!

tzz:
Resolves #647 as discussed with @dougm 

* autoloads `auth-source-search`
* only triggers using it if an entry in `govc-urls` does not have a user portion of the URL, in which case it's looked up as a logical `machine` in the `auth-sources`
* backwards compatible
* documentation updated
* `govc-urls` converted to a `defcustom` for user convenience

### DIFF
--- a/govc/emacs/govc.el
+++ b/govc/emacs/govc.el
@@ -47,6 +47,8 @@
 (require 'url-parse)
 (require 's)
 
+(autoload 'auth-source-search "auth-source")
+
 (defgroup govc nil
   "Emacs customization group for govc."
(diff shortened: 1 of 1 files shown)


dougm:
nice!

PR #652, opened 2017-01-03
Add object.collect command

dougm:
Generic interface to the property.Wait method, where properties can be collected from any managed object.

Emacs interface provides MOB (Managed Object Browser) functionality.


dougm:
mob

caglar10ur:
LGTM :sheep: :it:

PR #655, opened 2017-01-12
Remove _Task suffix from vapp methods

caglar10ur:
Also remove VApp suffix to avoid stutter.


caglar10ur:
Dropped the wrappers altogether. Will send change VIC code with the next govmomi vendor update

dougm:
Sounds good @caglar10ur - while we are renaming these, we could also drop the 'VApp' suffix.  We did this with the VirtualMachine methods, dropping the 'VM_Task' suffix.

caglar10ur:
Removed them @dougm

PR #656, opened 2017-01-17
Attach and list RDM/LUN

bastienbc:
govc does not provide command line options to:

- attach RDM using the device name.
- list available devices for a specific VM, retrieving various useful information ( UIDS, Device name, Device path, ... )

### DIFF
--- a/govc/vm/rdm/attach.go
+++ b/govc/vm/rdm/attach.go
@@ -1,0 +1,152 @@
+/*
+Copyright (c) 2014-2015 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rdm
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"strings"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/vim25/methods"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+type attach struct {
+	*flags.VirtualMachineFlag
+
+	device string
+}
+
+func init() {
+	cli.Register("vm.rdm.attach", &attach{})
+}
+
+func (cmd *attach) Register(ctx context.Context, f *flag.FlagSet) {
+
+	cmd.VirtualMachineFlag, ctx = flags.NewVirtualMachineFlag(ctx)
+	cmd.VirtualMachineFlag.Register(ctx, f)
+
+	f.StringVar(&cmd.device, "deviceName", "", "Device Name")
+}
+
+func (cmd *attach) Process(ctx context.Context) error {
+	if err := cmd.VirtualMachineFlag.Process(ctx); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (cmd *attach) Run(ctx context.Context, f *flag.FlagSet) error {
+	vm, err := cmd.VirtualMachine()
+	if err != nil {
+		return err
+	}
+
+	if vm == nil {
+		return flag.ErrHelp
+	}
+
+	devices, err := vm.Device(ctx)
+	if err != nil {
+		return err
+	}
+
+	controller, err := devices.FindSCSIController("")
+	if err != nil {
+		return err
+	}
+	var VM_withProp mo.VirtualMachine
+	err = vm.Properties(ctx, vm.Reference(), []string{"environmentBrowser"}, &VM_withProp)
+	if err != nil {
+		return err
+	}
+
+	//Query VM To Find Devices avilable for attaching to VM
+	var queryConfigRequest types.QueryConfigTarget
+	queryConfigRequest.This = VM_withProp.EnvironmentBrowser
+	cl, err := cmd.Client()
+	queryConfigResp, err := methods.QueryConfigTarget(ctx, cl, &queryConfigRequest)
+	if err != nil {
+		return err
+	}
+	vmConfigOptions := *queryConfigResp.Returnval
+	for _, ScsiDisk := range vmConfigOptions.ScsiDisk {
+		if !strings.Contains(ScsiDisk.Disk.CanonicalName, cmd.device) {
+			continue
+		}
+		var backing types.VirtualDiskRawDiskMappingVer1BackingInfo
+		backing.CompatibilityMode = "physicalMode"
+		backing.DeviceName = ScsiDisk.Disk.DeviceName
+		for _, descriptor := range ScsiDisk.Disk.Descriptor {
+			if string([]rune(descriptor.Id)[:4]) == "vml." {
+				backing.LunUuid = descriptor.Id
+				break
+			}
+		}
+		var device types.VirtualDisk
+		device.Backing = &backing
+		device.ControllerKey = controller.VirtualController.Key
+
+		var unitNumber *int32
+		scsiCtrlUnitNumber := controller.VirtualController.UnitNumber
+		var u int32
+		for u = 0; u < 16; u++ {
+			free := true
+			for _, device := range devices {
+				if device.GetVirtualDevice().ControllerKey == device.GetVirtualDevice().ControllerKey {
+					if u == *(device.GetVirtualDevice().UnitNumber) || u == *scsiCtrlUnitNumber {
+						free = false
+					}
+				}
+			}
+			if free {
+				unitNumber = &u
+				break
+			}
+		}
+		device.UnitNumber = unitNumber
+
+		spec := types.VirtualMachineConfigSpec{}
+
+		config := &types.VirtualDeviceConfigSpec{
+			Device:    &device,
+			Operation: types.VirtualDeviceConfigSpecOperationAdd,
+		}
+
+		config.FileOperation = types.VirtualDeviceConfigSpecFileOperationCreate
+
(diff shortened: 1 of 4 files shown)


govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,152 @@
...
+	f.StringVar(&cmd.device, "deviceName", "", "Device Name")
+}
+
+func (cmd *attach) Process(ctx context.Context) error {
+	if err := cmd.VirtualMachineFlag.Process(ctx); err != nil {
+		return err
+	}
+	return nil
+}
+
```
dougm: Can you add a Description here (example from vm.clone below).  This'll get included in the help output and the generated govc/USAGE.md doc.

```go
func (cmd *clone) Description() string {
	return `Clone VM to NAME.

Examples:
  govc vm.clone -vm template-vm new-vm`
}
```

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,152 @@
...
+	devices, err := vm.Device(ctx)
+	if err != nil {
+		return err
+	}
+
+	controller, err := devices.FindSCSIController("")
+	if err != nil {
+		return err
+	}
+	var VM_withProp mo.VirtualMachine
```
dougm: Can rename to avoid the go-lint warning below, just 'mvm' for example would be fine.

```
don't use underscores in Go names; var VM_withProp should be VMWithProp
```

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,152 @@
...
+		spec.DeviceChange = append(spec.DeviceChange, config)
+
+		task, err := vm.Reconfigure(ctx, spec)
+		if err != nil {
+			return err
+		}
+
+		err = task.Wait(ctx)
+		if err != nil {
+			return errors.New(fmt.Sprintf("Error adding device %+v \n with backing %+v \nLogged Item:  %s", device, backing, err))
```
dougm: from go-lint:
```
should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...)
```

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		return nil
+
+	}
+	return errors.New(fmt.Sprintf("Error: No LUN with device name containing %s found.", cmd.device))
```
dougm: from go-lint:
```
should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...)
```

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,152 @@
...
+	var VM_withProp mo.VirtualMachine
+	err = vm.Properties(ctx, vm.Reference(), []string{"environmentBrowser"}, &VM_withProp)
+	if err != nil {
+		return err
+	}
+
+	//Query VM To Find Devices avilable for attaching to VM
+	var queryConfigRequest types.QueryConfigTarget
+	queryConfigRequest.This = VM_withProp.EnvironmentBrowser
+	cl, err := cmd.Client()
```
dougm: check `err != nil` here.

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,152 @@
...
+		device.Backing = &backing
+		device.ControllerKey = controller.VirtualController.Key
+
+		var unitNumber *int32
+		scsiCtrlUnitNumber := controller.VirtualController.UnitNumber
+		var u int32
+		for u = 0; u < 16; u++ {
+			free := true
+			for _, device := range devices {
+				if device.GetVirtualDevice().ControllerKey == device.GetVirtualDevice().ControllerKey {
```
dougm: Looks like you need to rename 'device' in the loop (to 'd' for example), so you can compare to the device defined on line 105 above.

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,152 @@
...
+	vmConfigOptions := *queryConfigResp.Returnval
+	for _, ScsiDisk := range vmConfigOptions.ScsiDisk {
+		if !strings.Contains(ScsiDisk.Disk.CanonicalName, cmd.device) {
+			continue
+		}
+		var backing types.VirtualDiskRawDiskMappingVer1BackingInfo
+		backing.CompatibilityMode = "physicalMode"
+		backing.DeviceName = ScsiDisk.Disk.DeviceName
+		for _, descriptor := range ScsiDisk.Disk.Descriptor {
+			if string([]rune(descriptor.Id)[:4]) == "vml." {
```
dougm: `strings.HasPrefix(descriptor.Id, "vml.")`

govc/vm/rdm/ls.go
```diff
@@ -0,0 +1,114 @@
...
+	err = vm.Properties(ctx, vm.Reference(), []string{"environmentBrowser"}, &VM_withProp)
+	if err != nil {
+		return err
+	}
+
+	//Query VM To Find Devices available
+	var queryConfigRequest types.QueryConfigTarget
+	queryConfigRequest.This = VM_withProp.EnvironmentBrowser
+	cl, err := cmd.Client()
+	queryConfigResp, err := methods.QueryConfigTarget(ctx, cl, &queryConfigRequest)
```
dougm: Maybe wrap this in a helper, since its being used in 2 places:
```go
func (v VirtualMachine) QueryConfigTarget(ctx context.Context) (*types.ConfigTarget, error) {
	var vm mo.VirtualMachine

	err := v.Properties(ctx, v.Reference(), []string{"environmentBrowser"}, &vm)
	if err != nil {
		return nil, err
	}

	req := types.QueryConfigTarget{
		This: vm.EnvironmentBrowser,
	}

	res, err := methods.QueryConfigTarget(ctx, v.Client(), &req)
	if err != nil {
		return nil, err
	}

	return res.Returnval, nil
}
```

govc/vm/rdm/ls.go
```diff
@@ -0,0 +1,114 @@
...
+
+type infoResult struct {
+	Disks []types.VirtualMachineScsiDiskDeviceInfo
+}
+
+func (r *infoResult) Write(w io.Writer) error {
+	tw := tabwriter.NewWriter(os.Stdout, 2, 0, 2, ' ', 0)
+	for _, disk := range r.Disks {
+		fmt.Fprintf(tw, "Name:\t%s\n", disk.Name)
+		fmt.Fprintf(tw, "	Device name:\t%s\n", disk.Disk.DeviceName)
```
dougm: 2 space indent like the other commands, rather than tab.

bastienbc:
Hi!
First thank you to have a look into my code.
I am relatively  new to Go, so I am keeping any of your advises preciously.
Also I did not knew about golint, which is now installed and used :-)

About this pull request itself:
Should I change the date on the Copyright?
To 2014-2017?
Or else?
You mentioned it on an other pull request comment, so I was wondering.

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,149 @@
...
+func init() {
+	cli.Register("vm.rdm.attach", &attach{})
+}
+
+func (cmd *attach) Register(ctx context.Context, f *flag.FlagSet) {
+
+	cmd.VirtualMachineFlag, ctx = flags.NewVirtualMachineFlag(ctx)
+	cmd.VirtualMachineFlag.Register(ctx, f)
+
+	f.StringVar(&cmd.device, "deviceName", "", "Device Name")
```
dougm: Let's name the flag "device", we use that name in other commands.

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (cmd *attach) Description() string {
+	return `Attach DEVICE to VM with RDM.
+
+Examples:
+  govc vm.rdm.attach -vm VM -deviceName DEVICE`
```
dougm: s/deviceName/device/ here too.  Can you use a real example value for DEVICE?  An example from the output of device.ls, such as "disk-200-0"

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return err
+	}
+
+	vmConfigOptions, err := vm.QueryEnvironmentBrowser(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, ScsiDisk := range vmConfigOptions.ScsiDisk {
```
dougm: Go style nit: unexported vars start with lower case, 'scsiDisk' or just 'disk' here would be fine.

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,149 @@
...
+		if !strings.Contains(ScsiDisk.Disk.CanonicalName, cmd.device) {
+			continue
+		}
+		var backing types.VirtualDiskRawDiskMappingVer1BackingInfo
+		backing.CompatibilityMode = "physicalMode"
```
dougm: Could use the generated enum here: `string(types.VirtualDiskCompatibilityModePhysicalMode)`

govc/vm/rdm/attach.go
```diff
@@ -0,0 +1,149 @@
...
+		device.Backing = &backing
+		device.ControllerKey = controller.VirtualController.Key
+
+		var unitNumber *int32
+		scsiCtrlUnitNumber := controller.VirtualController.UnitNumber
+		var u int32
+		for u = 0; u < 16; u++ {
+			free := true
+			for _, d := range devices {
+				if d.GetVirtualDevice().ControllerKey == d.GetVirtualDevice().ControllerKey {
```
dougm: Still comparing the same device ('d'), shouldn't this compare the ControllerKey of 'd' and 'device' instead ?

object/virtual_machine.go
```diff
@@ -689,3 +689,24 @@ func (v VirtualMachine) Unregister(ctx context.Context) error {
 	_, err := methods.UnregisterVM(ctx, v.Client(), &req)
 	return err
 }
+
+// QueryEnvironmentBrowser is a helper to get the environmentBrowser property.
+func (v VirtualMachine) QueryEnvironmentBrowser(ctx context.Context) (*types.ConfigTarget, error) {
```
dougm: Let's name this method 'QueryConfigTarget', since 'EnvironmentBrowser' is just a parameter.

dougm:
@bastienbc thanks for the updates!  Just '2017' is fine for the copyright year.

dougm:
thanks @bastienbc !

PR #657, opened 2017-01-17
Add the option to describe a VM using the annotation option

bastienbc:
govc does not provide command line options to describe a VM.
Using the annotation field in VirtualMachineConfigSpec, vm.create and vm.clone now both have this option.


dougm:
@bastienbc nice, thanks!

PR #658, opened 2017-01-18
Add a command line option to change an existing disk attached to a VM

bastienbc:
I needed the ability to resize a VM disk once it has been created, so I made a new command: "vm.disk.change".

It allows search of a disk using its name, label, key or even the file path to the vmdk.
It only changes the size, and the mode.

### DIFF
--- a/govc/vm/disk/change.go
+++ b/govc/vm/disk/change.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2014-2015 VMware, Inc. All Rights Reserved.
+Copyright (c) 2017 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/govc/vm/disk/change.go
+++ b/govc/vm/disk/change.go
@@ -1,0 +1,179 @@
+/*
+Copyright (c) 2014-2015 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


govc/vm/disk/change.go
```diff
@@ -0,0 +1,179 @@
+/*
+Copyright (c) 2014-2015 VMware, Inc. All Rights Reserved.
```
caglar10ur: This needs to be 2017

govc/vm/disk/change.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return nil
+}
+
+func (cmd *change) FindDisk(ctx context.Context, list object.VirtualDeviceList) (*types.VirtualDisk, error) {
+	var disks []*types.VirtualDisk
+	for _, device := range list {
+		switch md := device.(type) {
+		case *types.VirtualDisk:
+			switch {
```
dougm: This looks the same as CheckDiskProperties?  If so, personally would prefer to see CheckDiskProperties return a bool and call it here to remove the duplicate code.

govc/vm/disk/change.go
```diff
@@ -0,0 +1,179 @@
...
+	spec.DeviceChange = append(spec.DeviceChange, config)
+
+	task, err := vm.Reconfigure(ctx, spec)
+	if err != nil {
+		return err
+	}
+
+	err = task.Wait(ctx)
+	if err != nil {
+		return errors.New(fmt.Sprintf("Error resizing main disk\nLogged Item:  %s", err))
```
dougm: ```should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...) (go-golint)```

govc/vm/disk/change.go
dougm: Ah yes, also in the RDM PR.

dougm: Yes, just '2017' is fine here.  Everything else looks good @bastienbc

PR #660, opened 2017-01-24
Add device option to WaitForNetIP

### LINKED ISSUE #653: waitForNetIp to take arguments allowing to wait for specific nic(s)
We are experiencing the following [issue](https://github.com/hashicorp/terraform/issues/7479) when using terraform to provision vsphere virtual machines.

If we add multiple NIC to a host, and the additional nics do not receive an IP address, terraform will not complete provisioning as it calls the waitForNetIp function.

In our scenario we're provisioning hosts that will use additional nics as bridge interfaces, and as such will have a mac address, but no IP.

To temporarily solve this I made this update to the waitForNetIp:

```
		for _, ips := range macs {
			if len(ips) > 0 {
				return true
			}
		}

		return false
```

It would be great if the WaitForNetIp could receive an array of nic# to specifically wait for.

This way we could mark one (or more) NIC as "management" in terraform and waitForNetIp could disregard the others.

Does it make sense?

dougm:
With this option, WaitForNetIP can specify NIC(s) by MAC address or device name.

Fixes #653

### DIFF
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -2169,6 +2169,7 @@ Examples:
   govc vm.clone -vm template-vm new-vm
 
 Options:
+  -annotation=              VM description
   -c=0                      Number of CPUs
   -customization=           Customization Specification Name
   -datastore-cluster=       Datastore cluster [GOVC_DATASTORE_CLUSTER]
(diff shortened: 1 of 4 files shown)


dougm:
In support of hashicorp/terraform#7479

caglar10ur:
LGTM :sheep: :it:

PR #663, opened 2017-02-11
Finder: support automatic Folder recursion

### LINKED ISSUE #190: finder.VirtualMachine and VC

### LINKED ISSUE #519: finder: Recursive DVS find
I need to recurse over a datacenter to list all existing Network resources (including DVS).

It looks like this is not supported: the recursion parameter is set to false in finder.go:548:

``` go
func (f *Finder) NetworkList(ctx context.Context, path string) ([]object.NetworkReference, error) {
    es, err := f.find(ctx, f.networkFolder, false, path)
        ...
}
```

Are there plan to provide a recursion-enabled version? If not, what's the rationale?

Thanks.

### LINKED ISSUE #540: finder.DatastoreList broken with Datastore Clusters / Storage Pods
When running a simple govmomi program against a vSphere 6.0 U2 server, `finder.DatastoreList` fails with a `datastore '*' not found` error. However `finder.DatastoreClusterList` returns the two clusters as expected.

dougm:
Prior to this change, the Finder behaved like the 'ls' command and
as such required a Folder name or wildcard in the path argument to
find resources within folders.  The original "list" mode of the Finder
is maintained, but will now switch to "find" mode if the path argument
does not contain a "/".  See find/doc.go for details.

The public Finder method signatures remain the same, with the exception of
ManagedObjectListChildren which now takes an optional (variadic) list of
types.

No new public methods or types have been added.

The 'list.Recurser' type is no longer exported and has moved to an unexported
type within the find package.  This part of the change assumes that 'list.Recurser'
was not used directly outside of the govmomi repo.

The tests depend on the Go based vCenter simulator, which currently lives
in the vmware/vic repo, but will be moving to the govmomi repo at some point.

Fixes #190
Fixes #519
Fixes #540

Closes #507
Closes #593


dougm:
The mentioned vcsim based tests are in vmware/vic#3908 (commit https://github.com/vmware/vic/pull/3908/commits/051d795fd0aafd514cb49da01b0c83895e521714)

PR #664, opened 2017-02-14
Support alternative './...' syntax for finder

caglar10ur:


dougm:
👍 

can you also update: https://github.com/vmware/govmomi/blob/master/govc/emacs/govc.el#L397

PR #672, opened 2017-02-25
 Add vsan and disk commands / helpers

dougm:
* Add DatastoreFileManager API wrapper

* Add HostVsanInternalSystem API wrappers

* Add govc datastore.disk.create command

* Add govc datastore.vsan.ls and datastore.vsan.rm commands

* Use DatastoreFileManager in govc datastore.rm command

### DIFF
--- a/govc/datastore/disk/create.go
+++ b/govc/datastore/disk/create.go
@@ -1,0 +1,98 @@
+/*
+Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package disk
+
+import (
+	"context"
+	"flag"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/units"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+type create struct {
+	*flags.DatastoreFlag
+
+	Bytes units.ByteSize
+}
+
+func init() {
+	cli.Register("datastore.disk.create", &create{})
+}
+
+func (cmd *create) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.DatastoreFlag, ctx = flags.NewDatastoreFlag(ctx)
+	cmd.DatastoreFlag.Register(ctx, f)
+
+	_ = cmd.Bytes.Set("10G")
+	f.Var(&cmd.Bytes, "size", "Size of new disk")
+}
+
+func (cmd *create) Process(ctx context.Context) error {
+	if err := cmd.DatastoreFlag.Process(ctx); err != nil {
+		return err
+	}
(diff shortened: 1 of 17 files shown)


govc/datastore/disk/create.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (cmd *create) Usage() string {
+	return "VMDK"
+}
+
+func (cmd *create) Description() string {
+	return `Create VMDK on DS.
+
+Examples:
+  govc datastore.disk.create -size 24G disks/disk1.vmdk`
```
hickeng: Is `disks` the datastore or is this on `datastore` by default?

govc/datastore/rm.go
```diff
@@ -43,6 +43,7 @@ func (cmd *rm) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.DatastoreFlag, ctx = flags.NewDatastoreFlag(ctx)
 	cmd.DatastoreFlag.Register(ctx, f)
 
+	f.BoolVar(&cmd.kind, "t", true, "Use file type to choose disk or file manager")
```
hickeng: might find out further down, but not sure what will be non-functional if this is set to false.

govc/datastore/vsan/ls.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	var mds mo.Datastore
+	err = ds.Properties(ctx, ds.Reference(), []string{"summary"}, &mds)
+	if err != nil {
+		return err
+	}
+
+	if mds.Summary.Type != "vsan" {
+		return flag.ErrHelp
```
hickeng: Perhaps include the fact that the target is not a vSAN datastore.

object/datastore_file_manager.go
```diff
@@ -0,0 +1,144 @@
...
+		p.Datastore = m.Datastore.Name()
+	}
+
+	return &p
+}
+
+func (m *DatastoreFileManager) markDiskAsDeletable(ctx context.Context, path *DatastorePath) {
+	r, n, err := m.Datastore.Download(ctx, path.Path, &soap.DefaultDownload)
+	if err != nil {
+		return
```
hickeng: not returning a cause here may make it hard to differentiate between errors:
* disk in use
* missing Host.Config.SystemManagement privilege
* connection error
* other

object/datastore_file_manager.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	defer r.Close()
+
+	if n > 2048 {
+		return // sanity check before reading; should be only a few hundred bytes
```
hickeng: This is true for the moment but we are discussing using the ddb to store the layer metadata. Regardless, if we've somehow accidentally downloaded the data file instead of the descriptor this will be Gigabytes too late - we could download only the first 512 bytes and check for known keys (for example) - not really thought about edge cases with this suggestion however.

govc/datastore/disk/create.go
dougm: Need to datastore.mkdir first if it doesn't already exist, I'll add that to the example.

govc/datastore/rm.go
dougm: We're using it in the datastore_file_manager.sh to test the vsan.dom commands

object/datastore_file_manager.go
dougm: We discussed in person, I'll add comment/log.

object/datastore_file_manager.go
dougm: Regarding ddb for metadata, hopefully we won't still be using this workaround by then :)

Regarding the download, we've only read the HTTP headers at this point, 'n' is the value of the Content-Length header.  Nothing reads the HTTP body until we call `s.Scan()` below.

dougm: Changed to use https://golang.org/pkg/io/#LimitedReader instead of just checking the Content-Length header.

PR #675, opened 2017-03-01
Host is optional for MarkAsVirtualMachine

dougm:
The vm.markasvm command had required the host flag,
but it is optional in the API.
Add the pool flag to avoid specifying a host.


dougm:
Discussed with @elibarb on slack

elibarb:
Thx!

PR #676, opened 2017-03-01
Add vm.guest.tools command

### LINKED ISSUE #674: Feature request: ability to mount vmware-tools iso
Hello,

I did not find a way with govc/govmomi to mount the vmware tools iso on a vm, like the vim-cmd vmsvc/tools.install CLI commands. 

May I'm looking for the wrong way to achieves that.

dougm:
Closes #674

### DIFF
--- a/govc/vm/guest/tools.go
+++ b/govc/vm/guest/tools.go
@@ -1,0 +1,116 @@
+/*
+Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


marema31:
I've tested on our infrastructure, it works as wanted.

The only remarks I have is if you run `govc vm.guest.tools` it return `govc: no argument`. I would think it will be returning the usage page like vm.change for example.

dougm:
ok great.  The message comes from the existing flag: https://github.com/vmware/govmomi/blob/master/govc/flags/search.go#L306

There's a few of them, will look at that as a separate PR.

PR #677, opened 2017-03-03
Adds FindManagedObject method.

jrmycanady:
This is the implementation of FindEntityView that is in other SDKs that I mentioned on Slack. Give it a look over and let me know if it fits and if so where. Below is an example. In my local repo this is actually a method off of the govmomi.Client as I find it easier to use that way. e.g. client.FindManagedObject(..)

### Final all hosts with "test" in the name from the root folder down and return only the name property
```
var hostSystems []mo.HostSystem
props := []string{"name"}
filter := make(map[string]string)
fitler["name"] = "esd"
find.FindManagedObject(ctx, client.Client, "HostSystem", filter, props, client.Client.ServiceContent.RootFolder, &hostSystems)
```

### DIFF
--- a/find/find_object.go
+++ b/find/find_object.go
@@ -1,0 +1,260 @@
+package find
+
+import (
+	"context"
+	"errors"
+	"reflect"
+	"regexp"
+	"strconv"
+
+	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/view"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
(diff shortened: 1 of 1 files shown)


dougm:
Thanks @jrmycanady , I'll take a look and give this a try soon.

dougm:
@jrmycanady this is great stuff.  I started trying it out by implementing a govc object.find command.  Moved things around in the process, but should still have the same functionality.  Next I want to look at some integration with the Finder.  Will probably leave the "list" mode there as-is for now, but ContainerView should cleanup + optimize the "find" mode.  Current changes are here, would be great to get your review:

https://github.com/vmware/govmomi/compare/master...dougm:object/find

dougm:
Merged/refactored in #693

PR #679, opened 2017-03-06
Add check for nil envelope

bruceadowns:
* guard for nil VirtualSystem and Network in envelope


dougm:
Maybe we could change `ArchiveFlag.ReadEnvelope` to return an empty `&ovf.Envelope{}`if fpath == ""?  Just to avoid the `e != nil` checks here and the existing ones too.

bruceadowns:
Ya, that's a good idea. I'll send an amendment.

dougm:
thanks @bruceadowns

PR #680, opened 2017-03-06
Remove deferred Close() call in follower's Read()

jzt:
This removes the deferred call to `Close()` in the datastore file follower's `Read()` method. This prevents `Close()` from being called prematurely when `EOF` is not encountered after a read.


dougm:
Thanks @jzt . Keep in mind followDatastoreFile Reader impl is meant for io.Copy (or similar) and its 'Close' method just stops the polling.  So the caller may need to call `DatastoreFile.Close` in addition to `followDatastoreFile.Close`, if `followDatastoreFile.Read` had been called with a small buffer.

PR #682, opened 2017-03-10
Add performance manager and govc metric commands

dougm:
* Convert counter IDs to and from stable names, with caching

* Query available metrics, per resource type and interval

* Query N metric samples by name for N resource instances

* Change level per historical interval and per counter

### DIFF
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -1743,6 +1743,127 @@ Options:
   -t=                       Object type
 ```
 
+## metric.change
+
+```
+Usage: govc metric.change [OPTIONS] NAME...
+
+Change counter NAME levels.
+
+Examples:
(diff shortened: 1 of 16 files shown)


cgtexmex:
LGTM

From a govc perspective two potential improvements:
- flag for detailed output (including timestamps, etc...I thought json would do that, but I got nothing)
- ability to tail/follow the metric at the specified interval

dougm:
@cgtexmex I meant to include support for -json, I'll do that.

You can get tail with the metric.sample '-n' flag.

I would also like a follow, and something that could generate a graph, but will save that for after this PR.

dougm:
@cgtexmex added the json support and a '-t' flag to metric.sample

cgtexmex:
LGTM -- Thanks!

PR #687, opened 2017-03-17
Add Bash completion script for govc

ericgray:
This completion script handles some of the basics such as subcommands and appropriate options for each.  Also supplies path information when working with VMs or listing other objects.

Environment variables for authentication must be set for full functionality.


dougm:
nice, thanks @ericgray

PR #689, opened 2017-03-22
Finder: support changing object root in find mode

dougm:
- Use path.Match everywhere, not filepath.Match

### DIFF
--- a/find/finder.go
+++ b/find/finder.go
@@ -34,12 +34,14 @@ type Finder struct {
 	client  *vim25.Client
 	r       recurser
 	dc      *object.Datacenter
+	si      *object.SearchIndex
 	folders *object.DatacenterFolders
 }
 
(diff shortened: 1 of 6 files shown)


find/finder.go
```diff
@@ -55,6 +57,35 @@ func (f *Finder) SetDatacenter(dc *object.Datacenter) *Finder {
...
+
+	ix := len(parts) - 1
+
+	if parts[ix] != "..." {
+		return false
+	}
+
+	rootPath := path.Join(root.Path, path.Join(parts[:ix]...))
+
+	if rootPath != root.Path {
```
hickeng: this test could be moved above the path.Join as:
```go
if ix == 0 {
   return true
}
```

dougm: Yes indeed, done.

PR #693, opened 2017-03-29
Add object.find command

dougm:
Move filtering to general purpose property.Filter:

- Reduce use of reflect

- Use filepath.Match instead of Regexp, consistent with existing Finder behavior

Move ContainerView based property collection to ContainerView.Retrieve:

- Symmetry with property.Collector.Retrieve

- Support []types.ObjectContent as Retrieve destination


dougm:
@jrmycanady tried adding you as a reviewer, but github isn't letting me for some reason.

jrmycanady:
@dougm Not sure why it won't let you but I will look it over tomorrow.

property/collector.go
```diff
@@ -164,6 +164,11 @@ func (p *Collector) Retrieve(ctx context.Context, objs []types.ManagedObjectRefe
 		return err
 	}
 
+	if d, ok := dst.(*[]types.ObjectContent); ok {
+		*d = res.Returnval
```
caglar10ur: just curious, why are we setting d here?

dougm: `dst interface{}`, if the type is a pointer to `[]types.ObjectContent` we just set dst to that slice and skip the call to `mo.LoadRetrievePropertiesResponse`

govc/object/find.go
```diff
@@ -0,0 +1,246 @@
...
+		if kind == k {
+			return true
+		}
+	}
+
+	return false
+}
+
+func init() {
+	cli.Register("object.find", &find{})
```
dougm: Debating the command name, "find" or "object.find", any votes?

dougm:
@jrmycanady great, thanks!

govc/object/find.go
ericgray: Is there anything other than "objects" to find?  If not, then "object." might be redundant.

dougm: No, have just been trying to avoid adding more top-level level commands.  But we already have the top-level "ls" command that applies to all objects, so "find" probably shouldn't have the object. prefix.  We could end up with a "datastore.find" that applies to files on the datastore, but that's ok, there are a bunch of other foo.ls commands for different things.  Think we can leave the code in the object directory, but name the command "find".

dougm:
@jrmycanady I've gone ahead and merged to include the new find command in the 0.14 govc release.  We can still tweak the new APIs if you need any changes.  Thanks again.

PR #694, opened 2017-03-29
Use Duration flag for vm.ip -wait flag

dougm:
- Default to 1hr (previous default was infinite)

- Don't error or print anything when time has expired


dougm:
@marema31 how about this, using a Duration flag instead of boolean for '-wait' ?

marema31:
IMHO It's a more clever solution since we can provide seconds to this flag it will be more versatile.

PR #699, opened 2017-04-07
Updated virtual_machine.go

aniketGslab:
Added new function - SearchSnapshot()
SearchSnapshot() is the simple clone of the findSnapshot().
The intention behind cloning this function is to have a functionality for Searching the snapshot which can be used in other packages as well  without breaking the current functionality of this library as some of the functions are dependent on findSnapshot()
This function will be useful in other packages.


dougm:
@aniketGslab no need to copy `findSnapshot` to a new method, why not just export the existing method as `FindSnapshot`?  Dependents can use this exported method too.

dougm:
Closed by #715

Issue #659, opened 2017-01-24
SPBM policy support in govmomi


BaluDontu:
This is a new feature request. 

Currently, we want to provide VSAN policy support for kubernetes storage volume provisioning. Given a policyName/policyID we would require a way to identify the list of compatible datastores that satisfies the given VSAN policy.

BaluDontu:
In Kubernetes, the user can create a storage class for storage volume provisioning in vSphere. We are extending this capability to allow user to provide a storage policy already existing on the vSphere. This way the volume will be created on the datastore that satisfies the policy.

For example, 

kind: StorageClass
apiVersion: storage.k8s.io/v1beta1
metadata:
  name: slow
provisioner: kubernetes.io/vsphere-volume
parameters:
  policy: Gold

Since vSphere UI allows us to create storage policies using SPBM. We would need help in using the existing SPBM functionality implemented in GO to get the required information. The following functions need to be implemented in govmomi.

1. GetAllStoragePolicies()
This can be implemented by making the following calls to SPBM

a) PbmQueryProfile (resourceType, profileCategory) 
- This get list of all SPBM storage policies.
- In here resourceType param is 

> resourceType=pbm.profile.ResourceType(resourceType="STORAGE")
 
and profileCategory is 

> profileCategory="REQUIREMENT"

- API declaration : https://opengrok.eng.vmware.com/source/xref/storage-main.perforce.1666/bora/vim/vmodl/pbm/profile/ProfileManager.java#316

b) PbmRetrieveContent (profileIds)
- Retrieves the content of all profile ID's. 
- API declaration : https://opengrok.eng.vmware.com/source/xref/storage-main.perforce.1666/bora/vim/vmodl/pbm/profile/ProfileManager.java#347

Sample python implementation is also found here - http://www.virtuallyghetto.com/2017/01/spbm-apis-are-now-included-in-pyvmomi-vsphere-sdk-for-python.html. Usage is pretty much direct calls.

2. GetMatchingDatastores (profileID)
- Gets list of all matching datastores for the given policy.

a) PbmCheckRequirements (PlacementHub[] hubsToSearch, ServerObjectRef placementSubjectRef, Requirement[] placementSubjectRequirement)
- Get list of complaint/non-complaint datastores for a given profile.
- API declaration : https://opengrok.eng.vmware.com/source/xref/storage-main.perforce.1666/bora/vim/vmodl/pbm/placement/PlacementSolver.java#213
[...]

PR #711, opened 2017-04-24
Add PBM support

dougm:


dougm:
Issue #659

BaluDontu:
/lgtm. I have verified the output and seems to have the correct results. 
Its better if we also have CreateProfile() and DeleteProfile() as well in pbm/client.go.

PR #713, opened 2017-04-27
Add AuthorizationManager {Enable,Disable}Methods

dougm:


chengwang86:
So with this PR, the VCH could add `Destroy_task` to the list of disabledMethods each time after a create a new containerVM. 

- If the VCH has global privilege and the operation succeeds, the `delete VM from disk` option will be grey on vsphere UI. 
- If the VCH does not have enough privilege (ops-user with limited privileges), the operation will fail and we simply log the failure error msg. 

Is this the workflow you propose?

chengwang86:
LGTM

dougm:
@chengwang86 yes, exactly.

govc/object/method.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (cmd *method) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.DatacenterFlag, ctx = flags.NewDatacenterFlag(ctx)
+	cmd.DatacenterFlag.Register(ctx, f)
+
+	f.StringVar(&cmd.name, "name", "", "Method name")
+	f.StringVar(&cmd.reason, "reason", "", "Reason for disabling method")
+	f.StringVar(&cmd.source, "source", "govc", "Source ID")
+	f.BoolVar(&cmd.enable, "enable", true, "Enable method")
```
hickeng: Should this be `enabled` rather than `enable`? That would be consistent with the esxcli style but honestly don't recall for the rest of govc.

PR #714, opened 2017-05-01
PBM Support in govmomi - Additional PBM functionality

BaluDontu:
Added additional wrapper functions for PBM support in govmomi. The below functionalities are included.

-  GetCompatibleDatastores
-  GetNonCompatibleDatastores
-  CreatePbmCapabilityProfileSpec
-  CreatePbmProfile
-  UpdatePbmProfile
-  DeletePbmProfile
-  QueryAssociatedEntity
-  QueryAssociatedEntities
-  GetPbmProfileIDByName

I have also included unit tests for all the above PBM functions. 

Current support is added for CREATING following PBM profiles:
1. VSAN profile.
2. VSAN + SIOC profile.
3. Tag Based profile.

@dougm : Please take a look.
@divyenpatel @tusharnt @pdhamdhere

### DIFF
--- a/pbm/client.go
+++ b/pbm/client.go
@@ -18,6 +18,7 @@ package pbm
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/vmware/govmomi/pbm/methods"
 	"github.com/vmware/govmomi/pbm/types"
@@ -79,7 +80,9 @@ func (c *Client) RetrieveContent(ctx context.Context, ids []types.PbmProfileId) 
 	return res.Returnval, nil
 }
 
-func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacementHub, ref *types.PbmServerObjectRef, preq []types.BasePbmPlacementRequirement) ([]types.PbmPlacementCompatibilityResult, error) {
+type PlacementCompatibilityResult []types.PbmPlacementCompatibilityResult
+
+func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacementHub, ref *types.PbmServerObjectRef, preq []types.BasePbmPlacementRequirement) (PlacementCompatibilityResult, error) {
 	req := types.PbmCheckRequirements{
 		This:                        c.ServiceContent.PlacementSolver,
 		HubsToSearch:                hubs,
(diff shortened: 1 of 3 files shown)


pbm/client.go
```diff
@@ -94,3 +95,170 @@ func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacemen
...
+				Type:  res.Hub.HubType,
+				Value: res.Hub.HubId,
+			})
+		}
+	}
+
+	return compatibleDatastores, nil
+}
+
+func (c *Client) GetNonCompatibleDatastores(ctx context.Context, datastores []vim.ManagedObjectReference, ref *types.PbmServerObjectRef, ids []types.PbmProfileId) ([]vim.ManagedObjectReference, error) {
```
divyenpatel: `GetCompatibleDatastores()` and `GetNonCompatibleDatastores()`, is almost same. can we move duplicate code to some private func.

pbm/client.go
```diff
@@ -94,3 +95,170 @@ func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacemen
...
+
+func (c *Client) CreatePbmProfile(ctx context.Context, capabilityProfileCreateSpec types.PbmCapabilityProfileCreateSpec) (types.PbmProfileId, error) {
+	req := types.PbmCreate{
+		This:       c.ServiceContent.ProfileManager,
+		CreateSpec: capabilityProfileCreateSpec,
+	}
+
+	res, err := methods.PbmCreate(ctx, c, &req)
+	if err != nil {
+		var profileID types.PbmProfileId
```
dougm: Change the method to return type `*types.PbmProfileId`, then you can use `nil` here and `&res.ReturnVal` below.

pbm/client.go
```diff
@@ -94,3 +95,170 @@ func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacemen
...
+	}
+
+	return res.Returnval, nil
+}
+
+func (c *Client) UpdatePbmProfile(ctx context.Context, id types.PbmProfileId, capabilityProfileUpdateSpec types.PbmCapabilityProfileUpdateSpec) error {
+	req := types.PbmUpdate{
+		This:       c.ServiceContent.ProfileManager,
+		ProfileId:  id,
+		UpdateSpec: capabilityProfileUpdateSpec,
```
dougm: Could just name the param `updateSpec` instead of `capabilityProfileUpdateSpec`.

pbm/pbm_util.go
```diff
@@ -0,0 +1,147 @@
...
+	Operator string
+	Value    string
+	DataType string
+}
+
+func CreatePbmCapabilityProfileSpec(pbmCreateSpec CapabilityProfileCreateSpec) (types.PbmCapabilityProfileCreateSpec, error) {
+	var pbmCapabilityProfileSpec types.PbmCapabilityProfileCreateSpec
+	capabilities, err := createCapabilityInstances(pbmCreateSpec.CapabilityList)
+	if err != nil {
+		return pbmCapabilityProfileSpec, err
```
dougm: Change method return type to `*types.PbmCapabilityProfileCreateSpec` and return `nil` here.

pbm/pbm_util.go
```diff
@@ -0,0 +1,147 @@
...
+		var propertyInstances []types.PbmCapabilityPropertyInstance
+		for _, propertyRule := range capabilityRule.PropertyList {
+			property := types.PbmCapabilityPropertyInstance{
+				Id: propertyRule.ID,
+			}
+			if propertyRule.Operator != "" {
+				property.Operator = propertyRule.Operator
+			}
+			var err error
+			switch strings.ToLower(propertyRule.DataType) {
```
dougm: Maybe you could use reflection instead of having to specify DataType, for example:
https://github.com/vmware/govmomi/blob/master/property/filter.go#L77

But I suppose that won't work well in the case of the "set" type.

pbm/client.go
dougm: Or single method + a struct that includes both to avoid making the remote call twice.

Another option, change `CheckRequirements` to return a helper type:
`type PlacementCompatibilityResult
[]types.PbmPlacementCompatibilityResult`

And attach helper methods to the helper type to filter/transform/etc in memory.

We do similar elsewhere, for example:
https://github.com/vmware/govmomi/blob/master/object/host_firewall_system.go#L82
https://github.com/vmware/govmomi/blob/master/object/authorization_manager.go#L40

pbm/client.go
BaluDontu: Addressed it.

pbm/client.go
BaluDontu: Addressed it.

pbm/pbm_util.go
BaluDontu: Addressed it.

pbm/client.go
BaluDontu: Addressed it. Used helper type, 
_type PlacementCompatibilityResult []types.PbmPlacementCompatibilityResult_

pbm/pbm_util.go
BaluDontu: Ya that's the problem. The schema might of different types including range, set etc. So I have decided to have the type field.

BaluDontu:
@dougm : Please check the latest commit.

pbm/client.go
```diff
@@ -94,3 +97,121 @@ func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacemen
 
 	return res.Returnval, nil
 }
+
+func (l PlacementCompatibilityResult) GetCompatibleDatastores() []types.PbmPlacementHub {
+	var compatibleDatastores []types.PbmPlacementHub
+
+	for _, res := range l {
+		if len(res.Error) == 0 {
```
divyenpatel: can we use `if res.Error==""`

pbm/client.go
```diff
@@ -94,3 +97,121 @@ func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacemen
...
+		}
+	}
+	return compatibleDatastores
+}
+
+func (l PlacementCompatibilityResult) GetNonCompatibleDatastores() []types.PbmPlacementHub {
+	var nonCompatibleDatastores []types.PbmPlacementHub
+
+	for _, res := range l {
+		if len(res.Error) > 0 {
```
divyenpatel: can we use `if res.Error!=""`

pbm/client.go
```diff
@@ -94,3 +97,121 @@ func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacemen
 
 	return res.Returnval, nil
 }
+
+func (l PlacementCompatibilityResult) GetCompatibleDatastores() []types.PbmPlacementHub {
```
dougm: Nit: remove "Get" prefix from this and the method below.

See: https://golang.org/doc/effective_go.html#Getters

pbm/client.go
```diff
@@ -94,3 +97,121 @@ func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacemen
...
+
+	for _, res := range l {
+		if len(res.Error) > 0 {
+			nonCompatibleDatastores = append(nonCompatibleDatastores, res.Hub)
+		}
+	}
+	return nonCompatibleDatastores
+}
+
+func (c *Client) CreatePbmProfile(ctx context.Context, capabilityProfileCreateSpec types.PbmCapabilityProfileCreateSpec) (*types.PbmProfileId, error) {
```
dougm: Let's also remove "Pbm" from the Client wrapper methods, they are already namespace protected being attached Clieng and in the pbm package.  I would have removed "Pbm" from the generated types too, but was too much trouble wrt to how Go+xml+reflect work.

pbm/client.go
```diff
@@ -94,3 +97,121 @@ func (c *Client) CheckRequirements(ctx context.Context, hubs []types.PbmPlacemen
...
+		return "", err
+	}
+
+	for i := range profiles {
+		profile := profiles[i].GetPbmProfile()
+		if profile.Name == profileName {
+			return profile.ProfileId.UniqueId, nil
+		}
+	}
+	return "", fmt.Errorf("No PBM profile found with name: %q", profileName)
```
dougm: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pbm/client.go
BaluDontu: Good to know all these coding standards. Addressed it.

pbm/client.go
BaluDontu: Addressed it.

pbm/client.go
BaluDontu: Addressed it.

BaluDontu:
@dougm : Good to know all these coding standards. Addressed all of them.

dougm:
@BaluDontu nice!

PR #715, opened 2017-05-04
Exported FindSnapshot() Method

aniketGslab:
I have exported FindSnapshot() so that other packages will be able to access it.

### DIFF
--- a/object/virtual_machine.go
+++ b/object/virtual_machine.go
@@ -518,11 +518,11 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


object/virtual_machine.go
```diff
@@ -518,11 +518,11 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
 	}
 }
 
-// findSnapshot supports snapshot lookup by name, where name can be:
+// FindSnapshpt supports snapshot lookup by name, where name can be:
```
dougm: typo: "shpt" should be "shot"

dougm:
:+1:

PR #716, opened 2017-05-04
New examples: hosts and virtualmachines. WIP

amandahla:
TODO:
- [x] Adding new examples for collect hosts and virtual machines information.

### DIFF
--- a/examples/hosts/main.go
+++ b/examples/hosts/main.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2015 VMware, Inc. All Rights Reserved.
+Copyright (c) 2017 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


dougm:
Thanks @amandahla 

Looks like these examples would be a good place to use the newer ContainerView APIs.  This reduces the number of round trips quite a bit.  An untested example below.  What do you think?

```go
	m := view.NewManager(c.Client)

	v, err := m.CreateContainerView(ctx, c.ServiceContent.RootFolder, []string{"VirtualMachine"}, true)
	if err != nil {
		exit(err)
	}

	var vms []mo.VirtualMachine

	err = v.RetrieveWithFilter(ctx, kind, []string{"summary"}, &vms, nil)
	if err != nil {
		exit(err)
	}

	_ = v.Destroy(ctx)
```

amandahla:
@dougm Thanks, I'll change them! :-)

amandahla:
@dougm I updated datastores also, ok? Did I get the ideia correctly? When we use view package, they get from all datacenters, right?

examples/hosts/main.go
```diff
@@ -0,0 +1,168 @@
+/*
+Copyright (c) 2015 VMware, Inc. All Rights Reserved.
```
dougm: Let's me these 2017

examples/virtualmachines/main.go
```diff
@@ -0,0 +1,180 @@
...
+	// Get summary property (VirtualMachineSummary)
+	// Reference for this data object:
+	// https://www.vmware.com/support/developer/vc-sdk/visdk25pubs/ReferenceGuide/vim.vm.Summary.html
+
+	var vms []mo.VirtualMachine
+	err = v.Retrieve(ctx, []string{"VirtualMachine"}, []string{"summary"}, &vms)
+	if err != nil {
+		exit(err)
+	}
+
```
dougm: At this point, you have all the data in memory, so there's no need for the go routines.  Just the for loop + print is fine.

amandahla: Hi @dougm  Changed now :)

dougm:
@amandahla great, thanks!

PR #719, opened 2017-05-09
Add options to host.autostart.add

weinrea:
This is to allow for per VM config to the HostAutoStartManager.
Also useful to configure an ordered auto startup.

Example:
govc host.autostart.add  -start-order=1 -start-delay=5 vm1 vm2
govc host.autostart.add  -start-order=3 -start-delay=30 vm3

Tested on ESXi 6.0.0


dougm:
@weinrea nice, thanks

PR #722, opened 2017-05-22
Allow caller to supply custom tail behavior

jzt:
This change defines a function that allows a user to supply their own custom `tail` behavior. the original `Tail` method has been replaced by a call to this function supplying the default expected `tail` behavior.


cgtexmex:
lgtm

PR #728, opened 2017-06-01
Adjust message slice passed to include

jzt:
This change slides the slice window one element to the right before passing the string into the `include` function. Previously, the string being passed in was `\nmessage` and is now adjusted to `message\n`.

It also fixes a small bug where a single line was returned even if a `tail` value of `0` was specified.


dougm:
:+1:

PR #734, opened 2017-06-06
Handle object names containing a '/'

### LINKED ISSUE #592: Cannot assign network containing a slash in its name
Our vswitches are named <network>/<cidr>, which would appear to be a legal name. (Our VMWare admin reports that javascript and powershell scripts can attach them to VMs no problem)

Because of the way govc handles paths, when executing the search in find/finder.go it looks like the incoming strings are split by '/'.

Meaning that a network named 'VLAN 172.16.0.0/8' would result in find/finder looking for a folder named 'VLAN 172.16.0.0' and an object named '8 inside of it.

I tried escaping with no luck using the following forms:

```
172.16.0.0\/8
172.16.0.0\//8
172.16.0.0//8
172.16.0.0%2f8
```

I don't like this naming convention, I wouldn't have named them with slashes in their names, I am not in control of the slashes. This seems like a legit bug to me given that none of the documentation explicitly forbids vSwitch names with slash in them..

dougm:
This only supports leaf nodes in the inventory path.
For example a Network named "my-net/work".

However, this change does not handle parents containing a '/'.
For example Datacenter named "my-data/center" and Cluster named "my-clus/ter",
these paths will not work:

  /my-data/center/vm/my-vm

  host/my-clus/ter/my-host

Fixes #592

### DIFF
--- a/find/recurser.go
+++ b/find/recurser.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/vmware/govmomi/list"
 	"github.com/vmware/govmomi/object"
@@ -177,6 +178,7 @@ func (r recurser) List(ctx context.Context, s *spec, root list.Element, parts []
 		return in, nil
 	}
 
+	all := parts
 	pattern := parts[0]
 	parts = parts[1:]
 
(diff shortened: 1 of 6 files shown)


Defilan:
This would be an AMAZING fix. We are running into this issue with our Terraform deployments. 

http://riffsy.com/image/EPlH_tinygif.gif

GarinKartes:
Glad to see this is making progress, we really need this

nkirschbaum:
+1 
I just ran into this issue and currently have no fix to get around it.

PR #736, opened 2017-06-08
Add VirtualDisk and function for import vmdk

lubronzhan:


dougm:
As we discussed moving away from using CloneVM, this is done in #805

PR #750, opened 2017-06-20
Move vcsim from vmware/vic to govmomi

dougm:


dougm:
cc @vmware/vic-maintainers @vmware/kov-maintainers

PR #753, opened 2017-06-21
Move toolbox from vmware/vic to govmomi

dougm:
While toolbox could standalone in its own repo, moving to vmware/govmomi as the new home for now:

- vcsim can implement GuestOperationsManager that drives the toolbox without requiring a real VC/ESX endpoint or use of
  the guest backdoor RPC channel.

- we use the govc guest.* commands to for end-to-end toolbox testing

- govmomi+govc tests currently depend on an old ttylinux + custom (outdated) build of open-vm-tools. These tests can be
  changed to vcsim and/or any lightweight linux distro with a simple deploy of toolbox/main.go (similar to the existing
  toolbox-test.sh)

- projects consuming toolbox are also consuming govmomi, so its 1 less repo to manage, but we can still move to its own
  repo down the road if desired.

Includes some non-functional refactoring:

- push some of the vix constants and types into their own package.

- simplify some of the names (most originals were copy-n-paste of the C names)

- drop request name from the CommandHandler, nothing uses it; even in open-vm-tools,
  its only for testing StartProgramInGuest


dougm:
vmware/vic#5548

PR #762, opened 2017-06-30
added linked clone abilities (snapshot & current machine state) to vm.clone

amitbet:
I tested the linked clone capabilities of the vm.create, but i wasn't able to get vSphere to clone a snapshot, so I added the functionality into vm.clone with 2 additional cmdline params.

### DIFF
--- a/govc/vm/clone.go
+++ b/govc/vm/clone.go
@@ -50,6 +50,8 @@ type clone struct {
 	customization string
 	waitForIP     bool
 	annotation    string
+	snapshot      string
+	link          string
 
 	Client         *vim25.Client
 	Datacenter     *object.Datacenter
@@ -101,6 +103,8 @@ func (cmd *clone) Register(ctx context.Context, f *flag.FlagSet) {
 	f.StringVar(&cmd.customization, "customization", "", "Customization Specification Name")
 	f.BoolVar(&cmd.waitForIP, "waitip", false, "Wait for VM to acquire IP address")
 	f.StringVar(&cmd.annotation, "annotation", "", "VM description")
+	f.StringVar(&cmd.snapshot, "snapshot", "current", "Snapshot name in the source VM for linked clones, to be used with -link=snapshot")
+	f.StringVar(&cmd.link, "link", "none", "Creates a linked clone from snapshot or source VM, values: snapshot | machine | none")
 }
 
 func (cmd *clone) Usage() string {
@@ -112,6 +116,16 @@ func (cmd *clone) Description() string {
 
 Examples:
   govc vm.clone -vm template-vm new-vm`
+}
+
+func (cmd *clone) searchForSnapshot(snapshotList []types.VirtualMachineSnapshotTree, snapNameToFind string) *types.ManagedObjectReference {
+	for _, sn := range snapshotList {
+		if sn.Name == snapNameToFind {
+			return &sn.Snapshot
+		}
+		return cmd.searchForSnapshot(sn.ChildSnapshotList, snapNameToFind)
+	}
+	return nil
 }
 
 func (cmd *clone) Process(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


dougm:
thanks @amitbet , any issues with the CLA?

govc/vm/clone.go
```diff
@@ -101,6 +103,8 @@ func (cmd *clone) Register(ctx context.Context, f *flag.FlagSet) {
 	f.StringVar(&cmd.customization, "customization", "", "Customization Specification Name")
 	f.BoolVar(&cmd.waitForIP, "waitip", false, "Wait for VM to acquire IP address")
 	f.StringVar(&cmd.annotation, "annotation", "", "VM description")
+	f.StringVar(&cmd.snapshot, "snapshot", "current", "Snapshot name in the source VM for linked clones, to be used with -link=snapshot")
```
dougm: Could we make the default `""` instead of `"current"`?

govc/vm/clone.go
```diff
@@ -101,6 +103,8 @@ func (cmd *clone) Register(ctx context.Context, f *flag.FlagSet) {
...
+	f.StringVar(&cmd.link, "link", "none", "Creates a linked clone from snapshot or source VM, values: snapshot | machine | none")
```
dougm: Could we make the default `""` instead of `"none"` ?

govc/vm/clone.go
```diff
@@ -114,6 +118,16 @@ Examples:
   govc vm.clone -vm template-vm new-vm`
 }
 
+func (cmd *clone) searchForSnapshot(snapshotList []types.VirtualMachineSnapshotTree, snapNameToFind string) *types.ManagedObjectReference {
```
dougm: Can you use the `VirtualMachine.FindSnapshot` instead? https://github.com/vmware/govmomi/blob/17b8c9ccb7f8c7b015d44c4ea39305c970a7bf31/object/virtual_machine.go#L525

govc/vm/clone.go
```diff
@@ -315,7 +329,46 @@ func (cmd *clone) cloneVM(ctx context.Context) (*object.Task, error) {
 		PowerOn:  false,
 		Template: cmd.template,
 	}
+	//datastoreref := cmd.Datastore.Reference()
```
dougm: just remove if the code isn't used.

govc/vm/clone.go
```diff
@@ -315,7 +329,46 @@ func (cmd *clone) cloneVM(ctx context.Context) (*object.Task, error) {
...
+		//get snapshots:
+		var o mo.VirtualMachine
+		vm := cmd.VirtualMachine
+		err = vm.Properties(ctx, vm.Reference(), []string{"snapshot"}, &o)
+		if err != nil {
+			return nil, err
+		}
 
+		if o.Snapshot == nil {
+			return nil, fmt.Errorf("Given VM does not have a snapshot to clone from. Create one, or use -link=machine to create a macine linked clone")
```
dougm: typo `macine`

govc/vm/clone.go
amitbet: What should the expected behavior be for "" ?

amitbet:
Ok, I made the changes.
when using the default (snapshot=""), govc will do a snapshot clone using the current snapshot, same behavior as before, just without the string "current".

govc/vm/clone.go
```diff
@@ -316,6 +320,49 @@ func (cmd *clone) cloneVM(ctx context.Context) (*object.Task, error) {
 		Template: cmd.template,
 	}
 
+	if cmd.link == "snapshot" {
+		relocateSpec.DiskMoveType = string(types.VirtualMachineRelocateDiskMoveOptionsCreateNewChildDiskBacking)
+		cloneSpec.Location = relocateSpec
+		//get snapshots:
+		var o mo.VirtualMachine
+		vm := cmd.VirtualMachine
+		err = vm.Properties(ctx, vm.Reference(), []string{"snapshot"}, &o)
```
dougm: Thanks @amitbet , I think we could simplify this further by changing `FindSnapshot` to return CurrentSnapshot if name == "" (diff below).  Then we don't need to load the snapshot property here and validation/errors are all handled within `FindSnapshot`.  What do you think?

```diff
modified   object/virtual_machine.go
@@ -534,6 +534,10 @@ func (v VirtualMachine) FindSnapshot(ctx context.Context, name string) (Referenc
 		return nil, errors.New("No snapshots for this VM")
 	}
 
+	if name == "" {
+		return o.Snapshot.CurrentSnapshot, nil
+	}
+
 	m := make(snapshotMap)
 	m.add("", o.Snapshot.RootSnapshotList)
```

govc/vm/clone.go
```diff
@@ -316,6 +320,49 @@ func (cmd *clone) cloneVM(ctx context.Context) (*object.Task, error) {
...
+		}
+
+		//if we should use a specific snapshot
+		if cmd.snapshot != "" && cmd.snapshot != "current" {
+			snap, err := vm.FindSnapshot(ctx, cmd.snapshot)
+			if err != nil {
+				return nil, err
+			}
+
+			if snap == nil {
```
dougm: Note that `FindSnapshot` does not return an nil snapshot without having returned a non-nil error, so there's no need for this check.

dougm: I suppose to keep it that way, we'd probably want to change the diff above to:
```diff
@@ -534,6 +534,14 @@ func (v VirtualMachine) FindSnapshot(ctx context.Context, name string) (Referenc
 		return nil, errors.New("No snapshots for this VM")
 	}
 
+	if name == "" {
+		if o.Snapshot.CurrentSnapshot != nil {
+			return o.Snapshot.CurrentSnapshot, nil
+		}
+
+		return nil, errors.New("No current snapshot for this VM")
+	}
+
 	m := make(snapshotMap)
 	m.add("", o.Snapshot.RootSnapshotList)
 
```

dougm:
@amitbet I tried this out today and made a few changes, mainly:

- The '-link' flag is now a bool

- The '-snapshot' flag requires an argument.  Rather than have a default for current, add a '-C' flag to snapshot.tree to get the current name.

https://github.com/vmware/govmomi/compare/master...dougm:pr-762

Let me know what you think, thanks!

PR #768, opened 2017-07-07
Fixing tar header and content order mismatch

morris-jason:
Experienced a situation where tar headers were invalid but content was still being written to the tar archive. Consuming the guest download in a tar reader was causing EOFs in the middle of the tar stream.

### DIFF
--- a/toolbox/hgfs/archive.go
+++ b/toolbox/hgfs/archive.go
@@ -229,25 +229,21 @@ func archiveRead(u *url.URL, tr *tar.Reader) error {
 
 // archiveWrite writes the contents of the given source directory to the given tar.Writer.
 func archiveWrite(u *url.URL, tw *tar.Writer) error {
-	dir := filepath.Dir(u.Path)
 
 	return filepath.Walk(u.Path, func(file string, fi os.FileInfo, err error) error {
 		if err != nil {
 			return filepath.SkipDir
 		}
 
-		name := strings.TrimPrefix(file, dir)[1:]
-
-		header, _ := tar.FileInfoHeader(fi, name)
-
-		header.Name = name
-
-		if header.Typeflag == tar.TypeDir {
-			header.Name += "/"
+		header, err := tar.FileInfoHeader(fi, fi.Name())
+		if err != nil {
+			return err
 		}
+		header.Name = strings.TrimPrefix(
+			strings.Replace(file, filepath.Dir(u.Path), "", -1),
+			string(filepath.Separator))
 
 		var f *os.File
-
 		if header.Typeflag == tar.TypeReg && fi.Size() != 0 {
 			f, err = os.Open(file)
 			if err != nil {
@@ -258,7 +254,9 @@ func archiveWrite(u *url.URL, tw *tar.Writer) error {
 			}
 		}
 
-		_ = tw.WriteHeader(header)
+		if err := tw.WriteHeader(header); err != nil {
+			return err
+		}
 
 		if f != nil {
 			_, err = io.Copy(tw, f)
@@ -266,5 +264,6 @@ func archiveWrite(u *url.URL, tw *tar.Writer) error {
 		}
 
 		return err
+
 	})
 }


vim25/soap/client.go
```diff
@@ -686,10 +686,7 @@ func (c *Client) Download(u *url.URL, param *Download) (io.ReadCloser, int64, er
 	}
 
 	r := res.Body
-
-	if strings.HasSuffix(u.Path, "/") {
```
dougm: This was intentional, to avoid the slim but possible case of hitting the gzip header sequence when downloading a regular file.  Is there an issue with having the client use a trailing "/" to specify a directory?  

Is this what you meant by "missing backslashes"?

toolbox/hgfs/archive.go
```diff
@@ -229,42 +229,37 @@ func archiveRead(u *url.URL, tr *tar.Reader) error {
...
-		_ = tw.WriteHeader(header)
+		if err := tw.WriteHeader(header); err != nil {
+			return err
+		}
 
-		if f != nil {
-			_, err = io.Copy(tw, f)
-			_ = f.Close()
+		content, err := os.Open(file)
+		defer content.Close()
```
dougm: This will panic if os.Open returns an error.
And why the change to use defer?

toolbox/hgfs/archive.go
```diff
@@ -229,42 +229,37 @@ func archiveRead(u *url.URL, tr *tar.Reader) error {
...
-		if header.Typeflag == tar.TypeReg && fi.Size() != 0 {
-			f, err = os.Open(file)
-			if err != nil {
-				if os.IsPermission(err) {
-					return nil
-				}
-				return err
-			}
+		if header.Typeflag != tar.TypeReg {
+			return nil
```
dougm: Then we don't get tar entries for symlink or directories.

toolbox/hgfs/archive.go
```diff
@@ -229,42 +229,37 @@ func archiveRead(u *url.URL, tr *tar.Reader) error {
...
 		}
 
-		_ = tw.WriteHeader(header)
+		if err := tw.WriteHeader(header); err != nil {
+			return err
```
dougm: Was there a case where WriteHeader returned an error related to the header itself?  Otherwise, if the pipe was closed, then the io.Copy below would return the same error.

toolbox/hgfs/archive.go
```diff
@@ -229,42 +229,37 @@ func archiveRead(u *url.URL, tr *tar.Reader) error {
...
 
-		if f != nil {
-			_, err = io.Copy(tw, f)
-			_ = f.Close()
+		content, err := os.Open(file)
+		defer content.Close()
+		if err != nil {
+			return err
+		}
+		if _, err := io.Copy(tw, content); err != nil {
```
dougm: Why add the if statement here, rather than just return error as the original code did?

toolbox/hgfs/archive.go
```diff
@@ -229,42 +229,37 @@ func archiveRead(u *url.URL, tr *tar.Reader) error {
 
 // archiveWrite writes the contents of the given source directory to the given tar.Writer.
 func archiveWrite(u *url.URL, tw *tar.Writer) error {
-	dir := filepath.Dir(u.Path)
 
 	return filepath.Walk(u.Path, func(file string, fi os.FileInfo, err error) error {
 		if err != nil {
-			return filepath.SkipDir
+			return err
```
dougm: This was intentional, behaves like the tar command (tho tar prints a warning in this case).

toolbox/hgfs/archive.go
morris-jason: Yeah that was supposed to go after the write. Fixed.

toolbox/hgfs/archive.go
morris-jason: reverted.

vim25/soap/client.go
morris-jason: I'll add it back in and just guarantee the user enters a backslash on my side.

toolbox/hgfs/archive.go
morris-jason: Didn't know it would panic. reverting.

morris-jason:
Updated the pr to be more accurate. Tests now passing.

toolbox/hgfs/archive.go
morris-jason: I was having issues on the consuming side with tar readers receiving EOFs for tar headers. I suspect there was an error writing the header, but the copy below still succeeded. That's why I'm returning here if there's an error.

toolbox/hgfs/archive.go
```diff
@@ -229,42 +229,36 @@ func archiveRead(u *url.URL, tr *tar.Reader) error {
...
+			string(filepath.Separator))
 
-		var f *os.File
+		if err := tw.WriteHeader(header); err != nil {
+			return err
+		}
 
 		if header.Typeflag == tar.TypeReg && fi.Size() != 0 {
-			f, err = os.Open(file)
+			f, err := os.Open(file)
```
dougm: Now the file isn't closed at all, so this leaks file descriptors.

The ignored errors from FileInfoHeader and WriteHeader were intentional, the additional checks aren't needed afaict.

And didn't see a response to this question in the first review "Can't tell how this is different without trying it, or what exactly was broken. Can you give some examples?"

morris-jason: I was having issues on the consuming side with tar readers receiving EOFs for tar headers. I suspect there was an error writing the header, but the copy still still succeeded.

toolbox/hgfs/archive.go
dougm: You would get an "unexpected EOF" if InitializeTransferFromGuest was called without the trailing "/".  Otherwise, I'm not seeing why these changes are needed.  Just want to understand what the issue(s) are.

morris-jason:
Closing, reproduced the errors regardless of these changes.

PR #778, opened 2017-07-12
toolbox: SendGuestInfo before the vmx asks us to

dougm:
We send the IP (and NIC info) in response to the broadcastIP request.
There can be ~30 seconds between the Reset request (1st call) and the broadcastIP request.
For existing VMs, the IP/NIC info is cached in the vmdb and doesn't take as long.

With this change, the IP/NIC info is pushed asap and takes < 1 second

Before:
```
  guest.toolsStatus  types.VirtualMachineToolsStatus  toolsOk
  Waiting for VM ip from toolbox...30.01
```
After:
```
  guest.toolsStatus  types.VirtualMachineToolsStatus  toolsOk
  Waiting for VM ip from toolbox...0.06
```

### DIFF
--- a/toolbox/service.go
+++ b/toolbox/service.go
@@ -211,6 +211,8 @@ func (s *Service) Dispatch(request []byte) []byte {
 
 // Reset is the default Handler for reset requests
 func (s *Service) Reset([]byte) ([]byte, error) {
+	s.SendGuestInfo() // Send the IP info ASAP
+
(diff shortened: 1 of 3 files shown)


dougm:
@Alphasite @casualjim I've only used the toolbox-test.sh to test, which ensures the VM has its DHCP address before starting the toolbox.  Other apps would need to ensure that the network is up + active before the toolbox starts to take advantage of this change.

PR #782, opened 2017-07-13
toolbox: support single file download via archive handler

dougm:


dougm:
Never liked having the soap.directoryReader there, now even less with soap.archiveReader there.. may move that elsewhere, such as to the new guest/toolbox client package.

PR #786, opened 2017-07-16
Tolerate repeated Close for file follower

hickeng:
Allow repeated calls to Close for datastore file follower without panic.


dougm:
lgtm

hickeng:
I've checked this fix into vmware/vic vendor for now so we don't directly need to update vendored govmomi version.

PR #787, opened 2017-07-16
Add view and filter support to object.collect command

dougm:
Refactor property wait related methods:

- Add property.WaitFilter helper

- Expose a more generic property.WaitForUpdates

- Remove property.WaitForView (use property.WaitForUpdates instead)

- Refactor event collector to use property.WaitForUpdates

- Add TraversalSpec method to *View types

### DIFF
--- a/event/processor.go
+++ b/event/processor.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -24,8 +24,6 @@ import (
 	"github.com/vmware/govmomi/view"
 	"github.com/vmware/govmomi/vim25/types"
 )
-
-const latestPageProp = "latestPage"
 
 type tailInfo struct {
 	t         *eventTailer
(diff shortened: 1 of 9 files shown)


event/processor.go
```diff
@@ -81,53 +79,47 @@ func (p *eventProcessor) run(ctx context.Context, tail bool) error {
...
+	props := []string{"latestPage"}
 
-		count := 0
-		// Retrieve the property from the objects in the ListView
-		err = property.WaitForView(ctx, property.DefaultCollector(p.mgr.Client()), listView.Reference(), collectors[0], []string{latestPageProp}, func(c types.ManagedObjectReference, pc []types.PropertyChange) bool {
-			if err = p.process(c, pc); err != nil {
+	if len(collectors) == 1 {
+		// only one object to follow, don't bother creating a view
+		return property.Wait(ctx, c, collectors[0], props, func(pc []types.PropertyChange) bool {
+			if err := p.process(collectors[0], pc); err != nil {
```
hickeng: Not a problem in current code, but what _would_ happen if outside code changed the `collectors[0]` entry?

event/processor.go
```diff
@@ -81,53 +79,47 @@ func (p *eventProcessor) run(ctx context.Context, tail bool) error {
...
 
-	// only one object to follow
-	err = property.Wait(ctx, property.DefaultCollector(p.mgr.Client()), collectors[0], []string{latestPageProp}, func(pc []types.PropertyChange) bool {
-		if err = p.process(collectors[0], pc); err != nil {
-			return false
-		}
+	defer list.Destroy(context.Background())
+
+	ref := list.Reference()
+	filter := new(property.WaitFilter).Add(ref, collectors[0].Type, props, list.TraversalSpec())
```
hickeng: Is the Type for all of the desired morefs to monitor guaranteed to be the same as `collectors[0]`? If so maybe a comment or assertion of this in addObject or during `collectors` construction.

toolbox/toolbox-test.sh
```diff
@@ -103,9 +105,13 @@ if [ -z "$(govc find / -type m -name "$vm")" ] ; then
   govc datastore.upload config.iso "$vm/config.iso" >/dev/null
   device=$(govc device.cdrom.add -vm "$vm")
   govc device.cdrom.insert -vm "$vm" -device "$device" "$vm/config.iso"
+
+  vm_path="$(govc find / -type m -name "$vm")"
+else
+  govc object.collect -s "$vm_path" -guest.toolsStatus toolsNot* # wait for previous toolbox to unregister
```
hickeng: is this something we could check from the guest side to avoid accidentally running two toolboxes?

event/processor.go
dougm: collectors is local to the run() method, and collectors[0] passes a copy of the types.ManagedObjectReference, so shouldn't be any issues there.

event/processor.go
dougm: Yes, they are all of type HistoryCollector and these are all non-exported methods that ensure the input is of type HistoryCollector.

toolbox/toolbox-test.sh
dougm: Not that I know of.  The case above just prevents an annoying stream of error messages in the case of:

- toolbox{,.test} was started over ssh
- the toolbox process is killed but channel not closed
- no toolbox is running, but the vmx isn't sure yet so it won't accept a new channel
- toolbox keeps trying over and over (error message each time)
- vmx at some point gives up on the old toolbox channel and accepts the new, and everything is fine.

PR #789, opened 2017-07-18
toolbox: avoid blocking the RPC channel when transferring process IO

dougm:
Prior to this change, if toolbox.Client.Run started a command that
took longer than a minute to run, it would timeout with:

  govc: ServerFaultCode: A general system error occurred: vix error codes = (3016, 0).

With the fault type returned in this case, the client can retry, rather than having to poll
for process status.

### DIFF
--- a/guest/toolbox/client.go
+++ b/guest/toolbox/client.go
@@ -39,6 +39,30 @@ type Client struct {
 	Authentication types.BaseGuestAuthentication
 }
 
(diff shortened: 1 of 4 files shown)


dougm:
Example:

```
govc guest.run -d - bash -c "'date && sleep 120 && date'"
```

Prior to this change that would have timed with:

 ```
govc: ServerFaultCode: A general system error occurred: vix error codes = (3016, 0)
```

PR #810, opened 2017-08-03
vcsim: add FileQuery support to datastore search

dougm:
- Create .vmdk files for VirtualDisk devices

- Make use of object.DatastorePath


cgtexmex:
LGTM

PR #828, opened 2017-08-29
govc/events: read -json flag and output events as json

adamdecaf:
I was trying to use the `-json` flag that shows up in `govc events --help`, but wasn't seeing it being used. This patch formats each event as a new json object per line.

I'd be happy to hop on the slack as this patch probably doesn't meet the project requirements.

### DIFF
--- a/govc/events/command.go
+++ b/govc/events/command.go
@@ -18,8 +18,10 @@ package events
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


dougm:
@adamdecaf thanks, any issues signing the CLA?  Looks like the CI build failed too.  I can review+help this week.

adamdecaf:
@dougm I haven't tried to sign the CLA. Mostly I was waiting to see if this change would be accepted.

In reading the code it was hard to find out the proper way to make a json response so I was just going about it without all the wrappers. Is there a better way to render the same response as json? 

I'll fix up the `go fmt` issue.

dougm:
@adamdecaf yes, we should support `-json` in the events command.  It it slightly trickier than the others due to the `-f` follow option.  Other commands implement the flags.OutputWriter interface rather than checking the value of `JSON` directly.  For example, datastore.info: https://github.com/vmware/govmomi/blob/master/govc/datastore/info.go#L119

For the events command, we could just return an error if both `-f` and `-json` are specified for now.  Maybe later consider another option if there is a need to support both.

govc/events/command.go
```diff
@@ -96,14 +98,35 @@ func (cmd *events) printEvents(ctx context.Context, obj *types.ManagedObjectRefe
...
 
-		fmt.Fprintf(os.Stdout, "[%s] [%s] %s\n",
-			event.CreatedTime.Local().Format(time.ANSIC),
-			cat,
-			msg)
+		when := event.CreatedTime.Local().Format(time.ANSIC)
+		if cmd.DatacenterFlag.OutputFlag.JSON {
+			err := printEventAsJson(os.Stdout, when, cat, msg)
+			if err != nil {
+				// do something
```
dougm: Just returning the error would be fine.

govc/events/command.go
```diff
@@ -96,14 +98,35 @@ func (cmd *events) printEvents(ctx context.Context, obj *types.ManagedObjectRefe
...
+			}
+		} else {
+			printEventAsStdout(os.Stdout, "[%s] [%s] %s\n", when, cat, msg)
+		}
+
```
dougm: no need to add a blank line here.

adamdecaf:
@dougm Signed!

PR #851, opened 2017-09-26
vim25: Move internal stuff to internal package

anfernee:
Move internal methods/types to internal package.

### DIFF
--- a/govc/host/esxcli/command.go
+++ b/govc/host/esxcli/command.go
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/govmomi/internal"
 )
 
 type Command struct {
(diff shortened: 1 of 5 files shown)


anfernee:
@dougm yeah, sure. that makes sense.

govc/host/esxcli/command.go
```diff
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/govmomi/internal"
```
anfernee: @dougm if I change it to `github.com/vmware/govmomi/vim25/internal` then it's not importable from here in `github.com/vmware/govmomi/govc`

internal package spec described here: https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit

anfernee: we can do `internal/vim25` though. it's not conflicting with methods/types, etc.

dougm: ok thanks @anfernee , let's stick with the top-level internal for now.

Issue #821, opened 2017-08-20
Error adding user


Akasurde:
Adding new user raises following error

  ```
  managed object not found: UserDirectory:UserDirectory
  ```

anfernee:
@Akasurde Same thing here. Can you clarify what command did you run, and what version of vcsim did you use?

Akasurde:
@anfernee Thanks for response.

I am facing this issue while using vcsim with Ansible module called `vmware_local_user_manager.py`

`vmware_local_user_manager.py` uses an API 

```
content.userDirectory.RetrieveUserGroups(...)
```
This is where vcsim fails.

You can reproduce this using following Ansible Playbook against latest vcsim code.

```yaml
---
- name: Add User
  hosts: localhost
  tasks:
    - name: Add user
      vmware_local_user_manager:
        validate_certs: false
        hostname: 0.0.0.0
        username: user
        password: pass
        local_user_name: dev_0001
        local_user_password: dev123
        local_user_description: Sample Development user
        state: present
```

dougm:
Fixed in #855

PR #855, opened 2017-10-02
Misc features in vcsim 

### LINKED ISSUE #819: Unable to add portgroup
While adding portgroup, I get following error message 

```
HostNetworkSystem:hostnetworksystem-85 does not implement: UpdateNetworkConfig
```

anfernee:
- Snapshot support
- UpdateNetworkConfig support
- UserDirectory support

Fixes #819 #821 (I assume it will fix those, not exactly sure because of lack of context)


simulator/virtual_machine.go
```diff
@@ -639,3 +723,54 @@ func (vm *VirtualMachine) ShutdownGuest(c *types.ShutdownGuest) soap.HasFault {
...
+
+	if tree == nil {
+		return result
+	}
+
+	for _, ss := range tree {
+		result = append(result, ss.Snapshot)
+		result = append(result, allSnapshotsInTree(ss.ChildSnapshotList)...)
+	}
+	return result
```
dougm: style: add an empty line between the statements (there's a few of these in the PR), we do that in most cases for return too:
`git grep -A1 -B2 'return ' -- simulator`

simulator/virtual_machine_test.go
```diff
@@ -446,3 +446,89 @@ func TestShutdownGuest(t *testing.T) {
...
+
+	finder := find.NewFinder(c.Client, false)
+	finder.SetDatacenter(object.NewDatacenter(c.Client, esx.Datacenter.Reference()))
+
+	vms, err := finder.VirtualMachineList(ctx, "*")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	vm := vms[0]
```
dougm: Note that for the test cases where you can use any instead of a type, you can avoid the Finder hassle with the `Map.Any` function:

```go
vm := object.NewVirtualMachine(c, Map.Any("VirtualMachine").Reference())
```

Akasurde:
@anfernee @dougm I look forward for these changes and will test with Ansible. Thanks.

govc/test/vm_snapshot.bats
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+test_vm_snapshot() {
```
dougm: ok to call the file and the func/test names just "snapshot" without the "vm" prefix, since the command doesn't have it either.

govc/test/vm_snapshot.bats
```diff
@@ -0,0 +1,124 @@
...
+  test_vm_snapshot $vm
+
+}
+
+@test "vm.snapshot" {
+  esx_env
+
+  vm=$(new_ttylinux_vm)
+
+  # test_vm_snapshot $vm
```
dougm: Is this test not working against ESX or was just taking too long to run?  You can use this to run a single test: https://github.com/sstephenson/bats/pull/117

(even nicer with bats.el :)

govc/test/vm_snapshot.bats
```diff
@@ -0,0 +1,124 @@
...
+  [ "$result" -eq 1 ]
+}
+
+@test "vm.snapshot vcsim" {
+  vcsim_env
+
+  vm=$(new_empty_vm)
+
+  test_vm_snapshot $vm
+
```
dougm: whoops, don't want a blank line there

simulator/user_directory.go
```diff
@@ -0,0 +1,77 @@
...
+	return u
+}
+
+func (u *UserDirectory) RetrieveUserGroups(req *types.RetrieveUserGroups) soap.HasFault {
+	body := &methods.RetrieveUserGroupsBody{}
+	var res []types.BaseUserSearchResult
+
+	for _, ug := range u.userGroup {
+		if req.FindUsers && !ug.Group || req.FindGroups && ug.Group {
+			compare := compareFunc(req.SearchStr, req.ExactMatch)
```
dougm: Looks like you could move this outside of the for loop, so it only needs to be called once.

simulator/user_directory.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	body.Res = &types.RetrieveUserGroupsResponse{
+		Returnval: res,
+	}
+
+	return body
+}
+
+func compareFunc(compared string, exactly bool) func(string) bool {
```
dougm: nice use of closures!

govc/test/vm_snapshot.bats
anfernee: I forgot to re-enable that..

simulator/user_directory.go
anfernee: good catch!

Issue #860, opened 2017-10-05
vcsim: AuthorizationManager support


Akasurde:
Unable to create role using `govc` and latest `govcsim`

```
$ govc role.ls
govc: ServerFaultCode: ManagedObjectNotFound

$ govc role.create MyRole
govc: ServerFaultCode: ManagedObjectNotFound
```

dougm:
I was discussing the lack of vcsim authz support with @lcastellano yesterday.  I'll open a PR that adds the properties for now, so the queries will work.  We can look at implementing the methods in separate PRs.

PR #862, opened 2017-10-05
vcsim: add AuthorizationManager

dougm:
Initial AuthorizationManager support for querying roles and permissions.

Towards #860


dougm:
esx.RoleList was captured against 6.5U1 via: https://gist.github.com/dougm/242dadc4f125237d51585444f083be23

Issue #864, opened 2017-10-07
Fix CustomFieldsManager.FindKey method signature


joaodrp:
The `CustomFieldsManager.FindKey` method returns the key of a given `CustomFieldDef`. The search is performed by comparing the input string argument with each `CustomFieldDef.Name` attribute. 

Therefore, the input string argument name must be `name` instead of `key`:

~~`func (m CustomFieldsManager) FindKey(ctx context.Context, key string) (int32, error)`~~
`func (m CustomFieldsManager) FindKey(ctx context.Context, name string) (int32, error)`

See https://github.com/vmware/govmomi/issues/854#issuecomment-333964457.

PR #865, opened 2017-10-07
Fix CustomFieldsManager.FindKey method signature

joaodrp:
Rename string input argument from 'key' to 'name'.

The FindKey method returns the key of a given CustomFieldDef. The
search is performed by comparing the input string argument with
each CustomFieldDef 'Name' attribute. Therefore, the input string
argument must be named 'name' instead of 'key'.

Resolves: #864


dougm:
thanks @joaodrp

PR #869, opened 2017-10-10
vcsim: Implement add/update/remove roles

anfernee:
Fixes: #860

### DIFF
--- a/simulator/authorization_manager.go
+++ b/simulator/authorization_manager.go
@@ -25,8 +25,16 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
+var systemPrivileges = []string{
+	"System.Anonymous",
+	"System.View",
+	"System.Read",
+}
+
 type AuthorizationManager struct {
 	mo.AuthorizationManager
+
+	nextId int32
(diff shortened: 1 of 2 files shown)


simulator/authorization_manager.go
```diff
@@ -96,3 +104,89 @@ func (m *AuthorizationManager) RetrieveRolePermissions(req *types.RetrieveRolePe
...
+
+	return body
+}
+
+func updateSystemPrivileges(privileges []string) []string {
+OUT:
+	for _, spr := range systemPrivileges {
+		for _, pr := range privileges {
+			if spr == pr {
+				continue OUT
```
dougm: Curious about the use of a label.  Isn't it the same in this case if you remove the label and just `break` here instead?

anfernee: yes, i can. but I have to use another boolean to judge whether to append it to the list. and this has label. i don't know. I picked this because it has less LOC...

dougm: Ok fine to leave it as-is, I was just curious (I haven't used labels before either)

PR #876, opened 2017-10-13
Fix govc events output

dougm:
With d4d94f4, 'govc events' outputs the following due to Printf usage:

  "[[Fri Oct 13 01:15:06 2017 info User root@192.168.2.1 logged in as govc/0.15.0]] [%!s(MISSING)] %!s(MISSING)"

- Change json field names to uppercase, as we do with other govc commands

- Add "Object" field to json and avoid printing the object header when '-json' is given

- Use time.Time for '-json', as we do in other commands

- A few renames
  - CreatedAt -> CreateTime (same as the Event field name, other types use this name too)
  - "write" -> "print" (other commands use "write")
  - "ev" -> "record"

### DIFF
--- a/govc/events/command.go
+++ b/govc/events/command.go
@@ -35,6 +35,8 @@ import (
 type events struct {
 	*flags.DatacenterFlag
 
+	write func(io.Writer, *record) error
+
 	Max   int32
 	Tail  bool
 	Force bool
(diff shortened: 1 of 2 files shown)


dougm:
@adamdecaf can you review?

The quick fix would have been the patch below, ended up making a few other changes while I was in there.

```diff
modified   govc/events/command.go
@@ -123,8 +123,8 @@ func printEventAsJson(w io.Writer, when, cat, msg string) error {
 	return e.Encode(ev{when, cat, msg})
 }

-func printEventAsStdout(w io.Writer, format string, args ...string) {
-	fmt.Fprintf(w, format, args)
+func printEventAsStdout(w io.Writer, format string, args ...interface{}) {
+	fmt.Fprintf(w, format, args...)
 }
```

adamdecaf:
👍 Nice catch.

PR #884, opened 2017-10-19
vcsim: validate authz privilege ids

dougm:
The privilege list is static, add check that the IDs are valid.

Fold the System.* ID merging into the validation loop.

### DIFF
--- a/simulator/authorization_manager.go
+++ b/simulator/authorization_manager.go
@@ -17,6 +17,8 @@ limitations under the License.
 package simulator
 
 import (
+	"strings"
+
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/simulator/esx"
 	"github.com/vmware/govmomi/vim25/methods"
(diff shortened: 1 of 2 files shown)


simulator/authorization_manager.go
```diff
@@ -215,15 +232,26 @@ func (m *AuthorizationManager) RemoveAuthorizationRole(req *types.RemoveAuthoriz
...
-		for _, pr := range privileges {
-			if spr == pr {
-				continue OUT
-			}
+func (m *AuthorizationManager) privIDs(ids []string) ([]string, *soap.Fault) {
+	system := make(map[string]struct{}, len(m.system))
+
+	for _, id := range ids {
+		if _, ok := m.privileges[id]; !ok {
+			return nil, Fault("", &types.InvalidArgument{InvalidProperty: "privIds"})
```
anfernee: ok.. so privileges in this list is reserved? (cannot be used?)

dougm: See: http://pubs.vmware.com/vsphere-6-5/index.jsp#com.vmware.wssdk.apiref.doc/vim.AuthorizationManager.html

"They are statically defined and never change for a single version of a product"

PR #886, opened 2017-10-20
Support import ova/ovf by URL

### LINKED ISSUE #651: URL supports for ova.import
I tried the latest govc CLI to upload an OVA at one web site with certain URL, but not working for me. Seem ova file has to be on the same machine as govc CLI.

Deploy ova from URL is supported in standard vsphere cli and web application. Can we have the same feature here?

thanks
-weiwen

anfernee:
Fixes #651

### DIFF
--- a/govc/importx/archive.go
+++ b/govc/importx/archive.go
@@ -18,16 +18,21 @@ package importx
 
 import (
 	"archive/tar"
+	"bytes"
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path"
-	"path/filepath"
+	"strings"
 
 	"github.com/vmware/govmomi/ovf"
+	"github.com/vmware/govmomi/vim25"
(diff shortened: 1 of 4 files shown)


anfernee:
Not ready for review. One more enhancement to do.

govc/importx/archive.go
```diff
@@ -127,23 +122,61 @@ func (t *TapeArchive) Open(name string) (io.ReadCloser, int64, error) {
...
+
+func (o Opener) OpenFile(path string) (io.ReadCloser, int64, error) {
+	if isRemotePath(path) {
+		return o.OpenRemote(path)
+	}
+	return o.OpenLocal(path)
+}
+
+func (o Opener) OpenRemote(url string) (io.ReadCloser, int64, error) {
+	resp, err := http.Get(url)
```
dougm: We want to use the command's vim25.Client here, so the TLS config applies (`-k`, `tls-ca-certs`, `-tls-known-hosts` flags, etc.)

anfernee: got it. will update it.

PR #894, opened 2017-10-24
vcsim: Generate similar ref value like VC

### LINKED ISSUE #889: vcsim generate Host moref different from a VC
Hi,

A VC has Host moref like the following.
HostSystem:host-10

Vcsim generates this info slightly different.
HostSystem:hostsystem-52

It will be ideal if vcsim generates Host moref in the same format as a VC. 

Thanks,
Xavier

anfernee:
Fixes #889

### DIFF
--- a/simulator/registry.go
+++ b/simulator/registry.go
@@ -26,6 +26,19 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
+// This is a map from a reference type name to a reference value name prefix.
+// It's a convention that VirtualCenter follows. The map is not complete, but
+// it should cover the most popular objects.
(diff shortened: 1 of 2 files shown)


simulator/registry.go
```diff
@@ -59,6 +68,15 @@ func TypeName(item mo.Reference) string {
 	return reflect.TypeOf(item).Elem().Name()
 }
 
+// ValuePrefix returns the value name prefix of a given object
+func ValuePrefix(typeName string) string {
```
dougm: No need to export this.  We could also change TypeName -> typeName while you're at it, no need to export that func either.

PR #895, opened 2017-10-24
govc: Fix VM clone when source doesn't have vNics

### LINKED ISSUE #880: "No network device found" cloning from a template without network adapter and govc
Every time i try to clone a vm from a template using govc I get "No network device found". The original template has not network device. I dont know if it is correct or there is another workaround for that operacion. Basically I am trying to do a "deploy from template"

anfernee:
Fixes #880

### DIFF
--- a/govc/vm/clone.go
+++ b/govc/vm/clone.go
@@ -270,12 +270,12 @@ func (cmd *clone) Run(ctx context.Context, f *flag.FlagSet) error {
 }
 
 func (cmd *clone) cloneVM(ctx context.Context) (*object.Task, error) {
-
 	// search for the first network card of the source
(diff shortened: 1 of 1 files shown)


dougm:
Maybe add these few (w/ * added next to them)
```console
% govc ls -i /... | cut -d- -f1 | sort -u
ClusterComputeResource:domain *
Datacenter:datacenter
Datastore:datastore
DistributedVirtualPortgroup:dvportgroup *
Folder:group *
HostSystem:host
Network:network
ResourcePool:resgroup
VirtualMachine:vm
VmwareDistributedVirtualSwitch:dvs *
```

anfernee:
@dougm you commented on the wrong PR.. but got it.

PR #898, opened 2017-10-25
import.spec not to assign deploymentOption

### LINKED ISSUE #487: import.spec should not assume Deployment param always exists for OVA/OVFs
It looks like the **import.spec** always assumes there's a **Deployment Type** param defined which is not always the case. This is something that needs to be inspected by parsing the OVF definition to see what options are available. 

The issue is that if you simply generate the spec and then modify just the OVF params and don't realize this, you'll get an fault when running import.ova command:

```
└─[0] <> govc import.ova -options=vesxi3.json /Volumes/Storage/Images/nested_esxi_appliance_v6.0u2.ova
govc: ServerFaultCode: A specified parameter was not correct: cisp.deploymentOption
```

Here's output of the import.spec:

```
└─[1] <> govc import.spec /Volumes/Storage/Images/nested_esxi_appliance_v6.0u2.ova | python -m json.tool
{
    "Deployment": "small",
    "DiskProvisioning": "thin",
    "IPAllocationPolicy": "dhcpPolicy",
    "IPProtocol": "IPv4",
    "InjectOvfEnv": false,
    "Name": null,
    "PowerOn": false,
    "PropertyMapping": [
        {
            "Key": "guestinfo.hostname",
            "Value": "esxi-1.primp-industries.com"
        },
        {
            "Key": "guestinfo.ipaddress",
            "Value": "192.168.1.190"
        },
        {
            "Key": "guestinfo.netmask",
            "Value": "255.255.255.0"
        },
        {
            "Key": "guestinfo.gateway",
            "Value": "192.168.1.1"
        },
        {
            "Key": "guestinfo.dns",
            "Value": "192.168.1.1"
        },
        {
            "Key": "guestinfo.domain",
            "Value": "primp-industries.com"
        },
        {
            "Key": "guestinfo.ntp",
            "Value": "192.168.1.1"
        },
        {
            "Key": "guestinfo.syslog",
            "Value": "192.168.1.150"
        },
        {
            "Key": "guestinfo.password",
            "Value": ""
        },
        {
            "Key": "guestinfo.ssh",
            "Value": "true"
        },
        {
            "Key": "guestinfo.createvmfs",
            "Value": "false"
        }
    ],
    "WaitForIP": false
}
```

Here's the OVA that I was using: https://bintray.com/artifact/download/photon-controller/esxi-appliances/nested_esxi_appliance_v6.0u2.ova

anfernee:
We used to forcefully assign deploymentOption among [small, medium, large]
when we call import.spec on an ova/ovf that doesn't include
deploymentOption. This could generate an option file that cannot be
accepted by import.ova or import.ovf. After this change we only assign
deploymentOption when it's included in ova/ovf file.

Fixes #487

### DIFF
--- a/govc/importx/options.go
+++ b/govc/importx/options.go
@@ -38,7 +38,7 @@ type Network struct {
 
 type Options struct {
 	AllDeploymentOptions []string `json:",omitempty"`
-	Deployment           string
+	Deployment           string   `json:",omitempty"`
 
 	AllDiskProvisioningOptions []string `json:",omitempty"`
 	DiskProvisioning           string
--- a/govc/importx/spec.go
+++ b/govc/importx/spec.go
@@ -29,9 +29,6 @@ import (
 )
 
 var (
-	// all possible ovf property values
-	// the first element being the default value
-	allDeploymentOptions         = []string{"small", "medium", "large"}
 	allDiskProvisioningOptions   = []string{"thin", "monolithicSparse", "monolithicFlat", "twoGbMaxExtentSparse", "twoGbMaxExtentFlat", "seSparse", "eagerZeroedThick", "thick", "sparse", "flat"}
 	allIPAllocationPolicyOptions = []string{"dhcpPolicy", "transientPolicy", "fixedPolicy", "fixedAllocatedPolicy"}
 	allIPProtocolOptions         = []string{"IPv4", "IPv6"}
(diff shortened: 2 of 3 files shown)


anfernee:
@dougm oops, forgot that.

anfernee:
@dougm added another change that fixes the broken boolean value in properties.

PR #899, opened 2017-10-25
Switch to kr/pretty package for the -dump flag

dougm:
The kr/pretty package is much closer to Go code that can compile compared to go-spew.
I'd been using the modified version of kr/pretty to generate simulator templates in the past.
But, had done with either a throwaway main.go or temporary hacks to existing govc commands.
The *.info and object.collect commands can now be used with the '-dump' flag to generate these templates.


dougm:
See doc updates on using the `-dump` flag to generate vcsim templates.

cc @joshw

simulator/esx/authorization_manager.go
```diff
@@ -18,7 +18,10 @@ package esx
 
 import "github.com/vmware/govmomi/vim25/types"
 
-var RoleList = []types.AuthorizationRole{
+// RoleList is the default template for the AuthorizationManager roleList property.
+// Capture method:
+//   govc object.collect -s -dump AuthorizationManager:ha-authmgr roleList
```
anfernee: this comment is really helpful. :+1:

PR #901, opened 2017-10-27
simulator: add vm.Runtime.BootTime

MarsBighead:
simulator: add vm.Runtime.BootTime

### DIFF
--- a/simulator/virtual_machine.go
+++ b/simulator/virtual_machine.go
@@ -48,6 +48,7 @@ type VirtualMachine struct {
 func NewVirtualMachine(parent types.ManagedObjectReference, spec *types.VirtualMachineConfigSpec) (*VirtualMachine, types.BaseMethodFault) {
 	vm := &VirtualMachine{}
 	vm.Parent = &parent
+	now := time.Now()
 
 	if spec.Name == "" {
 		return nil, &types.InvalidVmConfig{Property: "configSpec.name"}
@@ -99,6 +100,7 @@ func NewVirtualMachine(parent types.ManagedObjectReference, spec *types.VirtualM
 
 	vm.Runtime.PowerState = types.VirtualMachinePowerStatePoweredOff
 	vm.Runtime.ConnectionState = types.VirtualMachineConnectionStateConnected
+	vm.Runtime.BootTime = &now
 	vm.Summary.Runtime = vm.Runtime
 
 	vm.Summary.QuickStats.GuestHeartbeatStatus = types.ManagedEntityStatusGray


dougm:
Thanks, maybe we can set it here though ?
https://github.com/vmware/govmomi/blob/fb22d273209da5280c26a251b6451d62334aec1c/simulator/virtual_machine.go#L474-L480

MarsBighead:
I think it's ok, and we should take the simulator state into consideration.

dougm:
I agree that `Runtime.BootTime` should be set, I was just suggesting it be done in the same place as where we set `Runtime.Summary.BootTime`

MarsBighead:
But actually it doesn't works well at that place, and I can not get the value of `VirtualMachine.Summary.Runtime.BootTime` with `govmomi/examples/virtualmachines/main.go`.
`for _, vm := range vms {
		fmt.Printf("%s: %s\n", vm.Summary.Config.Name, vm.Summary.Config.GuestFullName)
		fmt.Printf("%v: %s\n", vm.Summary.Runtime.BootTime, vm.Summary.Config.GuestFullName)
	}`
The result 
`[hbu@hbu virtualmachines]$ go run main.go 
DC0_H0_VM0: otherGuest
<nil>: otherGuest
DC0_H0_VM1: otherGuest
<nil>: otherGuest
DC0_C0_RP0_VM0: otherGuest
<nil>: otherGuest
DC0_C0_RP0_VM1: otherGuest
<nil>: otherGuest
`

MarsBighead:
The place you suggest is in the `func Run()` which is only invoked by `func PowerOnVMTask()` and func `PowerOffVMTask()`.

Here are the above two func invoking information:
`
[hbu@hbu govmomi]$ find  . -type f -name "*.go"| xargs grep PowerOnVMTask
./simulator/virtual_machine.go:func (vm *VirtualMachine) PowerOnVMTask(c *types.PowerOnVM_Task) soap.HasFault {
[hbu@hbu govmomi]$ find  . -type f -name "*.go"| xargs grep PowerOffVMTask
./simulator/virtual_machine.go:func (vm *VirtualMachine) PowerOffVMTask(c *types.PowerOffVM_Task) soap.HasFault {
` 

It seems that they are only defined and never invoked.

MarsBighead:
·vcsim· provides the mocking VMs, and it was set the PowerOffState at the `func NewVirtualMachine()`.
`
	vm.Runtime.PowerState = types.VirtualMachinePowerStatePoweredOff 
`
It is reasonable not to set the `Runtime.Summary.BootTime` value at the position because of the VM tied to power off state. Why there is no parameter to control the VM state?

dougm:
A client needs to call the PowerOnVM_Task method to change the power state, and set the BootTime.
We could add an option so `Model.Create()` will power on the VMs it creates, PR on the way for that.

PR #900, opened 2017-10-26
vcsim: various property additions

dougm:
Fixes #897

PR #906, opened 2017-10-30
vcsim: add autostart option to power on VMs

dougm:
When a VM is created, the power state is off.  The autostart option will call vm.PowerOn when true.

Related: #900


dougm:
cc @MarsBighead

PR #916, opened 2017-11-09
Add support for cpu + mem allocation to vm.change command

dougm:
Refactored resource pool related flags and simulator so we can use the same logic
for VirtualMachine resource allocation.

### DIFF
--- a/govc/flags/resource_allocation_info.go
+++ b/govc/flags/resource_allocation_info.go
@@ -1,0 +1,85 @@
+/*
+Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flags
+
+import (
+	"context"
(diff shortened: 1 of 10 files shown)


dougm:
@embano1 added resource allocation flags to the `vm.change` command, but left `vm.create` as-is for now.  Does that work for your case?

govc/flags/resource_allocation_info.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+type ResourceAllocationFlag struct {
+	cpu, mem *types.ResourceAllocationInfo
+}
+
+func NewResourceAllocationFlag(cpu, mem *types.ResourceAllocationInfo) *ResourceAllocationFlag {
+	return &ResourceAllocationFlag{cpu, mem}
+}
+
```
anfernee: this is just smart.

govc/vm/change.go
```diff
@@ -53,10 +54,38 @@ func init() {
...
+	}
+
+	if r.Limit != nil {
+		return
+	}
+
+	if r.Reservation != nil {
+		return
+	}
+
```
anfernee: do we care `ExpandableReservation` here?

dougm: Good catch.  It seems `ExpandableReservation` cannot be set to `true` for the VM fields.  I'll see about just excluding the `{cpu,mem}.expandable` flags from the vm.change command.

embano1: Yes please exclude, bc it´s a resource pool flag.

embano1:
@dougm 
vm.change would work for now. we would just run it after the VM has been successfully created, e.g. by another provisioning tool (customer uses Foreman).

thx a ton!

govc/vm/change.go
dougm: ok done.

PR #918, opened 2017-11-14
Implemented vm.upgrade operation.

pupska:
Related to https://kb.vmware.com/s/article/1010675

### DIFF
--- a/govc/vm/upgrade.go
+++ b/govc/vm/upgrade.go
@@ -1,0 +1,81 @@
+package vm
+
+import (
+	"context"
+	"flag"
+
+	"fmt"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/task"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+type upgrade struct {
+	*flags.VirtualMachineFlag
+	version int
+}
+
+func init() {
+	cli.Register("vm.upgrade", &upgrade{})
+}
+
+func isAlreadyUpgraded(err error) bool {
+	if fault, ok := err.(task.Error); ok {
+		_, ok = fault.Fault().(*types.AlreadyUpgraded)
+		return ok
+	}
+
+	return false
+}
+
+func (cmd *upgrade) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.VirtualMachineFlag, ctx = flags.NewVirtualMachineFlag(ctx)
+	cmd.VirtualMachineFlag.Register(ctx, f)
+
+	f.IntVar(&cmd.version, "version", 0, "Target vm hardware version, by default -- latest available")
+}
+
+func (cmd *upgrade) Process(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


govc/vm/upgrade.go
```diff
@@ -0,0 +1,71 @@
...
+		target_version = fmt.Sprintf("vmx-%02d", cmd.version)
+	}
+
+	task, err := vm.UpgradeVM(ctx, target_version)
+	if err != nil {
+		return err
+	}
+	err = task.Wait(ctx)
+	if err != nil {
+		if err.Error() == "Virtual machine compatibility is already up-to-date." {
```
dougm: Rather than check the message string, you should be able to use:
```go
func isAlreadyUpgraded(err error) bool {
	if fault, ok := err.(task.Error); ok {
		_, ok = fault.Fault().(*types.AlreadyUpgraded)
		return ok
	}

	return false
}
```
Then:
```go
if err != nil {
   if isAlreadyUpraded(err) {
        fmt.Println(err.Error())
   }
}
```

govc/vm/upgrade.go
```diff
@@ -0,0 +1,71 @@
...
+	govc vm.upgrade -version=$version -vm.uuid $vm_uuid`
+}
+
+func (cmd *upgrade) Run(ctx context.Context, f *flag.FlagSet) error {
+	vm, err := cmd.VirtualMachine()
+	if err != nil {
+		return err
+	}
+
+	var target_version = ""
```
dougm: golint warns: `don't use underscores in Go names; var target_version should be targetVersion`

or you can just name it `version`

govc/vm/upgrade.go
```diff
@@ -0,0 +1,71 @@
...
+
+func init() {
+	cli.Register("vm.upgrade", &upgrade{})
+}
+
+func (cmd *upgrade) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.VirtualMachineFlag, ctx = flags.NewVirtualMachineFlag(ctx)
+	cmd.VirtualMachineFlag.Register(ctx, f)
+
+	f.IntVar(&cmd.version, "version", 0, "Target vm hardware version, by default -- latest available ")
```
dougm: remove trailing space

govc/vm/upgrade.go
```diff
@@ -0,0 +1,71 @@
...
+		return err
+	}
+	err = task.Wait(ctx)
+	if err != nil {
+		if err.Error() == "Virtual machine compatibility is already up-to-date." {
+			fmt.Println("Virtual machine compatibility is already up-to-date.")
+		} else {
+			return err
+		}
+
```
dougm: remove blank line

govc/vm/upgrade.go
```diff
@@ -0,0 +1,71 @@
...
+		return err
+	}
+	return nil
+}
+
+func (cmd *upgrade) Description() string {
+	return `Upgrade VMs to latest hardware version
+
+Examples:
+	govc vm.upgrade -vm $vm_name
```
dougm: 2 space indent instead of tab to match the existing examples format.

pupska:
@dougm thanks for your comments, I've pushed updated version.

PR #936, opened 2017-11-23
Fix #933

### LINKED ISSUE #933: panic: runtime at property.(*Collector).Retrieve
I'm collecting vms like this:

```
        mgr := view.NewManager(c)

	v, err := mgr.CreateContainerView(ctx, c.ServiceContent.RootFolder, []string{"VirtualMachine"}, true)
	if err != nil {
		return nil, err
	}

	defer v.Destroy(ctx)

	// Retrieve summary property for all machines
	var vmt []mo.VirtualMachine
	err = v.Retrieve(ctx, []string{"VirtualMachine"}, []string{"summary"}, &vmt)
	if err != nil {
		return nil, err
	}

```

Then I want to see the network names from which VM and made like this:

```
        pc := property.DefaultCollector(c)

	var nets []mo.Network
	err := pc.Retrieve(ctx, vm.Network, []string{"name", "summary"}, &nets)
	if err != nil {
		return nil, err
	}

	for _, net := range nets {
		name := strings.Replace(net.Name, ".", "_", -1)
		outputNetworkNames.Put("name", name)
	}
```

Then I got:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x230becc]

goroutine 90 [running]:
github.com/elastic/beats/metricbeat/module/vsphere/vendor/github.com/vmware/govmomi/property.(*Collector).Retrieve(0xc420051370, 0x3714980, 0xc420346200, 0x0, 0x0, 0x0, 0xc42022e110, 0x1, 0x1, 0x2423000, ...)
	/home/user/project/beats/src/github.com/elastic/beats/metricbeat/module/vsphere/vendor/github.com/vmware/govmomi/property/collector.go:157 +0x50c
github.com/elastic/beats/metricbeat/module/vsphere/virtualmachine.getNetworkNames(0x3714980, 0xc420346200, 0xc4202ec580, 0x0, 0x0, 0x0, 0x4, 0x1000000007, 0x0)
```

I don't know why this is happening here. Any help?

amandahla:
Check if object references is empty then avoid panic runtime error.

### DIFF
--- a/property/collector.go
+++ b/property/collector.go
@@ -121,6 +121,10 @@ func (p *Collector) RetrieveProperties(ctx context.Context, req types.RetrievePr
 // of the specified managed objects, with the relevant properties filled in. If
 // the properties slice is nil, all properties are loaded.
 func (p *Collector) Retrieve(ctx context.Context, objs []types.ManagedObjectReference, ps []string, dst interface{}) error {
+	if len(objs) == 0 {
+		return errors.New("object references is empty")
+	}
+
 	var propSpec *types.PropertySpec
 	var objectSet []types.ObjectSpec
 


dougm:
Thanks @amandahla !

PR #939, opened 2017-11-25
Corrects host.shutdown

1995parham:
host.shutdown expects HOST as an argument.

### DIFF
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -1881,7 +1881,7 @@ Options:
 ## host.service.ls
 
 ```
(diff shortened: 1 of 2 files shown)


dougm:
Thanks @1995parham , the USAGE.md doc is actually generated by running `make doc`.  Can you add a `Usage` method to host/shutdown.go similar to: https://github.com/vmware/govmomi/blob/master/govc/dvs/add.go#L59-L61

1995parham:
Thanks @dougm for your reply, I add usage method for host shutdown command.

dougm:
@1995parham thanks!

PR #955, opened 2017-12-11
Ignore AcquireLocalTicket errors

### LINKED ISSUE #952: "ServerFaultCode: The operation is not supported on the object"
I've used govc successfully in the past, and I have production jobs using it regularly, but today I'm trying to use govc with my homelab cluster, which it doesn't seem to like:

```
bard:~ $ govc ls
govc: ServerFaultCode: The operation is not supported on the object.
bard:~ $ govc about
govc: ServerFaultCode: The operation is not supported on the object.
```

Does that mean my VMUG EVALExperience license for my 3-node vSphere + vCenter HA setup doesn't include API access?

If so, can we get a better error message in govc telling users what the actual problem is?

Or do I need to enable the API server in vCenter somewhere? I've only ever worked with vSphere clusters others have created before. This home one is the first I've put together, and perhaps I screwed it up.

Sorry for the noob questions.

Thanks!

dougm:
If no username is provided, the govc client tries to acquire a local ticket.
When invoked remotely, ESX returns an InvalidRequestFault (VC returns NotSupportedFault).
So, rather than return an error here, fallthrough to Login() with the original User to
to avoid what would be a confusing error message.

Don't try AcquireLocalTicket at all against vCenter as it is unsupported.

Fixes #952

### DIFF
--- a/govc/flags/client.go
+++ b/govc/flags/client.go
@@ -428,11 +428,15 @@ func (flag *ClientFlag) newClient() (*vim25.Client, error) {
 	m := session.NewManager(c)
 	u := flag.url.User
 
-	if u.Username() == "" {
-		// Assume we are running on an ESX or Workstation host if no username is provided
-		u, err = flag.localTicket(ctx, m)
-		if err != nil {
-			return nil, err
+	if u.Username() == "" && !c.IsVC() {
+		// If no username is provided, try to acquire a local ticket.
+		// When invoked remotely, ESX returns an InvalidRequestFault.
+		// So, rather than return an error here, fallthrough to Login() with the original User to
+		// to avoid what would be a confusing error message.
+		luser, lerr := flag.localTicket(ctx, m)
+		if lerr == nil {
+			// We are running directly on an ESX or Workstation host and can use the ticket with Login()
+			u = luser
 		}
 	}
 


govc/flags/client.go
```diff
@@ -428,11 +428,15 @@ func (flag *ClientFlag) newClient() (*vim25.Client, error) {
...
 	u := flag.url.User
 
-	if u.Username() == "" {
-		// Assume we are running on an ESX or Workstation host if no username is provided
-		u, err = flag.localTicket(ctx, m)
-		if err != nil {
-			return nil, err
+	if u.Username() == "" && !c.IsVC() {
+		// If no username is provided, try to acquire a local ticket.
+		// When invoked remotely, ESX returns an InvalidRequestFault (VC returns NotSupportedFault).
```
bradfitz: this parenthetical is slightly confusing considering the conditional on line 431 excludes this path for VC.

elaborate or remove?

bradfitz:
Code LGTM. Thanks for the fix!

govc/flags/client.go
dougm: Removed.  Thanks for reviewing Brad!

govc/flags/client.go
```diff
@@ -428,11 +428,15 @@ func (flag *ClientFlag) newClient() (*vim25.Client, error) {
...
-	if u.Username() == "" {
-		// Assume we are running on an ESX or Workstation host if no username is provided
-		u, err = flag.localTicket(ctx, m)
-		if err != nil {
-			return nil, err
+	if u.Username() == "" && !c.IsVC() {
+		// If no username is provided, try to acquire a local ticket.
+		// When invoked remotely, ESX returns an InvalidRequestFault.
+		// So, rather than return an error here, fallthrough to Login() with the original User to
+		// to avoid what would be a confusing error message.
```
anfernee: so error happens here, we login with empty user, and get something like `LoginFault`, which is less confusing. Is that accurate?

dougm: exactly, InvalidLoginFault with the appropriate localized message.

PR #966, opened 2017-12-14
Add '-rescan-vmfs' option to host.storage.info

acortiana:
Add '-refreshvmfs' option to host.storage.info

### DIFF
--- a/govc/host/storage/info.go
+++ b/govc/host/storage/info.go
@@ -67,10 +67,11 @@ type info struct {
 	*flags.HostSystemFlag
 	*flags.OutputFlag
 
-	typ       infoType
-	rescan    bool
-	refresh   bool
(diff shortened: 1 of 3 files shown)


acortiana:
You are right, Doug. RefreshStorageSystem also performs a VMFS rescan.
I implemented this feature because I didn't know this.
Maybe this feature can be useful: in the GUI it's possible to do a VMFS rescan by itself and I think that this is faster than doing a full RefreshStorageSystem call.
No problem for the name of the flag: rescan-vmfs is ok.

dougm:
@acortiana ok sounds good, I didn't know the GUI had that option.  If you can just rename the flag that'd be great.

dougm:
Thanks @acortiana !

PR #967, opened 2017-12-15
vcsim: Honor the input spec in ReConfig VM

### LINKED ISSUE #956: vcsim: does not support 'NestedHVSupported' Property
I am unable to set 'NestedHVSupported' property of virtual machine using govc and vcsim.

```
# govc vm.info -json DC0_H0_VM5 | jq '.[][0].Capability.NestedHVSupported'
null

# govc vm.change -nested-hv-enabled=True -vm DC0_H0_VM5

# echo $?
0

# govc vm.info -json DC0_H0_VM5 | jq '.[][0].Capability.NestedHVSupported'
null
```

anfernee:
Including NestedHVSupported and other values, putting them into vm.Config if they are not nil.

Fix #956

### DIFF
--- a/simulator/virtual_machine.go
+++ b/simulator/virtual_machine.go
@@ -142,6 +142,52 @@ func (vm *VirtualMachine) apply(spec *types.VirtualMachineConfigSpec) {
 		}
 	}
 
+	applyb := []struct {
+		src *bool
+		dst **bool
+	}{
+		{spec.NestedHVEnabled, &vm.Config.NestedHVEnabled},
+		{spec.CpuHotAddEnabled, &vm.Config.CpuHotAddEnabled},
+		{spec.CpuHotRemoveEnabled, &vm.Config.CpuHotRemoveEnabled},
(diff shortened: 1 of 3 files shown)


simulator/virtual_machine.go
```diff
@@ -142,6 +143,29 @@ func (vm *VirtualMachine) apply(spec *types.VirtualMachineConfigSpec) {
 		}
 	}
 
+	apply2 := []struct {
```
dougm: Can we keep this typed and avoid reflection?  For example:
```go
	applyb := []struct {
		src *bool
		dst *bool
	}{
		{spec.NestedHVEnabled, vm.Config.NestedHVEnabled},
	}

	for _, f := range applyb {
		if f.src != nil {
			f.dst = types.NewBool(*f.src)
		}
	}
```

anfernee: Sure.. Most of them are `bool`s so it's mostly what it is.

anfernee:
@dougm I tried to run that on esx, actually this only changes `vm.Config`. It won't affect `Capability`, which I think kinda make sense.

PR #969, opened 2017-12-17
Use IsFileNotFound helper in Datastore.Stat

dougm:
Avoids a possible panic when waiting for SearchDatastore to complete.


AngieCris:
Fixes https://github.com/vmware/vic/issues/6972

PR #972, opened 2017-12-18
govc: allow columns in guest login password

### LINKED ISSUE #971: govc: allow columns in guest login password
Current implementation uses strings.Split: https://github.com/vmware/govmomi/blob/e15ff586b59d4bfeb27fc5e915645db6490b6dd1/govc/vm/guest/auth.go#L42

Which means, if the password has any column, it will be be truncated, example:

user:pass:word -> user, pass instead of user, pass:word.

An improved behavior would be to use SplitN(..., 2), to ignore columns after the first split.

I will link a pull request in a minute.

Thanks.

jeremy-clerc:
fixes #971 by using SplitN instead of Split and so keeping columns in the password.

### DIFF
--- a/govc/CHANGELOG.md
+++ b/govc/CHANGELOG.md
@@ -1,5 +1,9 @@
 # changelog
 
+### 0.17.0 (undef)
+
+* Allow columns in password for guest login.
+
 ### 0.16.0 (2017-11-08)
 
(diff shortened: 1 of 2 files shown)


dougm:
Thanks @jeremy-clerc !

PR #976, opened 2017-12-21
Add option to filter events by type

dougm:
Filtering specific event types reduces the chance of missing events,
while keeping the event history collector page size on the low side.

### DIFF
--- a/govc/test/events.bats
+++ b/govc/test/events.bats
@@ -54,6 +54,21 @@ load test_helper
   run govc events vm
   assert_success
   [ ${#lines[@]} -ge $nevents ]
+
+  run govc events -type VmPoweredOffEvent -type VmPoweredOnEvent "vm/$vm"
+  [ ${#lines[@]} -eq 0 ]
+
+  run govc vm.power -on "$vm"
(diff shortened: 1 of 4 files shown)


govc/test/events.bats
```diff
@@ -54,6 +54,21 @@ load test_helper
...
+  assert_success
+
+  run govc events -type VmPoweredOffEvent -type VmPoweredOnEvent "vm/$vm"
+  [ ${#lines[@]} -eq 1 ]
+
+  run govc vm.power -off "$vm"
+  assert_success
+
+  run govc events -type VmPoweredOffEvent -type VmPoweredOnEvent "vm/$vm"
+  [ ${#lines[@]} -eq 2 ]
```
hickeng: You could do with an additional VM operation (guest operation or reconfigure or something) so that you can assert that you do _not_ get an event for it.

dougm: Several events such as VmCreatedEvent, VmMacAssignedEvent, etc, are being filtered out here.. but yes could make it more explicit within the test.

PR #977, opened 2017-12-22
Add cluster vm override commands

dougm:
- Add cluster.override.change command to set VM DRS and HA overrides

- Add cluster.override.remove to remove VM overrides

- Add cluster.override.info to view current VM overrides

- Add cluster.rule.create '-depends' flag to support VM-to-VM dependency rules

### DIFF
--- a/govc/test/cluster.bats
+++ b/govc/test/cluster.bats
@@ -99,6 +99,62 @@ load test_helper
   run govc cluster.rule.remove -cluster DC0_C0 -name pod1
   assert_success
 
-  run govc cluster.rule.remove -cluster DC0_C0 -name pod1
+  run govc cluster.rule.remove -cluster DC0_C0 -name pod1 -depends
   assert_failure # rule does not exist
+
+  run govc cluster.rule.create -cluster DC0_C0 -name my_deps -depends
+  assert_failure # requires 2 groups
+
+  run govc cluster.group.create -cluster DC0_C0 -name my_app -vm DC0_C0_RP0_VM{4,5}
+  assert_success
+
+  run govc cluster.group.create -cluster DC0_C0 -name my_db -vm DC0_C0_RP0_VM{6,7}
+  assert_success
+
+  run govc cluster.rule.create -cluster DC0_C0 -name my_deps -depends my_app my_db
+  assert_success
+}
+
+@test "cluster.vm" {
+  vcsim_env -host 4 -vm 8
+
+  run govc cluster.override.info
+  assert_success "" # no overrides == empty output
+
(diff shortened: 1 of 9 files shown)


govc/cluster/vm/change.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if vm == nil {
+		return flag.ErrHelp
+	}
+
+	cluster, err := cmd.Cluster()
+	if err != nil {
+		return err
+	}
+
```
anfernee: do we need to check cluster here as well?

govc/cluster/vm/remove.go
```diff
@@ -0,0 +1,110 @@
...
+
+	if vm == nil {
+		return flag.ErrHelp
+	}
+
+	cluster, err := cmd.Cluster()
+	if err != nil {
+		return err
+	}
+
```
anfernee: ditto.

govc/test/cluster.bats
```diff
@@ -99,6 +99,65 @@ load test_helper
...
   assert_success
 
-  run govc cluster.rule.remove -cluster DC0_C0 -name pod1
+  run govc cluster.rule.remove -cluster DC0_C0 -name pod1 -depend
   assert_failure # rule does not exist
+
+  run govc cluster.rule.create -cluster DC0_C0 -name my_deps -depend
+  assert_failure # requires 2 groups
+
+  run govc cluster.rule.create -cluster DC0_C0 -name my_deps -depend my_app my_db
```
anfernee: this is `-depend` or `-depends`?, as well as the 3 cases above this.

govc/cluster/vm/change.go
dougm: No, the cluster flag will never return `nil, nil`. VirtualMachineFlag is optional in some commands, and can have multiple values in other commands.  So it's left to the caller to decide what to do if none are specified.

govc/test/cluster.bats
dougm: Good catch, it is `-depends`, those tests were expected to fail, but they were failing because of an invalid flag name.

dougm:
Thinking about the command names, esp. after adding `cluster.vm.remove`..
Currently we have:

| UI label | govc command |
|------------|----------------------|
| VM/Host Groups | cluster.group.* |
| VM/Host Rules | cluster.rule.* |
| VM Overrides | cluster.vm.* |

Thinking `cluster.vm.*` should be renamed to `cluster.override.*` in this PR.  Thoughts?

dougm:
Sticking with the new name, code is unchanged otherwise.

PR #983, opened 2018-01-04
vcsim: add session support

### LINKED ISSUE #923: vcsim: PropertyFilters not destroyed when Session is closed
PropertyFilters are not being destroyed when user terminates the session (e.g. login, logout).

**Steps to reproduce:**
1) Login
2) CreateFilter
3) Call WaitForUpdatesEx with empty version (returns a set of filters that user created)
4) Logout
5) Login
6) CreateFilter
7) Call WaitForUpdatesEx with empty version

**Results**
Executing these steps on vcsim server, second WaitForUpdatesEx returns UpdateSet containing 2 filters

**Expected Results**
Executing these steps on real VCenter, second WaitForUpdatesEx returns UpdateSet containing 1 filter

dougm:
Sessions are now tracked across requests, supporting per-session managed objects such as
PropertyCollector, PropertyFilter, ContainerView and the SessionManager itself.

Adds TerminateSession() support and includes proper support for Logout()

Fixes #923

### DIFF
--- a/simulator/session_manager.go
+++ b/simulator/session_manager.go
@@ -17,6 +17,11 @@ limitations under the License.
 package simulator
 
 import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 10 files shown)


simulator/session_manager.go
```diff
@@ -61,16 +72,35 @@ func (s *SessionManager) Login(login *types.Login) soap.HasFault {
...
+	s.remove(ctx.Session.Key)
 	return &methods.LogoutBody{Res: new(types.LogoutResponse)}
 }
 
+func (s *SessionManager) TerminateSession(ctx *Context, req *types.TerminateSession) soap.HasFault {
+	for _, id := range req.SessionId {
+		if id == ctx.Session.Key {
+			return &methods.TerminateSessionBody{Fault_: Fault("", &types.InvalidArgument{})}
+		}
+		s.remove(id)
```
anfernee: do we want to check first then remove to avoid removing half of the sessions?

dougm: Real ESX/VC actually behaves the same way as this.

PR #987, opened 2018-01-08
Added AttachScsiLun function

tdawe:
Added AttachScsiLun function to host_storage_system


object/host_storage_system.go
```diff
@@ -161,3 +161,14 @@ func (s HostStorageSystem) MarkAsNonLocal(ctx context.Context, uuid string) (*Ta
 
 	return NewTask(s.c, res.Returnval), nil
 }
+
+func (s HostStorageSystem) AttachScsiLun(ctx context.Context, uuid string) (*types.AttachScsiLunResponse, error) {
```
dougm: Thanks @tdawe .  Can you change the method to just return `error`?  The response has no fields, so it can just be ignored.

PR #994, opened 2018-01-11
Add CloneSession support to govc and vcsim

dougm:
- Fix simulator.Service.About to support optional context param

- Don't require govc Login, ServiceContent does not require it

### DIFF
--- a/govc/session/login.go
+++ b/govc/session/login.go
@@ -1,0 +1,195 @@
+/*
+Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


govc/session/login.go
```diff
@@ -0,0 +1,191 @@
...
+	if len(output) == 0 {
+		return nil
+	}
+
+	fmt.Fprintln(w, strings.Join(output, " "))
+
+	return nil
+}
+
+func (cmd *login) Logout(ctx context.Context) error {
```
anfernee: just a question: why Logout function here, who would call this?

dougm: I expanded the comments, it's called by cli.Run()

govc/session/login.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (cmd *login) setCookie(ctx context.Context, c *vim25.Client) error {
+	url := c.URL()
+	jar := c.Client.Jar
+	cookies := jar.Cookies(url)
+	add := true
+
+	cookie := &http.Cookie{
+		Name: "vmware_soap_session",
```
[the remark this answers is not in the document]
dougm: Done

PR #997, opened 2018-01-12
vcsim: add UpdateOptions support

### LINKED ISSUE #989: vcsim: hostsystem advance configuration
vcsim does not simulate all advance configuration parameter and fails to update option.

Python reproducer

```python
from pyVim.connect import SmartConnect, Disconnect
import ssl
import atexit
from pyVmomi import vim
from ansible.module_utils.six import integer_types, iteritems, string_types
from ansible.module_utils._text import to_text

def connect():
    context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
    context.verify_mode = ssl.CERT_NONE

    si = SmartConnect(host='0.0.0.0', user='user', pwd='pass', port=443, sslContext=context)
    atexit.register(Disconnect, si)
    content = si.RetrieveContent()
    return content

def get_obj(content, vimtype, name):
    """
    Return an object by name, if name is None the
    first found object is returned
    """
    obj = None
    container = content.viewManager.CreateContainerView(
        content.rootFolder, vimtype, True)
    for c in container.view:
        if name:
            if c.name == name:
                obj = c
                break
        else:
            obj = c
            break

    container.Destroy()
    return obj

content = connect()

cluster = get_obj(content, [vim.ClusterComputeResource], 'DC0_C0')
host = cluster.host[0]
option_manager = host.configManager.advancedOption
for option in option_manager.QueryOptions():
    print("%s => %s" % (option.key, option.value))

change_option_list = []
option_key = 'Config.HostAgent.log.level'
option_value = 'verbose'
change_option_list.append(vim.option.OptionValue(key=option_key, value=option_value))
option_manager.UpdateOptions(changedValue=change_option_list)
```

Stack trace 

```
$ python vm_host_config.py
Config.HostAgent.log.level => info
Traceback (most recent call last):
  File "vm_host_config.py", line 49, in <module>
    option_manager.UpdateOptions(changedValue=change_option_list)
  File "/Volumes/data/src/ansible/venv4/lib/python2.7/site-packages/pyVmomi/VmomiSupport.py", line 580, in <lambda>
    self.f(*(self.args + (obj,) + args), **kwargs)
  File "/Volumes/data/src/ansible/venv4/lib/python2.7/site-packages/pyVmomi/VmomiSupport.py", line 386, in _InvokeMethod
    return self._stub.InvokeMethod(self, info, args)
  File "/Volumes/data/src/ansible/venv4/lib/python2.7/site-packages/pyVmomi/SoapAdapter.py", line 1357, in InvokeMethod
    raise exc
pyVmomi.SoapAdapter.ParserError: 'xml document: <httplib.HTTPResponse instance at 0x108d15488> parse error at: line:2, col:516'
```

vcsim returns

```
2018/01/09 17:54:41 OptionManager:optionmanager-38 does not implement: UpdateOptions
```

dougm:
Fixes #989

### DIFF
--- a/simulator/option_manager.go
+++ b/simulator/option_manager.go
@@ -57,3 +57,42 @@ func (m *OptionManager) QueryOptions(req *types.QueryOptions) soap.HasFault {
 
 	return body
 }
+
+func (m *OptionManager) find(key string) *types.OptionValue {
+	for _, opt := range m.Setting {
+		setting := opt.GetOptionValue()
+		if setting.Key == key {
+			return setting
+		}
+	}
+	return nil
+}
+
+func (m *OptionManager) UpdateOptions(req *types.UpdateOptions) soap.HasFault {
+	body := new(methods.UpdateOptionsBody)
+
+	for _, change := range req.ChangedValue {
+		setting := change.GetOptionValue()
+
+		// We don't currently include the entire list of default settings for ESX and vCenter,
+		// this prefix is currently used to test the failure path.
+		// Real vCenter seems to only allow new options if Key has a "config." prefix.
+		// TODO: consider behaving the same, which would require including 2 long lists of options in vpx.Setting and esx.Setting
+		if strings.HasPrefix(setting.Key, "ENOENT.") {
+			body.Fault_ = Fault("", &types.InvalidName{Name: setting.Key})
+			return body
+		}
+
+		opt := m.find(setting.Key)
+		if opt != nil {
+			// This is an existing option.
+			opt.Value = setting.Value
+			continue
+		}
+
+		m.Setting = append(m.Setting, change)
+	}
+
+	body.Res = new(types.UpdateOptionsResponse)
+	return body
+}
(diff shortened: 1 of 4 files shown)


dougm:
I had started UpdateOptions support here: https://github.com/vmware/vic/pull/6815#issuecomment-346141157

Had stashed and forgot about, because I wasn't sure then (or now) about the TODO: regarding vpx/esx Setting.  But it should be ok for now as-is.

simulator/option_manager.go
```diff
@@ -57,3 +57,41 @@ func (m *OptionManager) QueryOptions(req *types.QueryOptions) soap.HasFault {
...
+		}
+
+		// We don't currently include the entire list of default settings for ESX and vCenter,
+		// this prefix is currently used to test the failure path.
+		// Real vCenter seems to only allow new options if Key has a "config." prefix.
+		// TODO: consider behaving the same, which would require including 2 long lists of options in vpx.Setting and esx.Setting
+		if strings.HasPrefix(setting.Key, "ENOENT.") {
+			body.Fault_ = Fault("", &types.InvalidName{Name: setting.Key})
+			return body
+		}
```
anfernee: move this block before `m.find(setting.Key)`?

dougm: It's after since originally the condition was `if !strings.HasPrefix(setting.Key, "config.")`, but in the current form it does make sense to have it before.  I'll move it.

PR #1004, opened 2018-01-17
govc: fix host.info CPU usage

dougm:
Need to use CPU threads as the logical processor count, rather than CPU cores.


dougm:
cc @embano1

PR #1013, opened 2018-01-25
Leverage contexts in http uploads

erikh:
I hope this finds you well. I am working on a deploy tool to deploy to ESX hosts, I am learning that many of the contexts I pass are not used at lower levels. In short, I am using contexts in a signal handler to reap deploys on termination. It's just not doing that right now. ;)

I attempted to add this to the upload and lease calls so I can get moving on my own project. I hope it finds you well.

I use `(*http.Request).WithContext` to store the context with the request itself, and in other cases I select on the context.

Thanks,

-Erik

### DIFF
--- a/govc/importx/archive.go
+++ b/govc/importx/archive.go
@@ -182,5 +182,5 @@ func (o Opener) OpenRemote(link string) (io.ReadCloser, int64, error) {
 		return nil, 0, err
 	}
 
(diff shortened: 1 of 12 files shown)


erikh:
cc @jojimt

erikh:
I also added context fixes for Download and progress.NewReader operations.

PR #1016, opened 2018-01-26
vcsim: add EventManager support

dougm:
- Add '-dump' support to govc events command, so we can grab events as Go code

- Workaround 6.5 bug where an internal event type (HostSubSpecificationUpdateEvent) was causing an NPE

- Avoid Go 1.9.3 url.Parse error in cli.bats

- PropertyCollector should return empty arrays

### DIFF
--- a/simulator/event_manager.go
+++ b/simulator/event_manager.go
@@ -1,0 +1,386 @@
+/*
+Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package simulator
+
+import (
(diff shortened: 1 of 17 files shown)


dougm:
cc @kars7e @embano1

simulator/event_manager.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+// filePage copies the manager's latest events into the collector's page with Filter applied.
+func (c *EventHistoryCollector) fillPage(size int) {
+	l := c.page.Len()
+	delta := size - l
+
+	if delta < 0 {
+		// Shrink ring size
+		c.page = c.page.Unlink(l - delta*-1)
```
anfernee: here I suppose you only need to unlink `-delta` right?

dougm: Yes indeed, I'll fix that.

simulator/view_manager.go
```diff
@@ -149,7 +149,7 @@ func (v *ContainerView) include(o types.ManagedObjectReference) bool {
 	return v.types[o.Type]
 }
 
-func (v *ContainerView) add(root mo.Reference, seen map[types.ManagedObjectReference]bool) {
+func walk(root mo.Reference, f func(child types.ManagedObjectReference)) {
```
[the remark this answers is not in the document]
dougm: Yeah, I meant to pick another name but had the same problem ;)  The usage of it may walk the whole sub-tree, examples:
- ContainerView with with Recursive option set
- EventHistoryCollector with RecursionOption Children or RecursionOption All

It isn't exported so I left the name as-is for now.

PR #1018, opened 2018-01-26
vcsim: stats related fixes

dougm:
- Use the curl '-f' flag so the tests actually fail on 404 Not Found

- Add some Model validation to the tests

- Use Model.Count to get the current totals when stats are fetched

- Remove 'omitempty' tags from Model field that represent object types,
  making it easier to validate when an object count should be '0'

### DIFF
--- a/govc/test/vcsim.bats
+++ b/govc/test/vcsim.bats
@@ -23,17 +23,20 @@ load test_helper
 }
 
 @test "vcsim about" {
-  vcsim_env
+  vcsim_env -dc 2 -cluster 3 -vm 0 -ds 0
 
   url="https://$(govc env GOVC_URL)"
 
-  run curl -sk "$url/about"
+  run curl -skf "$url/about"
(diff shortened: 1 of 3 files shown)


govc/test/vcsim.bats
```diff
@@ -23,17 +23,20 @@ load test_helper
...
 
-  run curl -sk "$url/debug/vars"
+  run curl -skf "$url/debug/vars"
   assert_success
 
-  run curl -sk "$url/debug/vcsim/vars"
-  assert_success
+  model=$(curl -sfk "$url/debug/vars" | jq .vcsim.Model)
+  [ "$(jq .Datacenter <<<"$model")" == "2" ]
+  [ "$(jq .Cluster <<<"$model")" == "6" ]
```
anfernee: shouldn't this be 3?

dougm: vcsim will create `-cluster x -dc`, so `2 x 3 == 6` in this case.

PR #1019, opened 2018-01-27
govc: object.collect support for raw filters

dougm:
This change allows for replaying CreateFilter requests from other clients and
generating Go versions of CreateFilter requests than can be dropped into tests.

### DIFF
--- a/govc/object/collect.go
+++ b/govc/object/collect.go
@@ -22,6 +22,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"os"
 	"reflect"
 	"strings"
 	"text/tabwriter"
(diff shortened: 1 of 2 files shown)


govc/object/collect.go
```diff
@@ -272,62 +313,75 @@ func (cmd *collect) Run(ctx context.Context, f *flag.FlagSet) error {
...
 
-		v, err := m.CreateContainerView(ctx, ref, cmd.kind, true)
-		if err != nil {
+			defer v.Destroy(ctx)
+
+			for _, kind := range cmd.kind {
+				filter.Add(v.Reference(), kind, props, v.TraversalSpec())
+			}
+		}
+	} else {
```
dougm: Not that most the diff above was just moving the existing Run() body into an `if`, so we can have this `else`..

PR #1029, opened 2018-02-07
govc: fix vm.clone to use -net flag when source does not have a NIC

### LINKED ISSUE #960: govc vm.clone -net.address doesn't work, silently does nothing?
I'm trying to set an explicit MAC address for a cloned VM. If I do this:

```
govc vm.destroy --persist-session=false cloned_vm || true
govc vm.clone -vm frozenbasevm -c=2 -m=2048 -net=Untrusted -net.address=00:50:56:bb:ae:51 -on=true -link=true -ds=nfs_ds916 cloned_vm
```

... and it clones & boots and everything works, except the VM comes up with a random MAC address.

I get no error that my `-net.address` was wrong or otherwise ignored.

I'd expect it to either work, or fail and tell me why it won't work.

How can I help debug?

dougm:
- Add vcsim support for VirtualDeviceConfigSpecOperationEdit

- vcsim CloneVM now includes source VM's devices and applies VirtualMachineRelocateSpec.DeviceChange

Fixes #960

### DIFF
--- a/simulator/virtual_machine.go
+++ b/simulator/virtual_machine.go
@@ -513,19 +513,54 @@ func (vm *VirtualMachine) configureDevice(devices object.VirtualDeviceList, spec
 	return nil
 }
 
-func removeDevice(devices object.VirtualDeviceList, device types.BaseVirtualDevice) object.VirtualDeviceList {
-	var result object.VirtualDeviceList
+func (vm *VirtualMachine) removeDevice(devices object.VirtualDeviceList, spec *types.VirtualDeviceConfigSpec) object.VirtualDeviceList {
+	key := spec.Device.GetVirtualDevice().Key
 
 	for i, d := range devices {
-		if d.GetVirtualDevice().Key == device.GetVirtualDevice().Key {
-			result = append(result, devices[i+1:]...)
-			break
+		if d.GetVirtualDevice().Key != key {
+			continue
 		}
 
-		result = append(result, d)
+		devices = append(devices[:i], devices[i+1:]...)
+
+		switch device := spec.Device.(type) {
+		case *types.VirtualDisk:
+			if spec.FileOperation == types.VirtualDeviceConfigSpecFileOperationDestroy {
+				var file string
+
+				switch b := device.Backing.(type) {
+				case types.BaseVirtualDeviceFileBackingInfo:
+					file = b.GetVirtualDeviceFileBackingInfo().FileName
+				}
+
+				if file != "" {
+					dc := Map.getEntityDatacenter(Map.Get(*vm.Parent).(mo.Entity))
+					dm := Map.VirtualDiskManager()
+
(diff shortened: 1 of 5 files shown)


dougm:
@bradfitz vm.clone was ignoring the -net flag if the source VM doesn't have a NIC.  Is that the case in your environment?

simulator/virtual_machine.go
```diff
@@ -513,19 +513,54 @@ func (vm *VirtualMachine) configureDevice(devices object.VirtualDeviceList, spec
...
 	for i, d := range devices {
-		if d.GetVirtualDevice().Key == device.GetVirtualDevice().Key {
-			result = append(result, devices[i+1:]...)
-			break
+		if d.GetVirtualDevice().Key != key {
+			continue
 		}
 
-		result = append(result, d)
+		devices = append(devices[:i], devices[i+1:]...)
```
anfernee: not sure if you can do this inside an iteration, while you change devices while iterating it. it looks suspicious.

dougm: There isn't another iteration after this though, we `break` out of the loop after the `switch` statement.

bradfitz:
@dougm, the VM I'm cloning from is powered off, but it had a working network before I powered it off.

dougm:
@bradfitz ok, if you could provide the output of `govc vm.info -dump $source_vm_name`, I might be able to use that to reproduce.

simulator/virtual_machine.go
anfernee: ok, i see, i missed it.

bradfitz:
@dougm, sorry for the delay. See:

https://gist.github.com/bradfitz/0509583acfaafc2c0905bf503286d9cb

I'm pretty sure that's the right one. If not, it might've been this template:

https://gist.github.com/bradfitz/7c538c86958064c0e98562760ec5a4fd

dougm:
@bradfitz sorry, I got so distracted with the other issue and adding the vcsim support, I must not have actually tried using `-net.address` with vm.clone.  We were only updating the `Backing` field, which does not contain the MacAddress.  Updated the PR to also set the MacAddress and a new test, works as expected against vcsim.  But the change still isn't applied against a real vCenter.

I'll look into it more, if it just isn't supported by the CloneVm method for some reason, we can just remove the net.address flag from vm.clone and stick with using vm.change.

dougm:
@bradfitz ok, `-net.address` does work fine with vm.clone after the last change, but it requires the `-net` flag:
```console
% govc vm.clone -net "VM Network" -net.address 00:00:0f:a7:a0:f1 -vm foo bar                                                                                                                                                                                                              
[09-02-18 18:08:08] Cloning /dc1/vm/foo to bar...OK
% govc device.info -vm bar ethernet-* | grep MAC                                                                                                                                                                                                                                          
  MAC Address:      00:00:0f:a7:a0:f1
```
I'll leave it that way for now.

PR #1031, opened 2018-02-08
Add support for Datacenter.PowerOnMultiVM

dougm:
- Add PowerOnVM method to object.Datacenter

- Add PowerOnMultiVMTask method to simulator.Datacenter

- Add '-M' flag to govc vm.power command

- Make it easier to call vcsim_env multiple times in a single test

- Add vcsim support

### DIFF
--- a/object/datacenter.go
+++ b/object/datacenter.go
@@ -88,3 +88,42 @@ func (d Datacenter) Destroy(ctx context.Context) (*Task, error) {
 
 	return NewTask(d.c, res.Returnval), nil
 }
+
+// PowerOnVM powers on multiple virtual machines with a single vCenter call.
+// If called against ESX, serially powers on the list of VMs and the returned *Task will always be nil.
+func (d Datacenter) PowerOnVM(ctx context.Context, vm []types.ManagedObjectReference, option ...types.BaseOptionValue) (*Task, error) {
+	if d.Client().IsVC() {
+		req := types.PowerOnMultiVM_Task{
+			This:   d.Reference(),
+			Vm:     vm,
+			Option: option,
+		}
+
+		res, err := methods.PowerOnMultiVM_Task(ctx, d.c, &req)
+		if err != nil {
+			return nil, err
+		}
+
+		return NewTask(d.c, res.Returnval), nil
+	}
+
+	for _, ref := range vm {
+		obj := NewVirtualMachine(d.Client(), ref)
+		task, err := obj.PowerOn(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		err = task.Wait(ctx)
+		if err != nil {
+			// Ignore any InvalidPowerState fault, as it indicates the VM is already powered on
+			if f, ok := err.(types.HasFault); ok {
+				if _, ok = f.Fault().(*types.InvalidPowerState); !ok {
+					return nil, err
+				}
+			}
+		}
+	}
+
+	return nil, nil
+}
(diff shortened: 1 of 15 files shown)


cgtexmex:
Any concern w/number of VMs to powerOn?  Provide a bit of safety from an erroneous call that might create a powerOn _storm"?  In other words should a default limit be enforced and allow the user to override?

dougm:
@cgtexmex don't want to put an arbitrary limit in, leave it to vpxd to handle it or the caller.

cgtexmex:
godspeed vpxd...godspeed...

object/datacenter.go
```diff
@@ -88,3 +88,42 @@ func (d Datacenter) Destroy(ctx context.Context) (*Task, error) {
...
+		task, err := obj.PowerOn(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		err = task.Wait(ctx)
+		if err != nil {
+			// Ignore any InvalidPowerState fault, as it indicates the VM is already powered on
+			if f, ok := err.(types.HasFault); ok {
+				if _, ok = f.Fault().(*types.InvalidPowerState); !ok {
```
hickeng: For absolute assurance we can check the `ExistingState` field in the InvalidPowerState error. I don't know what other states are possible, but I'm hesitant to assume that any error will be because it's powered on.

object/datacenter.go
```diff
@@ -88,3 +88,42 @@ func (d Datacenter) Destroy(ctx context.Context) (*Task, error) {
...
+			return nil, err
+		}
+
+		return NewTask(d.c, res.Returnval), nil
+	}
+
+	for _, ref := range vm {
+		obj := NewVirtualMachine(d.Client(), ref)
+		task, err := obj.PowerOn(ctx)
+		if err != nil {
```
hickeng: This will stop powering on VMs on the first error instead of building a set of errors for those in the input that don't succeed. That makes sense if the reason for the failure is resource limit related, but I suspect it's notably divergent from what the VC operation does.
Either way I think it would be useful to note that the first failure stops the operation in the function doc.

simulator/datacenter.go
```diff
@@ -20,15 +20,40 @@ import (
...
+}
+
+// NewDatacenter creates a Datacenter and its child folders.
+func NewDatacenter(f *Folder) *Datacenter {
+	dc := &Datacenter{
+		isESX: f.Self == esx.RootFolder.Self,
+	}
+
+	if dc.isESX {
+		dc.Datacenter = esx.Datacenter
```
hickeng: should there be a return here? I'm not sure if the `putChild` and `createFolders` are idempotent for ESX.
Doesn't look like `createFolders` is - looks like it'll overwrite an existing folder the in map.

object/datacenter.go
dougm: It only returns here if it failed to create the task.

simulator/datacenter.go
dougm: We need the same logic for both.  Using esx.Datacenter here just ends up using the ESX moref values such as `ha-datacenter`, instead of the VPX values.

object/datacenter.go
dougm: PRs welcome to improve the ESX mode.  I debated including it at all, not too late to remove it either :)

simulator/datacenter.go
hickeng: It looks like createFolders will create new folders on ESX even if the dc folders already exist - will that not remove any inventory held by those folders?

dougm: The `NewDatacenter` method is only called once internally when vcsim is in ESX mode.  And the `Folder.CreateDatacenter` method is disabled in ESX mode.

PR #1032, opened 2018-02-08
Fix build on Windows

### LINKED ISSUE #1028: Windows simulator build is broken.
There's currently a compile error since _unix.go isn't a build flag that gets skipped on Windows:

```
$ GOOS=windows go build github.com/vmware/govmomi/simulator
# github.com/vmware/govmomi/simulator
simulator/os_unix.go:23:11: undefined: syscall.Statfs_t
simulator/os_unix.go:25:9: undefined: syscall.Statfs
simulator/os_windows.go:21:6: (*Datastore).stat redeclared in this block
	previous declaration at simulator/os_unix.go:21:6
```

anfernee:
Fixes #1028

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,4 @@ before_install:
 
 script:
   - make check test
+  - GOOS=windows make install
--- a/simulator/os_unix.go
+++ b/simulator/os_unix.go
@@ -1,3 +1,5 @@
+//+build !windows
+
 /*
 Copyright (c) 2017 VMware, Inc. All Rights Reserved.
 


.travis.yml
```diff
@@ -9,4 +9,5 @@ before_install:
   - make vendor
 
 script:
-  - make check test
+  - make check test install
```
dougm: No need to add install here, test already does that as the binaries are needed to run the bats tests.

.travis.yml
```diff
@@ -9,4 +9,5 @@ before_install:
...
+  - GOOS=windows make install
```
dougm: :+1:

PR #1034, opened 2018-02-13
object: Return correct helper object for OpaqueNetwork

vancluever:
Currently, `object.NewReference` returns a generic `Network` object when
called for a `OpaqueNetwork` MO. This has the side effect of breaking
`EthernetCardBackingInfo`, which will no longer contain the information
relevant to NSX that is required to correctly attach to the
logical switch that the backing is trying to connect to.

This corrects this so that such references are correctly returned
`OpaqueNetwork` objects, from which the correct relevant backing formation
can be extracted.


dougm:
thanks @vancluever !

PR #1041, opened 2018-02-20
Property collector verifies the response propset

### LINKED ISSUE #1035: vm.create  segmentation violation
Hello, having some issues creating VMs in our vCenter (6.5.0)

govc version 
0.16.0

running a basic vm create, all my GOVC params are sourced,

```
[mike@mrxmaster2:vmreaper]$ govc vm.create -net.adapter=vmxnet3 -c=2 -disk=60G -g otherGuest64 -m 2048 -on=false njofisheye01

[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9da71d]

goroutine 1 [running]:
github.com/vmware/govmomi/object.DistributedVirtualPortgroup.EthernetCardBackingInfo(0xc42031a820, 0x13, 0xc420168840, 0xc42031a340, 0x1b, 0xc420312cf0, 0x10, 0x18ad580, 0xc4200140d8, 0x0, ...)
        /home/vmware/release/src/github.com/vmware/govmomi/object/distributed_virtual_portgroup.go:47 +0x24d
github.com/vmware/govmomi/object.(*DistributedVirtualPortgroup).EthernetCardBackingInfo(0xc42037f680, 0x18ad580, 0xc4200140d8, 0x0, 0x0, 0x2a8, 0x10256c0)
        <autogenerated>:1 +0x7e
github.com/vmware/govmomi/govc/flags.(*NetworkFlag).Device(0xc4201b1ab0, 0x0, 0xc42033d9a8, 0xc420485610, 0x412f2d)
        /home/vmware/release/src/github.com/vmware/govmomi/govc/flags/network.go:114 +0x7f
github.com/vmware/govmomi/govc/vm.(*create).addNetwork(0xc4201bab40, 0xc420448320, 0x2, 0x2, 0xc420448320, 0x2, 0x2, 0x0, 0x0)
        /home/vmware/release/src/github.com/vmware/govmomi/govc/vm/create.go:435 +0x33
github.com/vmware/govmomi/govc/vm.(*create).createVM(0xc4201bab40, 0x18ad580, 0xc4200140d0, 0x0, 0x0, 0xc420057d30)
        /home/vmware/release/src/github.com/vmware/govmomi/govc/vm/create.go:306 +0x15c
github.com/vmware/govmomi/govc/vm.(*create).Run(0xc4201bab40, 0x18ad580, 0xc4200140d0, 0xc420064540, 0x0, 0x0)
        /home/vmware/release/src/github.com/vmware/govmomi/govc/vm/create.go:262 +0x2cc
github.com/vmware/govmomi/govc/cli.Run(0xc4200100c0, 0xa, 0xa, 0xc88700)
        /home/vmware/release/src/github.com/vmware/govmomi/govc/cli/command.go:156 +0x522
main.main()
        /home/vmware/release/src/github.com/vmware/govmomi/govc/main.go:82 +0x63
```

whats weird is that I used to be able to create VMs before, I tried this both w govc 0.12 and 0.16 and same result. Wondering ifs something to do with our vCenter version, we recently upgraded from 5.5 to 6.5, but it doesnt look like its even communicating w vCenter at the stage where it breaks down.

anfernee:
When property collector retrieves a certian propset, it makes sure the response contains all the input propset. This will report error when it failed to retrieve property set on objects from remote server. 

Fix #1035

### DIFF
--- a/property/collector.go
+++ b/property/collector.go
@@ -19,6 +19,7 @@ package property
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/methods"
@@ -35,13 +36,19 @@ import (
 type Collector struct {
 	roundTripper soap.RoundTripper
 	reference    types.ManagedObjectReference
+
+	// Verify all requested PropSet appears in the response PropSet
+	verifyAllPropSetNotNil bool
 }
+
+type responseVerifier func(objs []types.ManagedObjectReference, ps []string, res *types.RetrievePropertiesResponse) error
 
 // DefaultCollector returns the session's default property collector.
 func DefaultCollector(c *vim25.Client) *Collector {
 	p := Collector{
-		roundTripper: c,
-		reference:    c.ServiceContent.PropertyCollector,
+		roundTripper:           c,
+		reference:              c.ServiceContent.PropertyCollector,
+		verifyAllPropSetNotNil: true,
 	}
 
 	return &p
@@ -173,6 +180,12 @@ func (p *Collector) Retrieve(ctx context.Context, objs []types.ManagedObjectRefe
 		return nil
 	}
 
+	if p.verifyAllPropSetNotNil {
+		if err := verifyAllPropSetNotNil(ps, res); err != nil {
+			return err
+		}
+	}
+
 	return mo.LoadRetrievePropertiesResponse(res, dst)
 }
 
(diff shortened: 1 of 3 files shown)


dougm:
Fields that can't be collected should be captured by the `MissingSet` field:
http://pubs.vmware.com/vsphere-6-5/index.jsp#com.vmware.wssdk.apiref.doc/vmodl.query.PropertyCollector.ObjectContent.html

We check that here too: https://github.com/vmware/govmomi/blob/3c1bff8adbaf7408746a1f3b28ace708639b9e3e/vim25/mo/retrieve.go#L51-L57

Not sure why we don't hit that path in #1035 though.  Probably just checking for nil to avoid the panic in `DistributedVirtualPortgroup.EthernetCardBackingInfo` would be good enough for now.

anfernee:
@dougm in theory `MissingSet` should have captured uncollected fields. however, I tried against both VC/ESX, I found `MissingSet` is not always set for missing fields. We might still need this. I will add comments.

anfernee:
Reading the document again. I think what it means is `MissingSet` contains the fields that collector failed to collect, and it has a failure reason attached to it. For the cases like collected value is `nil`, it doesn't appear in either `MissingSet` or `PropSet`.

dougm:
`nil` is a valid property value, so it won't show up in `MissingSet`, nor should a `nil` value result in an error.
The `MissingSet` is populated if a property value cannot be collected due to permissions error or the object has been deleted, etc.

anfernee:
@dougm exactly. The problem is if the requested `PropSet` is something like `config.abc` and it's a `nil`. It will be missing in the response `PropSet`. It's not just left `config.abc` as unset, it also left `config` as unset. Then later when we do `vm.config.abc` it just went panic. 

I think we should fill the response `PropSet` explicitly say `config.abc` is `nil`, so that the config will be loaded in the original object. WDYT?

dougm:
That isn't the case in #1035 , since the `DistributedVirtualPortgroup.Config` isn't a pointer type.  But the `DistributedVirtualSwitch` is a pointer that `EthernetCardBackingInfo` doesn't check for `nil`.  Even if `Config` were a pointer here, I think just having the caller check both fields for `nil` is fine.  We do so elsewhere, for example with VMs 'Guest' is nil unless tools is running: https://github.com/vmware/govmomi/blob/master/govc/vm/info.go#L150

anfernee:
let me close this and look for a better solution for this issue.

PR #1044, opened 2018-02-21
Issues 1039 979

anfernee:
- Implement Destroy task for HostSystem
-  nit PortKeys in DistributedVirtualPortgroup

### DIFF
--- a/simulator/host_system.go
+++ b/simulator/host_system.go
@@ -173,6 +173,25 @@ func CreateStandaloneHost(f *Folder, spec types.HostConnectSpec) (*HostSystem, t
 	return host, nil
 }
 
+func (h *HostSystem) DestroyTask(req *types.Destroy_Task) soap.HasFault {
+	task := CreateTask(h, "destroy", func(t *Task) (types.AnyType, types.BaseMethodFault) {
+		if len(h.Vm) > 0 {
+			return nil, &types.ResourceInUse{}
+		}
+
+		f := Map.getEntityParent(h, "Folder").(*Folder)
+		f.removeChild(h.Reference())
+
+		return nil, nil
+	})
+
+	return &methods.Destroy_TaskBody{
+		Res: &types.Destroy_TaskResponse{
+			Returnval: task.Run(),
+		},
+	}
+}
+
 func (h *HostSystem) EnterMaintenanceModeTask(spec *types.EnterMaintenanceMode_Task) soap.HasFault {
 	task := CreateTask(h, "enterMaintenanceMode", func(t *Task) (types.AnyType, types.BaseMethodFault) {
 		h.Runtime.InMaintenanceMode = true
--- a/simulator/host_system_test.go
+++ b/simulator/host_system_test.go
@@ -130,3 +130,79 @@ func TestNewHostSystem(t *testing.T) {
 		t.Fatal("expected hs.Summary.Runtime == &hs.Runtime; got !=")
 	}
 }
+
+func TestDestroyHostSystem(t *testing.T) {
+	ctx := context.Background()
+	m := VPX()
+
+	defer m.Remove()
+
+	err := m.Create()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	s := m.Service.NewServer()
+	defer s.Close()
+
+	c := m.Service.client
+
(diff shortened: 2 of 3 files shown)


object/host_system.go
```diff
@@ -151,3 +151,16 @@ func (h HostSystem) ExitMaintenanceMode(ctx context.Context, timeout int32) (*Ta
 
 	return NewTask(h.c, res.Returnval), nil
 }
+
+func (h HostSystem) Destroy(ctx context.Context) (*Task, error) {
```
dougm: No need to add this method, it is already defined in the embedded `Common` struct.

simulator/host_system.go
```diff
@@ -173,6 +173,20 @@ func CreateStandaloneHost(f *Folder, spec types.HostConnectSpec) (*HostSystem, t
 	return host, nil
 }
 
+func (h *HostSystem) DestroyTask(req *types.Destroy_Task) soap.HasFault {
+	task := CreateTask(h, "destroy", func(t *Task) (types.AnyType, types.BaseMethodFault) {
+		f := Map.getEntityParent(h, "Folder").(*Folder)
```
dougm: We probably need to check that the host doesn't have any VMs running on it, may need to be in disconnected state, etc.  ok to leave as-is for now though, maybe just add a TODO.

simulator/host_system_test.go
```diff
@@ -130,3 +130,38 @@ func TestNewHostSystem(t *testing.T) {
 		t.Fatal("expected hs.Summary.Runtime == &hs.Runtime; got !=")
 	}
 }
+
+func TestDestroyHostSystem(t *testing.T) {
+	ctx := context.Background()
+	m := ESX()
```
dougm: Don't think real ESX allows Destroy on the 1 HostSystem.  But this is also fine to leave as-is for now.

simulator/host_system.go
anfernee: @dougm I see. Will make the change.

simulator/host_system_test.go
anfernee: got it. gonna check why it passed..

simulator/host_system.go
```diff
@@ -173,6 +173,26 @@ func CreateStandaloneHost(f *Folder, spec types.HostConnectSpec) (*HostSystem, t
 	return host, nil
 }
 
+func (h *HostSystem) DestroyTask(req *types.Destroy_Task) soap.HasFault {
+	task := CreateTask(h, "destroy", func(t *Task) (types.AnyType, types.BaseMethodFault) {
+		for _, vmref := range h.Vm {
+			vm := Map.Get(vmref).(*VirtualMachine)
+			vm.Runtime.ConnectionState = types.VirtualMachineConnectionStateDisconnected
```
dougm: Don't want to change ConnectionState here "When the API is provided directly by a server product, such as ESX Server, then the disconnected state is not possible", see: http://pubs.vmware.com/vsphere-6-5/index.jsp#com.vmware.wssdk.apiref.doc/vim.VirtualMachine.ConnectionState.html

Also note that with the new locking mechanism, the HostSystem (`h`) will be locked here, but VirtualMachine will not.  So you'd want to use `Map.WithLock(vm, func() {...})`

I think the task should return an error such as `ResourceInUse` if `len(h.Vm) != 0`, like what happens against real VC.
A host can't be destroyed unless VMs have been migrated elsewhere (or VMs have been deleted).  There's probably other checks that real VC makes, but again I think it's fine to leave those checks out for now.

anfernee: AH. I see. Thanks for pointing out the correct behavior.

simulator/host_system.go
```diff
@@ -173,6 +173,31 @@ func CreateStandaloneHost(f *Folder, spec types.HostConnectSpec) (*HostSystem, t
 	return host, nil
 }
 
+func (h *HostSystem) DestroyTask(req *types.Destroy_Task) soap.HasFault {
+	if len(h.Vm) > 0 {
```
dougm: Don't do this here, real VC doesn't fail the creation of tasks, only the running of tasks (which you have right below).

anfernee:
cc @dougm

PR #1047, opened 2018-02-23
Print Table of Contents in usage.md for govc

alexbozhenko:
Found good example of toc using markdown here:
https://stackoverflow.com/a/448929/1572363

### DIFF
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -30,6 +30,215 @@ but appear via `govc $cmd -h`:
   -vm.uuid=                 Find VM by UUID
 ```
 
+<details><summary>Contents</summary>
+
+ - [about](#about)
+ - [about.cert](#aboutcert)
+ - [cluster.add](#clusteradd)
+ - [cluster.change](#clusterchange)
+ - [cluster.create](#clustercreate)
+ - [cluster.group.change](#clustergroupchange)
+ - [cluster.group.create](#clustergroupcreate)
+ - [cluster.group.ls](#clustergroupls)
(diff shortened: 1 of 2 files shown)


dougm:
@alexbozhenko thanks, I had done similar originally but felt a toc with every command was too much.  What if we had a toc of top-level sections such as "cluster", "datacenter", "vm", etc, that jump to the first command in each group?  Maybe each section can have their own toc too, with the sub commands.
What do you think?

alexbozhenko:
Hi @dougm. Well, for use case scenario like: "I never used any vmware apis, clis, only used web ui, so I have no idea what command exists", having full TOC is very useful. It won't blow up the page, since it just adds several more pages, and imho, having per-section TOC is less convenient.
May be hide it under '[details](https://gist.github.com/joyrexus/16041f2426450e73f5df9391f7f7ae5f)' tag, if you think it is too big

clouderati:
I feel that a full ToC is quite busy / long.

I like @dougm's suggestion. Another similar suggestion: put top-level commands (e.g., datastore, clusters, host, vm, etc.) in the ToC, then put sub-commands (e.g., vm.power, datastore.info) hidden under a details tag.

Thoughts?

dougm:
Oh, the collapsed details tag would be nice, had no idea that could be done in markdown.

alexbozhenko:
Updated the commit. 
How about this? All toc is collapsed by default.

dougm:
@alexbozhenko looks great, thanks!

PR #1057, opened 2018-03-06
In progress.Reader emit final report on EOF.

wupeka:
In go1.10 archive/tar Reader is emitting n, io.EOF with n > 0 (which is valid), this PR makes progress.Reader emit a Report in this case so that we get to 100.0% always.


dougm:
Thanks @wupeka !

PR #1067, opened 2018-03-16
Add goreleaser to automate release process

frapposelli:
This sets up `goreleaser` to cut a new `govc` release and automate the publish process through Travis.

It requires `DOCKER_USERNAME`, `DOCKER_PASSWORD` and `GITHUB_TOKEN` to be available on the Travis job for govmomi in order to make deploy succeed (will set up once it's merged).

### DIFF
--- a/govc/version/command.go
+++ b/govc/version/command.go
@@ -20,13 +20,10 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"log"
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
 )
-
-var gitVersion string
 
 type version struct {
 	*flags.EmptyFlag
@@ -35,11 +32,6 @@ type version struct {
 }
 
 func init() {
-	// Check that git tag in the release builds match the hardcoded version
-	if gitVersion != "" && gitVersion[1:] != flags.Version {
-		log.Panicf("version mismatch: git=%s vs govc=%s", gitVersion[1:], flags.Version)
-	}
-
 	cli.Register("version", &version{})
 }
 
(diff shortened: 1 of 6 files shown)


govc/version/command.go
```diff
@@ -36,8 +36,8 @@ type version struct {
 
 func init() {
 	// Check that git tag in the release builds match the hardcoded version
-	if gitVersion != "" && gitVersion[1:] != flags.Version {
-		log.Panicf("version mismatch: git=%s vs govc=%s", gitVersion[1:], flags.Version)
+	if gitVersion != "" && gitVersion != flags.Version {
```
dougm: `flags.Version` doesn't have the `v` prefix, but the git tag does (existing ones do at least).  We could also just remove this check.  The check was added when we added the hard-coded `flags.Version`.  I'd be ok removing `flags.Version`, only added it since `go get` does not inject gitVersion, the hardcoded version at least gave a hint what the `go get` folks were using.

frapposelli:
kept `flags.Version` hardcoded to reflect current dev version, `flags.GitVersion` is set at release time, amended `ParseVersion()` to deal with dashes.

govc/version/command.go
```diff
@@ -48,8 +40,17 @@ func (cmd *version) Register(ctx context.Context, f *flag.FlagSet) {
...
+
+	if len(flags.GitVersion) > 0 {
+		ver = flags.GitVersion
+	} else {
+		ver = flags.Version
+	}
+
 	if cmd.require != "" {
-		v, err := flags.ParseVersion(flags.Version)
+
```
dougm: remove this empty line?

govc/version/command.go
```diff
@@ -48,8 +40,17 @@ func (cmd *version) Register(ctx context.Context, f *flag.FlagSet) {
 }
 
 func (cmd *version) Run(ctx context.Context, f *flag.FlagSet) error {
+	var ver string
```
dougm: Cosmetic but maybe instead:
```go
ver := flags.GitVersion
if ver == "" {
    ver = flags.Version
}
```

frapposelli:
:sheep: :it:

PR #1072, opened 2018-03-20
Add -firmware parameter to 'govc vm.create'

### LINKED ISSUE #1071: When creating a new VM with govc it is not possible to specify "EFI" instead of "BIOS" boot
In the VMWare GUI there is the option "Boot with EFI instead of BIOS".

I was unable to achieve the same thing using **govc**. Neither with `govc vm.create` nor with `govc device.boot` .

Did I miss the option? Otherwise I think it should really be implemented. If you give me some pointer on where/how to do it I will even try myself.

omarkohl:
Without this option it is not possible to specify that a VM should use
EFI instead of BIOS as firmware type.

The default behaviour (using BIOS) is unchanged.

fixes #1071

### DIFF
--- a/govc/vm/create.go
+++ b/govc/vm/create.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"strings"
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
@@ -50,6 +51,7 @@ type create struct {
 	force      bool
 	controller string
 	annotation string
+	firmware   string
 
 	iso              string
 	isoDatastoreFlag *flags.DatastoreFlag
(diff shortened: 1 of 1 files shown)


omarkohl:
Can someone confirm that the general approach makes sense? I'll then have a look at the failing tests.

vim25/types/types.go
```diff
@@ -48377,7 +48377,7 @@ type VirtualMachineConfigInfo struct {
 	VAppConfig                   BaseVmConfigInfo                           `xml:"vAppConfig,omitempty,typeattr"`
 	VAssertsEnabled              *bool                                      `xml:"vAssertsEnabled"`
 	ChangeTrackingEnabled        *bool                                      `xml:"changeTrackingEnabled"`
-	Firmware                     string                                     `xml:"firmware,omitempty"`
+	Firmware                     GuestOsDescriptorFirmwareType              `xml:"firmware,omitempty"`
```
dougm: The code in vim25/types is generated, you don't want to modify them manually.  The type also needs to be a `string` in this case, must be same as the xsi type.

govc/vm/create.go
```diff
@@ -109,6 +110,7 @@ func (cmd *create) Register(ctx context.Context, f *flag.FlagSet) {
 	f.BoolVar(&cmd.force, "force", false, "Create VM if vmx already exists")
 	f.StringVar(&cmd.controller, "disk.controller", "scsi", "Disk controller type")
 	f.StringVar(&cmd.annotation, "annotation", "", "VM description")
+	f.BoolVar(&cmd.firmwareEfi, "firmware-efi", false, "Use firmware type EFI (otherwise BIOS)")
```
dougm: I'd prefer to use the same type as the field as we do elsewhere for enums.  Even though there's only 2 options in this case, a bool is tempting to use.
```go
f.StringVar(&cmd.firmware, "firmware", "", "Firmware type [bios|efi]")
```
Some existing enum flag examples:
```go
govc/datastore/create.go:132:	f.Var(&cmd.Type, "type", fmt.Sprintf("Datastore type (%s)", strings.Join(allTypes, "|")))
govc/device/scsi/add.go:53:		fmt.Sprintf("SCSI controller type (%s)", strings.Join(ctypes, "|")))
govc/device/usb/add.go:48:		fmt.Sprintf("USB controller type (%s)", strings.Join(ctypes, "|")))
govc/dvs/portgroup/spec.go:41:		fmt.Sprintf("Portgroup type (%s)", strings.Join(ptypes, "|")))
```

PR #1076, opened 2018-03-22
govc: add json support to find command

### LINKED ISSUE #826: json output does not work with "govc find"
It seems that the json output is not implemented for the "find" function.
The help command says that the json output option is available:

./govc find --help
Usage: ./govc find [OPTIONS] [ROOT] [KEY VAL]...
[...]
Options:
  -cert=                          Certificate [GOVC_CERTIFICATE]
[...]
  -json=false                     Enable JSON output

But it doesn't work.

dougm:
Fixes #826

### DIFF
--- a/govc/object/find.go
+++ b/govc/object/find.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/vmware/govmomi/govc/cli"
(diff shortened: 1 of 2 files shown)


govc/object/find.go
```diff
@@ -172,6 +173,19 @@ func (cmd *find) rootMatch(ctx context.Context, root object.Reference, client *v
 	return content != nil
 }
 
+type findResult []string
+
+func (r findResult) Write(w io.Writer) error {
+	for i := range r {
+		fmt.Println(r[i])
```
anfernee: write it to `w`?

dougm: yes, done.

PR #1088, opened 2018-04-05
Refactor LoginExtensionByCertificate tunnel usage

dougm:
Originally, a call to soap.Client.SetCertificate configured the client's proxy to use the SDK for every requests.
However, the SDK tunnel is only required for the LoginExtensionByCertificate call itself.  After we have a session,
the normal path can be used without the SDK tunnel.  Use of the SDK tunnel is now limited to within the SessionManager
LoginExtensionByCertificate wrapper.

- Add govc session.login -extension flag

- Add vcsim LoginExtensionByCertificate implementation

- Add govc extension.setcert -set flag (handy for generating test certs)

- Remove govmomi.Client LoginExtensionByCertificate related methods

### DIFF
--- a/govc/session/login.go
+++ b/govc/session/login.go
@@ -44,6 +44,7 @@ type login struct {
 	cookie string
 	token  string
 	actas  string
+	ext    string
 }
 
 func init() {
@@ -63,6 +64,7 @@ func (cmd *login) Register(ctx context.Context, f *flag.FlagSet) {
 	f.StringVar(&cmd.cookie, "cookie", "", "Set HTTP cookie for an existing session")
 	f.StringVar(&cmd.token, "token", "", "Use SAML token for login")
 	f.StringVar(&cmd.actas, "actas", "", "Issue SAML token with this token's identity")
+	f.StringVar(&cmd.ext, "extension", "", "Extension name")
 }
 
 func (cmd *login) Process(ctx context.Context) error {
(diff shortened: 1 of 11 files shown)


dougm:
Note that PR #1064 broke `LoginExtensionByCertificate()`.  The vSAN team depends on this method, so I wanted to simplify things and make it possible to test the login method against vcsim.

@anfernee we also could revisit your vcsim support for extensions with this in place.

govc/session/login.go
```diff
@@ -212,6 +220,8 @@ func (cmd *login) Run(ctx context.Context, f *flag.FlagSet) error {
 		cmd.Login = cmd.setCookie
 	case cmd.token != "":
 		cmd.Login = cmd.loginByToken
+	case cmd.ext != "":
+		cmd.Login = cmd.loginByExtension
```
lichen782: would you mind show me where is the cmd(*login).Login is called?

session/manager.go
```diff
@@ -89,18 +90,37 @@ func (sm *Manager) Login(ctx context.Context, u *url.Userinfo) error {
...
+	u := c.URL()
+	if u.Hostname() != "sdkTunnel" {
+		sc := c.Tunnel()
+		c = &vim25.Client{
+			Client:         sc,
+			RoundTripper:   sc,
+			ServiceContent: c.ServiceContent,
+		}
+		// When http.Transport.Proxy is used, our thumbprint checker is bypassed, resulting in:
+		// "Post https://sdkTunnel:8089/sdk: x509: certificate is valid for $vcenter_hostname, not sdkTunnel"
```
lichen782: are we able to post to this hostname? is it from localhost on VC?

vim25/soap/client.go
```diff
@@ -181,6 +180,8 @@ func (c *Client) NewServiceClient(path string, namespace string) *Client {
 		}
 	}
 
+	client.u.RawQuery = vc.RawQuery
```
lichen782: what's this used for?

PR #1092, opened 2018-04-09
 Add ssoadmin client and commands

### LINKED ISSUE #807: Need to create a vCenter user using govmomi or GOVC
I see `govc host.account.create` allows creating new ESXi user on the host.

I could not find anything to create new user on vCenter server. Is it possible to create a new vCenter user using govmomi or govc?

@SandeepPissay @tusharnt @dougm

dougm:
- Add ssoadmin.Client for managing users

- Add govc sso.user id, ls, create, update, rm commands

- Use ssoadmin in sts tests to create solution user

Fixes #807

### DIFF
--- a/govc/sso/user/update.go
+++ b/govc/sso/user/update.go
@@ -1,0 +1,113 @@
+/*
+Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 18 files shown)


govc/sso/user/update.go
```diff
@@ -0,0 +1,120 @@
...
+	if f.NArg() != 1 {
+		return flag.ErrHelp
+	}
+	id := f.Arg(0)
+
+	updated := false
+	merge := func(src *string, current string) {
+		if *src == "" {
+			*src = current
+		} else {
```
anfernee: if `src == current`, updated still false, no?

dougm: I added a comment to the merge func.  And removed the `updated` flag, no need to optimize in this case.

PR #1109, opened 2018-04-28
Add STS support for token renewal

### LINKED ISSUE #1093: Add STS support for token renewal
The STS client supports issuing a new token, but not renewing an existing token.  Renew support is required to extend the token lifetime beyond the initial 5 minute max granted by the server when issuing a new token.

dougm:
Mostly a refactor as "Issue" and "Renew" requests are similar, but also adds new types and their canonicalizers.

Removed time.Time wrapper, using string type instead.
This avoids an issue where xml.Marshal was dropping trailing zeros from the timestamp,
causing an invalid body digest.

Fixes #1093

### DIFF
--- a/govc/session/login.go
+++ b/govc/session/login.go
@@ -45,7 +45,6 @@ type login struct {
 	life   time.Duration
 	cookie string
 	token  string
-	actas  string
 	ext    string
 }
 
@@ -65,8 +64,7 @@ func (cmd *login) Register(ctx context.Context, f *flag.FlagSet) {
 	f.BoolVar(&cmd.long, "l", false, "Output session cookie")
 	f.StringVar(&cmd.ticket, "ticket", "", "Use clone ticket for login")
 	f.StringVar(&cmd.cookie, "cookie", "", "Set HTTP cookie for an existing session")
-	f.StringVar(&cmd.token, "token", "", "Use SAML token for login")
-	f.StringVar(&cmd.actas, "actas", "", "Issue SAML token with this token's identity")
+	f.StringVar(&cmd.token, "token", "", "Use SAML token for login or as issue identity")
 	f.StringVar(&cmd.ext, "extension", "", "Extension name")
 }
 
@@ -97,7 +95,7 @@ Examples:
   govc session.login -u host -extension com.vmware.vsan.health -cert rui.crt -key rui.key
   token=$(govc session.login -u host -cert user.crt -key user.key -issue) # HoK token
   bearer=$(govc session.login -u user:pass@host -issue) # Bearer token
-  token=$(govc session.login -u host -cert user.crt -key user.key -issue -actas "$bearer")
+  token=$(govc session.login -u host -cert user.crt -key user.key -issue -token "$bearer")
   govc session.login -u host -cert user.crt -key user.key -token "$token"`
 }
 
@@ -149,14 +147,30 @@ func (cmd *login) issueToken(ctx context.Context, vc *vim25.Client) (string, err
 	req := sts.TokenRequest{
 		Certificate: c.Certificate(),
 		Userinfo:    cmd.Userinfo(),
-		Delegatable: true,
-		ActAs:       cmd.actas,
+		Renewable:   true,
+		Token:       cmd.token,
 		Lifetime:    cmd.life,
+	}
+
+	if req.Certificate == nil {
+		req.Delegatable = true // Bearer token request
 	}
 
 	s, err := c.Issue(ctx, req)
 	if err != nil {
 		return "", err
+	}
+
+	if req.Token != "" {
+		duration := s.Lifetime.Expires.Sub(s.Lifetime.Created)
+		if duration < req.Lifetime {
+			// The granted lifetime is that of the bearer token, which is 5min max.
+			// Extend the lifetime via Renew.
+			req.Token = s.Token
+			if s, err = c.Renew(ctx, req); err != nil {
+				return "", err
+			}
+		}
 	}
 
 	return s.Token, nil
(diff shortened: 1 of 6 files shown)


govc/session/login.go
```diff
@@ -97,7 +95,7 @@ Examples:
   govc session.login -u host -extension com.vmware.vsan.health -cert rui.crt -key rui.key
   token=$(govc session.login -u host -cert user.crt -key user.key -issue) # HoK token
   bearer=$(govc session.login -u user:pass@host -issue) # Bearer token
-  token=$(govc session.login -u host -cert user.crt -key user.key -issue -actas "$bearer")
+  token=$(govc session.login -u host -cert user.crt -key user.key -issue -token "$bearer")
```
akutz: Hi @dougm,

I like this change. I think the flag `token` is much more natural and friendly.

govc/session/login.go
```diff
@@ -149,16 +147,30 @@ func (cmd *login) issueToken(ctx context.Context, vc *vim25.Client) (string, err
...
 		Userinfo:    cmd.Userinfo(),
-		Delegatable: true,
-		ActAs:       cmd.actas,
+		Renewable:   true,
+		Token:       cmd.token,
 		Lifetime:    cmd.life,
 	}
 
+	if req.Certificate == nil {
+		req.Delegatable = true // Bearer token request
```
akutz: Hi @dougm,

I like reviewing your PRs because it helps me to appreciate your coding style. I might prefer `req.Delegatable = req.Certificate == nil`, but I respect the way your choice is self-documenting and easy to follow.

sts/internal/types.go
```diff
@@ -365,31 +396,54 @@ func (c *Conditions) C14N() string {
 	if c.ProxyRestriction != nil {
 		names = append(names, &c.ProxyRestriction.XMLName)
 	}
-	if c.RenewRestrictionType != nil {
-		names = append(names, &c.RenewRestrictionType.XMLName)
+
+	for i := range c.Condition {
+		switch r := c.Condition[i].(type) {
```
akutz: Hi @dougm,

I haven't looked, but I assume there are the only two `Condition` types? No default case is needed?

sts/internal/types.go
```diff
@@ -508,16 +570,34 @@ func (r *RequestSecurityToken) toString(c14n bool) string {
...
-		fmt.Sprintf(`<KeyType>%s</KeyType>`, r.KeyType),
-		fmt.Sprintf(`<SignatureAlgorithm>%s</SignatureAlgorithm>`, r.SignatureAlgorithm),
-		fmt.Sprintf(`<UseKey Sig="%s"></UseKey>`, r.UseKey.Sig),
-		`</RequestSecurityToken>`,
 	}
 
-	return strings.Join(body, "")
+	if r.RenewTarget == nil {
+		body = append(body,
+			fmt.Sprintf(`<Renewing Allow="%t" OK="%t"></Renewing>`, r.Renewing.Allow, r.Renewing.OK),
```
akutz: Hi @dougm,

Would a `bytes.Buffer` be more performant here? I'm not sure, but since you're essentially just joining the lines together anyway, writing to a buffer directly with `fmt.Fprintf` and then getting the string may end up being more efficient?

govc/session/login.go
```diff
@@ -149,16 +147,30 @@ func (cmd *login) issueToken(ctx context.Context, vc *vim25.Client) (string, err
...
+	}
+
 	s, err := c.Issue(ctx, req)
 	if err != nil {
 		return "", err
 	}
 
+	duration := s.Lifetime.Expires.Sub(s.Lifetime.Created)
```
akutz: Hi @dougm,

For some reason my previous comment here did not get added to my review. I wondered if the following variation might fall in line more with your coding style and at the same time keep `duration` out of the scope of the function:

```go
if req.Token != "" {
 	duration := s.Lifetime.Expires.Sub(s.Lifetime.Created)
 	if duration != req.Lifetime
```

govc/session/login.go
```diff
@@ -149,16 +147,30 @@ func (cmd *login) issueToken(ctx context.Context, vc *vim25.Client) (string, err
...
+	if duration != req.Lifetime && req.Token != "" {
```
zjs: Do we need to consider the case where `duration > req.Lifetime`? I would hope that would not be possible, but in that case it doesn't seem like we'd want to renew.

govc/session/login.go
```diff
@@ -65,8 +64,7 @@ func (cmd *login) Register(ctx context.Context, f *flag.FlagSet) {
 	f.BoolVar(&cmd.long, "l", false, "Output session cookie")
 	f.StringVar(&cmd.ticket, "ticket", "", "Use clone ticket for login")
 	f.StringVar(&cmd.cookie, "cookie", "", "Set HTTP cookie for an existing session")
-	f.StringVar(&cmd.token, "token", "", "Use SAML token for login")
-	f.StringVar(&cmd.actas, "actas", "", "Issue SAML token with this token's identity")
+	f.StringVar(&cmd.token, "token", "", "Use SAML token for login or issue identity")
```
zjs: It took me a few re-reads to understand what was meant by "or issue identity". Perhaps "Use SAML token for login or as issue identity" would be slightly clearer to some users?

sts/internal/types.go
dougm: It might be, but it isn't often that clients issue a token.  I would wait on optimizing for a separate PR if we think it's worthwhile, since the other existing C14N's should get the same treatment in that case.  I'd also prefer to put time into getting go-xmldsig + go-saml to work for us, rather than optimize the code in this file that I'd love to just throw away :)

Issue #1121, opened 2018-05-09
wrong capacity and free size of datastore.info against vcsim


leslie-qiwa:
If using vcsim as backend and run `govc datastore.info -k`, the capacity and free space are wrong, as below. Seems like `simulator/os_unix.go` already calculate size by MB. 
```
Name:        LocalDS_0
  Path:      /DC0/datastore/LocalDS_0
  Type:      local
  URL:       /var/folders/p1/bx0fgbnj0ljbygmb7hr32zj00000gn/T/govcsim-DC0-LocalDS_0-186611435
  Capacity:  0.2 GB
  Free:      0.1 GB
```

In addition, if VM is created on the datastore, expectation is VM disk size should be counted as well although disk file is not created physically.

I have a working patch, please let me know if I should submit one PR for it or not. Thanks.

dougm:
@leslie-qiwa a PR to fix would be great, thanks.

leslie-qiwa:
Thanks. Will add more tests on my current patch, and submit before next week.

dougm:
Fixed by #1125

PR #1125, opened 2018-05-12
vcsim: display correct capacity and free size of datastore.info

leslie-qiwa:
https://github.com/vmware/govmomi/issues/1121

This only fix stat on linux, not windows. 

Not add test part because I don't have windows machine to test, so it may break if govmomi's CI system has windows platform.

### DIFF
--- a/simulator/os_unix.go
+++ b/simulator/os_unix.go
@@ -29,12 +29,10 @@ func (ds *Datastore) stat() error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


leslie-qiwa:
Thanks for the suggestion, and detail fix. I'll incorporate and push again!

dougm:
thanks again @leslie-qiwa

PR #1133, opened 2018-05-18
Add -sharing option to vm.disk.create and vm.disk.attach

### LINKED ISSUE #1102: sharingMultiWriter flag
Hi all,

Is that any chance that possibility to put sharingMultiWriter on disk will be added, to avoid manipulation of of .vmx files.

Thanks a lot in advance,

Dan Zegelman

dougm:
Fixes #1102

### DIFF
--- a/govc/test/vm.bats
+++ b/govc/test/vm.bats
@@ -436,17 +436,57 @@ load test_helper
 
   local name=$(new_id)
 
-  run govc vm.disk.create -vm $vm -name $name -size 1G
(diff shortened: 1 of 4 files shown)


dzegelman:
I think same flag must be introduced in vm.disk.attach command too, as multiwriter disk is created and then attached to another node.

dougm:
@dzegelman thanks for reviewing.  Indeed, I've updated to include the `-sharing` flag in `vm.disk.attach`

PR #1152, opened 2018-06-06
Add option to mark VM as template on OVF/OVA import in govc

rowanjacobs:
We have several use cases inside our org where we need to upload an OVF or OVA file using govc and immediately mark it as a template. This PR adds a `MarkAsTemplate` option in the `import.ovf/ova` spec.

### DIFF
--- a/govc/importx/options.go
+++ b/govc/importx/options.go
@@ -55,10 +55,11 @@ type Options struct {
 
 	Annotation string `json:",omitempty"`
 
-	PowerOn      bool
-	InjectOvfEnv bool
-	WaitForIP    bool
-	Name         *string
+	MarkAsTemplate bool
+	PowerOn        bool
+	InjectOvfEnv   bool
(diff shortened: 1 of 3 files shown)


erikh:
I'll try it later, but is this vsphere only or ESX as well? Sorry if my question is naive.

dougm:
@erikh right, MarkAsTemplate is vCenter only and will fail against ESX.  I think that's ok here since it is disabled by default.  Though we could add a note to the flag description and/or validate earlier that the client is connected to a vCenter instance.

dougm:
It's part of the spec you can specify with `-options`.  To see an example spec:
```console
% govc import.spec foo.ova  | jq .
{
  "Deployment": "tiny",
  ...
  "MarkAsTemplate": false,
  "PowerOn": false,
  "InjectOvfEnv": false,
  "WaitForIP": false,
  "Name": null
}

```

PR #1154, opened 2018-06-07
Optionally check root CAs for validity

### LINKED ISSUE #1153: SetRootCAs silently ignores invalid certs
SetRootCAs does not error if the root ca cert offered is invalid and therefore cannot be added to a pool:

https://github.com/vmware/govmomi/blob/7736cdc786f6d1a3a5110a717c361e65d4d00b16/vim25/soap/client.go#L203

It would be valuable to bubble up this error, probably through the error returned. That way consumers can choose to deal with it at client creation time. Currently the actual connection attempt fails at runtime, masking the actual problem.

cc @mariantalla

mariantalla:
When ValidateRootCAs is set to true, each certificate SetRootCAs trys to
add to the root CA pool is checked for validity; if any certificate is
not valid, the method returns an error.

By default, ValidateRootCAs is false, thus this is an opt-in. If consumers of the method want to receive errors, they have to explicitly set `soapClient.ValidateRootCAs = true` in their client structs. Existing consumers will not see any difference in the behaviour of the method.

The test certificate has been generated with:

```
openssl req -x509 -nodes -newkey rsa:4096 -keyout valid-cert.key.pem -out valid-cert.pem -days $(( 200 * 365 )) -subj "/C=UK/L=London/O=Example Company/OU=Example Org/CN=www.example.com"
```

Closes #1153

cc @hoegaarden


mariantalla:
@hoegaarden and I are currently checking on CLA.

mariantalla:
@dougm We haven't made any changes to the CHANGELOG as part of this PR as we don't want to interfere with the project's release versioning. Let us know how to best go about it please.

cc @hoegaarden

dougm:
Regarding the ChangeLog, you can just add a new heading `### unreleased`.  Then we just manually change to version + date when we create the release.

hoegaarden:
- We are happy to go with mandatory validation. We just thought we'd be careful and thus went for opt-in as a start. But I guess you still want to have an opt-out, right? Or would you just go full in and do the validation no matter what?
- We will add the the `unreleased` changelog then! :+1:

dougm:
@hoegaarden I think we can make it mandatory for now.  Would prefer to avoid adding a new exported field if we can.

vim25/soap/client.go
```diff
@@ -72,11 +72,29 @@ type Client struct {
 	Version   string // Vim version
 	UserAgent string
 
+	// ValidateRootCAs: if set to true, the soap client will ensure the root CA
+	// certificates are valid in SetRootCAs. If at least one of the certs is
+	// invalid, SetRootCAs will return an error.
+	// Defaults to false; which means that the client will silently ignore any
+	// cert validation errors during SetRootCAs.
+	ValidateRootCAs bool
```
dougm: ok to leave the field for tests to use, let's make it unexported `validateRootCAs`.

vim25/soap/client.go
```diff
@@ -72,11 +72,29 @@ type Client struct {
...
+
 	cookie string
 }
 
 var schemeMatch = regexp.MustCompile(`^\w+://`)
 
+type ErrInvalidCACertificate struct {
```
dougm: Let's avoid exporting this type, `errInvalidCACertificate` instead would be fine.

vim25/soap/client.go
```diff
@@ -72,11 +72,29 @@ type Client struct {
...
+	File string
+}
+
+func (e ErrInvalidCACertificate) Error() string {
+	return fmt.Sprintf(
+		"Invalid certificate '%s', cannot be used as a trusted CA certfificate",
```
dougm: go-lint prefers lower case, `invalid` instead.  And typo: `certfificate`

vim25/soap/client.go
hoegaarden: Well, when we don't export it, how can people who use the package check, if the error was because of an invalid certificate or because something else (reading the file failed ...)?

hoegaarden: As it is possible to add multiple certs in one go (`cert1:cert2:cert3`) I think it is especially important for users to have access to `ErrInvalidCACertificate.File` to check which of the certs is invalid.

hoegaarden:
@dougm PTAL

vim25/soap/client.go
dougm: I don't see where calling code would make use of the custom error type though, other than in the tests.  If an invalid CA file is provided, the generic `error` interface type with File name in the `Error` message as it is now should be all the user needs.

Just trying to avoid adding to govmomi's too many exported types unless we are certain it is needed.  We can always choose to export it later if there are use case(s) that would require the custom type.  What do you think?

hoegaarden:
@dougm Is there anything missing, anything we should address? Thanks!

dougm:
@hoegaarden yes, my earlier comment regarding the custom error type: https://github.com/vmware/govmomi/pull/1154#discussion_r194191968

hoegaarden:
Oh ... sorry, I must have missed that comment!
As you said, if we have evidence that this error type is useful somewhere, we can just export it -- let's keep it unexported for now.

Issue #1156, opened 2018-06-13
govc host.info is miscalculating the logical CPU count


kay-ge:
This line of code is incorrect:

https://github.com/vmware/govmomi/blob/f5c88d235d8c95f474a481c3bc0418ef1ba31743/govc/host/info.go#L142

The total number of logical CPUs is simply h.NumCpuThreads.  NumCpuThreads is the total number of threads across all CPUs, not per CPU package, so no multiplication is necessary.

Example incorrectly-multiplied output snippet:

Manufacturer:    Dell Inc.
 Logical CPUs:    112 CPUs @ 2195MHz
 Processor type:  Intel(R) Xeon(R) Gold 5120 CPU @ 2.20GHz
 CPU usage:       1720 MHz (0.7%)
 Memory:          195266MB
 Memory usage:    73724 MB (37.8%)

Example ESX UI screenshot showing the actual logical processor count:
![image](https://user-images.githubusercontent.com/19159458/41328809-fb734cc2-6e7e-11e8-80ba-f039e78106d9.png)

PR #1159, opened 2018-06-13
Fixed govc host.info logical CPU count

kay-ge:
This command was incorrectly multiplying the number of CPU packages by the
total number of CPU threads.  The number of CPU threads is across all CPUs,
so it can be used instead.

Example output prior to the fix:

Name:              internal
  Path:            /path/to/ESX/host
  Manufacturer:    Dell Inc.
  Logical CPUs:    112 CPUs @ 2195MHz
  Processor type:  Intel(R) Xeon(R) Gold 5120 CPU @ 2.20GHz
  CPU usage:       1654 MHz (0.7%)
  Memory:          195266MB
  Memory usage:    73725 MB (37.8%)
  Boot time:       2018-05-11 23:17:05.118699 +0000 UTC
  State:           connected

Example output with the fix applied:

Name:              internal
  Path:            /path/to/ESX/host
  Manufacturer:    Dell Inc.
  Logical CPUs:    56 CPUs @ 2195MHz
  Processor type:  Intel(R) Xeon(R) Gold 5120 CPU @ 2.20GHz
  CPU usage:       1566 MHz (1.3%)
  Memory:          195266MB
  Memory usage:    73715 MB (37.8%)
  Boot time:       2018-05-11 23:17:05.118699 +0000 UTC
  State:           connected

Resolves: #1156

### DIFF
--- a/govc/host/info.go
+++ b/govc/host/info.go
@@ -139,7 +139,7 @@ func (r *infoResult) Write(w io.Writer) error {
 		s := host.Summary
 		h := s.Hardware
 		z := s.QuickStats
-		ncpu := int32(h.NumCpuPkgs * h.NumCpuThreads)
+		ncpu := int32(h.NumCpuThreads)
 		cpuUsage := 100 * float64(z.OverallCpuUsage) / float64(ncpu*h.CpuMhz)
 		memUsage := 100 * float64(z.OverallMemoryUsage) / float64(h.MemorySize>>20)
 


embano1:
> Looks like we broke this in #1004 (cc @embano1)

@dougm no, we didn't break it. We just did not fix it correctly :) It was broken before and just using `h.NumCpuThreads` was the right approach.

Great to see this fixed.

Issue #1165, opened 2018-06-26
Testing support for govc tags commands 


jiatongw:
Client bindings for the tagging API in govc has finished. Issues can be found here #1139.  
This issue is related with adding testing support for govc tags commands.

jiatongw:
@dougm

PR #1166, opened 2018-06-26
Add testing support for govc tags commands

jiatongw:
This PR is to add testing support for govc tags commands. 

Issue can be found there #1165

### DIFF
--- a/govc/test/tags.bats
+++ b/govc/test/tags.bats
@@ -1,0 +1,102 @@
+#!/usr/bin/env bats
+
+load test_helper
+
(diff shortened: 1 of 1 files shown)


dougm:
@jiatongw please use 2-space indent instead of 4, as we have with the other *.bats and *.sh scripts.

There is much more use of 'awk' in these tests compared to what we see in the other tests.. It looks like you could use the `run` command with the `$lines` variable instead for most cases: https://github.com/sstephenson/bats#run-test-other-commands

PR #1167, opened 2018-07-02
Add UUID helper for VirtualMachine MO

frapposelli:
Easy way to fetch VM UUID.

### DIFF
--- a/object/virtual_machine.go
+++ b/object/virtual_machine.go
@@ -799,3 +799,15 @@ func (v VirtualMachine) UpgradeVM(ctx context.Context, version string) (*Task, e
 
 	return NewTask(v.c, res.Returnval), nil
 }
+
+// UUID is a helper to get the UUID of the virtual machine object.
+func (v VirtualMachine) UUID(ctx context.Context) string {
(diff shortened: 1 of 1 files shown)


object/virtual_machine.go
```diff
@@ -799,3 +799,15 @@ func (v VirtualMachine) UpgradeVM(ctx context.Context, version string) (*Task, e
 
 	return NewTask(v.c, res.Returnval), nil
 }
+
+// UUID is a helper to get the UUID of the virtual machine object.
+func (v VirtualMachine) UUID(ctx context.Context) string {
```
markpeek: Just curious, should this function return an error instead of an empty string?

dougm: Good point, if we don't return the error, let's document the behavior in the godoc section.

frapposelli: Generically speaking, a VM object always has a UUID, as the main purpose of this helper is to have an easy way to fetch it I think having a single `string` return to handle is better than handling a `(string, error)`.

But I absolutely agree that we should document the behavior in godoc 😄

PR #1170, opened 2018-07-09
vcsim: add support for PropertyCollector incremental updates

### LINKED ISSUE #922: vcsim: PropertyCollector WaitForUpdatesEx incremental change detection
Putting up a feature request to support incremental change detection for PropertyFilter in PropertyCollector (WaitForUpdatesEx) - version number should increase incrementally as well as support ObjectUpdateKind (enter, leave, modify).

Using WaitForUpdatesEx on real VCenter we only get the response when at least one of the properties in the specified filter has changed. The returned version is a number string increasing by 1 (initial being an empty string) with each response.

dougm:
Prior to this change, clients could only call WaitForUpdatesEx once,
which would return the current state.  The next invocation with Version
given would fail as unsupported.  Clients can now use WaitForUpdatesEx
to wait for multiple updates.

The basic idea is that methods should now update object fields using
the `Registry.Update` method instead of setting fields directly.
`Registry.Update` uses the same PropertyChange structure that is sent as
part of the WaitForUpdatesEx response.  `Registry.Update` will also apply
the PropertyChange(s) to the local object, removing the need to set
fields directly.

- govc commands now use Client.Flag.WithCancel to catch SIGINT,
  allowing proper cleanup of collectors

- Add vcsim ListView support to ViewManager

Fixes #922

### DIFF
--- a/govc/flags/client.go
+++ b/govc/flags/client.go
@@ -28,8 +28,10 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/vmware/govmomi/session"
@@ -624,4 +626,30 @@ func (flag *ClientFlag) Environ(extra bool) []string {
 	}
 
 	return env
+}
+
+// WithCancel calls the given function, returning when complete or canceled via SIGINT.
+func (flag *ClientFlag) WithCancel(ctx context.Context, f func(context.Context) error) error {
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGINT)
+
+	wctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	done := make(chan bool)
+	var werr error
+
+	go func() {
+		defer close(done)
+		werr = f(wctx)
+	}()
+
+	select {
+	case <-sig:
+		cancel()
+		<-done // Wait for f() to complete
+	case <-done:
+	}
+
+	return werr
 }
--- a/simulator/race_test.go
+++ b/simulator/race_test.go
@@ -24,8 +24,10 @@ import (
 	"time"
 
 	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/event"
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/view"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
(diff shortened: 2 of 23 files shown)


dougm:
Still some tests and doc to add, but the basics are functioning now.  A few examples that now work against vcsim:

``` console
% govc tasks -f

% govc events -f

% govc object.collect -wait 1h -type m / runtime.powerState
```

dougm:
cc @ausevj @hickeng @kars7e @embano1

dougm:
Added more tests and property path filters since opening the PR, ready for review now.

simulator/race_test.go
```diff
@@ -82,30 +131,19 @@ func TestRace(t *testing.T) {
 				t.Fatal(err)
```
jiatongw: Here is the tricky part, about ```t.Fatal()``` in goroutine. I have read similar code before when I learn Golang and searched. There is a group discussion: ```Whenever t.Fatal is called in a goroutine, it is swallowed and lost. ``` (reference [https://groups.google.com/forum/#!topic/vault-tool/zNlHJLxoo3w](https://groups.google.com/forum/#!topic/vault-tool/zNlHJLxoo3w), also here [https://github.com/ipfs/go-ipfs/issues/2043](https://github.com/ipfs/go-ipfs/issues/2043)). One of their guys also used ```staticcheck``` to check their code. So, for double check, I also ran it against this file. Result is the same as theirs.  So, we can discuss about that. 
Line 119, 126, 131, 152

jiatongw:
Generally, I pulled your PR and ran some commands, it works well.

simulator/race_test.go
dougm: Good catch, I've changed those to use `t.Error` instead.

govc/flags/client.go
```diff
@@ -625,3 +627,29 @@ func (flag *ClientFlag) Environ(extra bool) []string {
...
+
+	done := make(chan bool)
+	var werr error
+
+	go func() {
+		defer close(done)
+		werr = f(wctx)
+	}()
+
+	sig := make(chan os.Signal)
```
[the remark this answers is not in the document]
dougm: That reference doesn't say why though.  Go's docs are a bit better: https://golang.org/pkg/os/signal/#Notify
Your comment here also made me realize the `sig` channel needs to be created earlier, before the go routine calling `f` is started.  Changed so that happens first.

jiatongw:
@dougm  Travis CI build failed here. `simulator/property_collector_test.go` line 404, 442 also need to change to  `t.Error`

jiatongw:
Looks good to me 👍

PR #1172, opened 2018-07-10
govc: show rule details for ClusterVmHostRuleInfo rules in cluster.rule.ls

UBessle:
We've head some problems managing our cluster affinity rules with govc. govc wasn't showing any infos in cluster.rule.ls for 'ClusterVmHostRuleInfo' type rules, that match a host group to a VM group. 
With the proposed change cluster.rule.ls can now show you the type of the rule and the additional details for 'ClusterVmHostRuleInfo' type rules.

### DIFF
--- a/govc/cluster/rule/info.go
+++ b/govc/cluster/rule/info.go
@@ -1,0 +1,92 @@
+/*
+Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rule
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/vmware/govmomi/govc/cli"
+)
+
+type info struct {
+	*InfoFlag
+}
+
+func init() {
+	cli.Register("cluster.rule.info", &info{})
+}
+
+func (cmd *info) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.InfoFlag, ctx = NewInfoFlag(ctx)
+	cmd.InfoFlag.Register(ctx, f)
+}
+
+func (cmd *info) Process(ctx context.Context) error {
+	return cmd.InfoFlag.Process(ctx)
+}
+
+func (cmd *info) Description() string {
+	return `Provides detailed infos about cluster rules, their types and rule members.
+
+Examples:
+  govc cluster.rule.info -cluster my_cluster
+  govc cluster.rule.info -cluster my_cluster -name my_rule`
+}
+
+func (cmd *info) Run(ctx context.Context, f *flag.FlagSet) error {
+	var res ruleResult
+
+	rules, err := cmd.Rules(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, rule := range rules {
+		ruleName := rule.GetClusterRuleInfo().Name
+		ruleInfo := GetExtendedClusterRuleInfo(rule)
+		if cmd.name == "" || cmd.name == ruleName {
+			res = append(res, fmt.Sprintf("Rule: %s", ruleName))
+			res = append(res, fmt.Sprintf("  Type: %s", ruleInfo.ruleType))
+			switch ruleInfo.ruleType {
+			case "ClusterAffinityRuleSpec", "ClusterAntiAffinityRuleSpec":
+				names, err := cmd.Names(ctx, *ruleInfo.refs)
+				if err != nil {
+					cmd.WriteResult(res)
+					return err
+				}
+
+				for _, ref := range *ruleInfo.refs {
+					res = append(res, fmt.Sprintf("  VM: %s", names[ref]))
+				}
+			case "ClusterVmHostRuleInfo":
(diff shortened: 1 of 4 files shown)


dougm:
Thanks @UBessle , I will review soon.  The `cluster.rule` test fails with your change, can you take a look?

```
not ok 10 cluster.rule
# (from function `assert_equal' in file ./test_helper.bash, line 206,
#  from function `assert_output' in file ./test_helper.bash, line 215,
#  from function `assert_success' in file ./test_helper.bash, line 190,
#  in test file ./cluster.bats, line 76)
#   `assert_success "$(printf "%s\n" DC0_C0_RP0_VM{0,1,2,3})"' failed
# expected: DC0_C0_RP0_VM0
# DC0_C0_RP0_VM1
# DC0_C0_RP0_VM2
# DC0_C0_RP0_VM3
# actual:   ClusterAffinityRuleSpec:
# DC0_C0_RP0_VM0
# DC0_C0_RP0_VM1
# DC0_C0_RP0_VM2
# DC0_C0_RP0_VM3
```

https://travis-ci.org/vmware/govmomi/builds/402096008

UBessle:
Hi Doug,

yes i've seen the problem with the test case. Unfortunately that was only after i've submitted my pull request.

But i'm not sure, what's the best way to proceed further. Here are the alternatives i can imagine

(1) I could fix the test code. We would stay with the extended and improved functionality of cluster.rule.ls allowing us to distinguish the different affinity rule types based on the output of cluster.rule.ls. But this would introduce a breaking change for the cluster.rule.ls function for other users of govc.

(2) I could remove the rule type info from the cluster.rule.ls output, so that the test doesn't break. But then we would stick with an incomplete information output about the listed affinity rule. When listing a rule, you couldn't decide from the output whether  the rule says, that two VM's must belong on one cluster node, or two VM's must belong to two different nodes, or if the two result strings name a VM group and a host group.

(3) I could remove the rule type info from the cluster.rule.ls output only for those rule types, that where supported by the original govc code (VM affinity rule and VM anti affinity rule). That would lead to a situation, where no test case breaks so we have a non-breaking change for other govc users and we could at least distinguish VM-to-Host-group rules from simple affinity rules. But then we would have a little bit inconsistent behaviour of cluster.rule.ls and as today nobody could distinguish between VM affinity rules and VM anti affinity rules based on the output from cluster.rule.ls

(4) I could introduce an additional option (-show-rule-type=false), causing the additional rule type info only to be shown if -show-rule-type=true is present on the command line. On this way i could introduce the extended information output as non-breaking change

What's your preferred alternative?

Cheers, Uwe

dougm:
Hi Uwe,

Thanks for the details, I'm just trying out your PR now.  In general, govc `ls` commands are just 1 line per item.  There are some `ls` commands that have multiple columns per row, such as `session.ls`.  Some may use a `-l` flag for example to include more detail/columns, but still just 1 line each.  The govc `info` commands use multiple lines per item, most in the format of:
```
Name: foo
  Type: some-type
  Members: one, two, three
Name: bar
  Type: another-type
  Members: four, five, six
``` 
And the `-json` flag on such commands includes the entire structure for each.  
The details for cluster rules probably don't fit well in an `ls` command, since each type has a different set of fields.  Maybe a new `cluster.rule.info` command would work better in this case, what do you think?

UBessle:
yeah, i could introduce a new `cluster.rule.info`command, to provide the extended information
```
Name: <rule name>
  Type: <rule type>
  <rule parameter 1 name>: <rule parameter 1 value>
  <rule parameter 2 name>: <rule parameter 2 value>
```
for example 
```
$ govc cluster.rule.info -dc=<datacenter-name> -cluster=<cluster-name>
Name: redhat_subscription 
  Type: ClusterVmHostRuleInfo
  VmGroupName: subscription_redhat_VMs
  AffineHostGroupName: redhat_subscriptions_hosts
  AntiAffineHostGroupName:
Name: shpoadb01_02_together
  Type: ClusterAffinityRuleSpec
  VM: shpoadb01
  VM: shpoadb02
```
This would introduce no breaking change and gives the user all the information that's needed.

The remaining question is, what to do with the `cluster.rule.ls` command. IMHO the current functionality is misleading. Currently 
* `cluster.rule.ls` without `-name` lists all rule names. 
* `cluster.rule.ls -name=rulename` lists all VM names for affinity and anti-affinity rules. 
* `cluster.rule.ls -name=rulename` lists nothing for ClusterVmHostRuleInfo and ClusterDependencyRuleInfo rules. 

I would suggest, that `cluster.rule.ls -name=rulename` lists the detail values for ClusterVmHostRuleInfo and ClusterDependencyRuleInfo rules without infos regarding the rule types and the different parameter meanings. For example
```
$ govc cluster.rule.ls -dc=<datacenter-name> -cluster=<cluster-name> -name=redhat_subscription 
subscription_redhat_VMs
redhat_subscriptions_hosts
$ govc cluster.rule.ls -dc=<datacenter-name> -cluster=<cluster-name> -name=shpoadb01_02_together
shpoadb01
shpoadb02
```
Additionally it could be helpful, to have a -l flag which shows rule types when listing all rules. For example
```
$ govc cluster.rule.ls -dc=<datacenter-name> -cluster=<cluster-name> -l
shpoadb01_02_together       ClusterAffinityRuleSpec
ioaotdb01_02_together         ClusterAffinityRuleSpec
windows_subscription            ClusterVmHostRuleInfo
redhat_subscription               ClusterVmHostRuleInfo
```
`cluster.rule.ls -l -name=` would provide the additional parameter types info when listing one rule. e.g.
```
$ govc cluster.rule.ls -dc=<datacenter-name> -cluster=<cluster-name> -name=redhat_subscription -l
subscription_redhat_VMs      VmGroupName
redhat_subscriptions_hosts   AffineHostGroupName
$ govc cluster.rule.ls -dc=<datacenter-name> -cluster=<cluster-name> -name=shpoadb01_02_together -l
shpoadb01                             VM
shpoadb02                             VM

dougm:
Thanks Uwe, this looks perfect!

PR #1177, opened 2018-07-15
devbox: use rsync instead of NFS

dougm:
Also removes Linux specific commands, so this can be used on MacOSX.


jiatongw:
It seems not work for me. ```./create.sh``` returns ```govc: Cannot complete file creation operation.``` after echo ```creating VM...``` in line 43. But for now I am able to run a k8s cluster in vsphere so we can discuss this later. 

FYI, how do I run k8s cluster in vsphere:
I created a VM, then either SSH or using remote console, ran a script I wrote which was referred from ```vcp-local-cluster-ip.sh``` and then the cluster was up.

PR #1184, opened 2018-07-20
Add method to move host into cluster

mjm:
Wraps the `MoveInto_Task` on `ClusterComputeResource`.

### DIFF
--- a/object/cluster_compute_resource.go
+++ b/object/cluster_compute_resource.go
@@ -69,11 +69,16 @@ func (c ClusterComputeResource) AddHost(ctx context.Context, spec types.HostConn
 	return NewTask(c.c, res.Returnval), nil
 }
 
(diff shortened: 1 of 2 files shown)


object/cluster_compute_resource.go
```diff
@@ -68,3 +68,17 @@ func (c ClusterComputeResource) AddHost(ctx context.Context, spec types.HostConn
 
 	return NewTask(c.c, res.Returnval), nil
 }
+
+func (c ClusterComputeResource) MoveInto(ctx context.Context, host *HostSystem) (*Task, error) {
```
dougm: Looks good, but maybe this should be `host ...*object.HostSystem` to support multiple hosts?

mjm: Of course! I was only using this with a single host so I didn't think of it, but you're right. I'll update it to support more than one host.

PR #1192, opened 2018-08-03
WaitOption.MaxWaitSeconds in simulator

thib-ack:
Hi,

I saw recently the new simulator package which is amazing, but I also had problems with the MaxWaitSeconds parameter of WaitForUpdatesEx() function; especially with the 0 value which means : 

> do one update calculation and return any results. [(Doc)](https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vmodl.query.PropertyCollector.WaitOptions.html#maxWaitSeconds)

With the current code, if I set a value of 0, it always returns **nothing, immediately**, even if there are updates that should be returned (with an exception of the very first call when Version == "").

The main problem in the current code (from what I understood) is coming from the line:
`simulator/property_collector.go:632`
`... = context.WithTimeout(context.Background(), time.Second*time.Duration(*max))`

which (in case of value *max = 0) creates a timer that fires immediately. 
And then on line,`simulator/property_collector.go:680`
inside the next `select` statement, we fall in the `context.DeadlineExceeded` case without even checking if there is anything updated

I propose a patch for this, 
If the MaxWaitSeconds is zero:
- we don't create a timer (context with timeout)
- we return an empty result instead of continue to loop when there is no updates available

I also added a testcase for this. Maybe the test is not optimal but it fails (does not return) without the patch.

Thanks,
Thibaut.

_PS: (I have to admit, I'm a little bit obsessed with the WaitOption.MaxWaitSeconds parameter of WaitForUpdates() ... cf. #741 ahah)_

### DIFF
--- a/simulator/property_collector.go
+++ b/simulator/property_collector.go
@@ -627,9 +627,14 @@ func (pc *PropertyCollector) apply(ctx *Context, update *types.UpdateSet) types.
 
 func (pc *PropertyCollector) WaitForUpdatesEx(ctx *Context, r *types.WaitForUpdatesEx) soap.HasFault {
 	wait, cancel := context.WithCancel(context.Background())
+	oneUpdate := false
 	if r.Options != nil {
 		if max := r.Options.MaxWaitSeconds; max != nil {
-			wait, cancel = context.WithTimeout(context.Background(), time.Second*time.Duration(*max))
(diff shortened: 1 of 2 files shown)


dougm:
@thib-ack would be great to get this merged, are you able to sign the CLA?

thib-ack:
Sorry for the delay, 
that's done.
Thanks

PR #1209, opened 2018-08-22
Avoid use of Finder all param in govc

### LINKED ISSUE #1208: NewFinder's "all" argument and -json || -dump flags
Passing the `-json` parameter dumps all the objects from vcenter and ignores filters. Why is it done this way?

https://github.com/vmware/govmomi/blob/badd514933ffc1d67c88e189fe16cb6d525be0d4/govc/flags/datacenter.go#L90

Is it reasonable just to add an `-all` option? In my current use case I'd like to receive the json for specific request, but `-json` just dumps everything.

dougm:
Use of '-json' flag in any command sets the Finder all param to true.
Only the 'ls' command makes use of this property collection, so limit it to there.

Fixes #1208

### DIFF
--- a/govc/flags/datacenter.go
+++ b/govc/flags/datacenter.go
@@ -77,7 +77,7 @@ func (flag *DatacenterFlag) Process(ctx context.Context) error {
 	})
 }
 
-func (flag *DatacenterFlag) Finder() (*find.Finder, error) {
+func (flag *DatacenterFlag) Finder(all ...bool) (*find.Finder, error) {
 	if flag.finder != nil {
 		return flag.finder, nil
 	}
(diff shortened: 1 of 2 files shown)


govc/flags/datacenter.go
```diff
@@ -87,7 +87,11 @@ func (flag *DatacenterFlag) Finder() (*find.Finder, error) {
 		return nil, err
 	}
 
-	finder := find.NewFinder(c, flag.JSON || flag.Dump)
+	allFlag := false
+	if len(all) != 0 {
```
kayrus: I'd add `if len(all) == 1 {`

dougm: sure, incoming.

PR #1210, opened 2018-08-22
govc: Print DVS rules for dvportgroup

kayrus:
ATM `dvs.portgroup.info` command doesn't print dvs rules. This is a WIP branch, am I on the right direction?

### DIFF
--- a/govc/build.sh
+++ b/govc/build.sh
@@ -8,6 +8,15 @@ git_version=$(git describe --dirty)
   exit 1
 fi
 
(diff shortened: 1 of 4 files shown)


govc/build.sh
```diff
@@ -8,6 +8,15 @@ git_version=$(git describe --dirty)
   exit 1
 fi
```
dougm: build.sh should be deleted since we moved to goreleaser for building releases, you can just use go install: https://github.com/vmware/govmomi/blob/master/CONTRIBUTING.md#installing-from-source

PR #1222, opened 2018-09-01
Misc clean up upon review

bruceadowns:


bruceadowns:
Hi Doug, something's odd because travis is complaining about gofmt, which is one of the things I attempted to fix. Further, doing a fresh clone and `make check` fails. I'm not sure how previous travis builds were succeeding.

bruceadowns:
It looks like the goimports behavior may differ between linux and macos.

dougm:
Which version of Go are you using?  Could be a difference between 1.10 that Travis CI uses vs 1.11 just released?

bruceadowns:
Both 1.10 and 1.11 behave the same for me. Is `make check` clean for you on macos?

I tracked it down a bit more, the gofmt/goimports difference I'm seeing is related to variable alignment when there is a variable name > 20 characters. On linux, variables that have a name > 20 characters are aligned separately than those with those whose name has <=20 characters. On macos, all members are always aligned.

bruceadowns:
And here's the golang bug - https://github.com/golang/go/issues/27177

According to the bug, the alignment problem described above has been fixed in go 1.11.

Note, I mistakenly considered this a linux vs macos issue. My mistake on macos is that I had not re-compiled goimports using go 1.10.

bruceadowns:
Hi @dougm, bumping the go version to 1.11 resolved the formatting complaints. Let me know if you'd like to good forward with this change or have other ideas.

PR #1221, opened 2018-08-30
Added basic PerfManager simulator functionality

prydin:
## Summary
Implemented the following methods: QueryAvailablePerfMetric, QueryPerf, QueryPerfCounter and QueryPerfProviderSummary.

In this release, the interval manipulation functions are not implemented, nor is QueryPerfComposite.

The performance data is generated from a small lab. I've captured 100 samples for each metric and that sequence is repeated to form a continuous time series. I've added some gaussian noise to the values to make it look more "realistic".

Tested with govc commands and the vSphere plugin for telegraf.

## Known issues:
*  Interval manipulation methods aren't implemented
* QueryPerfComposite isn't implemented
* Some data may not add up in a correct way. For example, instance data on datastores for sizes of VM files, swapfiles etc. may not add up to the total usage. 
* Does not work with vR Ops, due to missing internal methods.

PR #1230, opened 2018-09-11
Added PerformanceManager simulator

prydin:
## Summary
Implemented the following methods: QueryAvailablePerfMetric, QueryPerf, QueryPerfCounter and QueryPerfProviderSummary.

In this release, the interval manipulation functions are not implemented, nor is QueryPerfComposite.

The performance data is generated from a small lab. I've captured 100 samples for each metric and that sequence is repeated to form a continuous time series. I've added some gaussian noise to the values to make it look more "realistic".

Tested with govc commands and the vSphere plugin for telegraf.

## Known issues:
*  Interval manipulation methods aren't implemented
* QueryPerfComposite isn't implemented
* Some data may not add up in a correct way. For example, instance data on datastores for sizes of VM files, swapfiles etc. may not add up to the total usage. 
* Does not work with vR Ops, due to missing internal methods.


prydin:
This PR replaces #1221

Issue #1232, opened 2018-09-11
upload/download file into a VM failing


ticapix:
Hi,

I'm trying to use govc to upload/download file into a VM.
The client is correctly sending and receiving a InitiateFileTransferToGuest soap envelope with the URL to use for the GET or PUT operations.

Unfortunately, govc client doesn't seem to follow the spec https://www.vmware.com/support/developer/converter-sdk/conv50_apireference/vim.vm.guest.FileManager.html#initiateFileTransferToGuest and replace the hostname part of the url with the esx host IP, which in our case make the url invalid.

The code is in this function https://github.com/vmware/govmomi/blob/master/guest/file_manager.go#L130

Is there any reason for not changing the url hostname, only when it equals '*' ?

Thanks,

dougm:
Fixed by #1233

PR #1233, opened 2018-09-12
replace hostname only if unset

ticapix:
This a proposition to fix issue #1232 where TransferURL changes the hostname part of the url even if it's not set to '*' and hence not using the url as it is returned by the server.

(I took the specs from https://www.vmware.com/support/developer/converter-sdk/conv50_apireference/vim.vm.guest.FileManager.html#initiateFileTransferToGuest)

### DIFF
--- a/guest/file_manager.go
+++ b/guest/file_manager.go
@@ -128,6 +128,11 @@ func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error
 	}
 
 	needsHostname := turl.Hostname() == "*"
+
+	if needsHostname {
+		turl.Host = m.c.URL().Host // Also use Client's port, to support port forwarding
+	}
+
 	if !m.c.IsVC() {
 		return turl, nil // we already connected to the ESX host and have its thumbprint
 	}
--- a/guest/file_manager.go
+++ b/guest/file_manager.go
@@ -127,7 +127,7 @@ func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error
 		return nil, err
 	}
 
-	if !m.c.IsVC() {
+	if url.Hostname() != "*" {
 		return url, nil // we already connected to the ESX host and have its thumbprint
 	}
 
--- a/guest/file_manager.go
+++ b/guest/file_manager.go
@@ -122,25 +122,26 @@ func (m FileManager) DeleteFile(ctx context.Context, auth types.BaseGuestAuthent
 // The InitiateFileTransfer{From,To}Guest methods return a URL with the host set to "*" when connected directly to ESX,
 // but return the address of VM's runtime host when connected to vCenter.
 func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error) {
-	url, err := m.c.ParseURL(u)
+	turl, err := url.Parse(u)
 	if err != nil {
 		return nil, err
 	}
 
-	if url.Hostname() != "*" {
-		return url, nil // we already connected to the ESX host and have its thumbprint
+	needsHostname := turl.Hostname() == "*"
+	if !m.c.IsVC() {
+		return turl, nil // we already connected to the ESX host and have its thumbprint
 	}
 
-	name := url.Hostname()
-	port := url.Port()
+	name := turl.Hostname()
+	port := turl.Port()
 
 	m.mu.Lock()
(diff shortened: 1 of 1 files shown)


guest/file_manager.go
```diff
@@ -127,7 +127,7 @@ func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error
 		return nil, err
 	}
 
-	if !m.c.IsVC() {
+	if url.Hostname() != "*" {
```
dougm: The call above (`m.c.ParseURL`) already replaces `*` with the connection hostname: https://github.com/vmware/govmomi/blob/4316838a111c783776b1d9e06a6f7f61e507feeb/vim25/soap/client.go#L600-L602

So `url.Hostname()` would never be == `*` at this point, right?  Is your initial client connection directly to an ESX host or to vCenter?

ticapix: Indeed, my proposal is incorrect and `Hostname` should never be == `*`.

The current govc client behavior is, in case it's connected to vCenter, it replaces the url's hostname part by the ip of the ESX host.
In our setup, the ESX hosts have private IP making the `vc.Client.Upload` call to fail.
Our vCenter returns an url that shouldn't be modified by the client and be used directly as it is to access the ESX host.

To not break dependent packages and follow the specs, shouldn't the client try first the url as it is returned by the vCenter and modify the hostname only if the first attempt fails ?

dougm: ok, how about instead of calling `m.c.ParseURL()`, just call `url.Parse()` directly in place, something like:
```go
        needsHostname := turl.Hostname() = "*"
        if needsHostname {
                turl.Host = u.Host
        }
	if !m.c.IsVC() {
		return turl, nil // we already connected to the ESX host and have its thumbprint
	}
```
(note: renamed the `url` var to `turl` so the url package can be imported)

Then add an `if needsHostname {}` around this block of code, so we still collect the thumbprint without overwriting the hostname: https://github.com/vmware/govmomi/blob/4316838a111c783776b1d9e06a6f7f61e507feeb/guest/file_manager.go#L166-L183

ticapix: The go client is connected to vSphere via public internet.
The ESX host are in a private network and as a client, I can't access that network directly. 

The content of the InitiateFileTransferToGuestResponse from vSphere is
```xml
	<soapenv:Body>
		<InitiateFileTransferToGuestResponse xmlns="urn:vim25">
			<returnval>https://pcc-aaa-bbb-ccc-ddd.ovh.com:40050/guestFile?id=23&amp;token=xxx</returnval>
		</InitiateFileTransferToGuestResponse>
	</soapenv:Body>
```

My understanding is that because the client is connected to vSphere, `m.c.IsVC` will evaluate to `true` and will replace the hostname part of the url by the ESX host private IP. Later the `Upload` \ `Download` calls will fail

I don't expect that behavior.

I'm expecting the client to trust and follow the urls returned in the `InitiateFileTransferToGuestResponse` / `InitiateFileTransferFromGuestResponse`, without modification.

Is that possible ?

dougm: Yes, with the suggestions I made above, adding `if needsHostname {}` above avoids the hostname from being overwritten when `InitiateFileTransferToGuestResponse.returnval != "*"`.  And preserves the original behavior of overwriting when it is `== *`.  And with the suggested changes, the host thumbprint is collected in both cases, rather than just returning early.

guest/file_manager.go
```diff
@@ -122,25 +122,26 @@ func (m FileManager) DeleteFile(ctx context.Context, auth types.BaseGuestAuthent
...
 func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error) {
-	url, err := m.c.ParseURL(u)
+	turl, err := url.Parse(u)
 	if err != nil {
 		return nil, err
 	}
 
+	needsHostname := turl.Hostname() == "*"
 	if !m.c.IsVC() {
-		return url, nil // we already connected to the ESX host and have its thumbprint
```
dougm: Since the call to `m.c.ParseURL` is removed, we need to do this before returning: https://github.com/vmware/govmomi/blob/4316838a111c783776b1d9e06a6f7f61e507feeb/vim25/soap/client.go#L602

This should work with your patch:
```go
        if needsHostname {
                turl.Host = m.c.URL().Host
        }
```

PR #1237, opened 2018-09-20
govc: save sessions using sha256 ID

sigma:
As sha1 becomes less secure, let's move to sha256 where possible.

Since the client hashes the URI to determine what stored session to reuse,
potential collisions (as unlikely as they might be) could open the way to
session hijacking. Using sha256 affords better protection.

### DIFF
--- a/govc/flags/client.go
+++ b/govc/flags/client.go
@@ -18,7 +18,7 @@ package flags
 
 import (
 	"context"
-	"crypto/sha1"
+	"crypto/sha256"
 	"crypto/tls"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


dougm:
Thanks @sigma , lgtm.  Are you able to sign the CLA?

sigma:
@dougm working on it, CLAs are always fun :)

sigma:
@dougm is there a way to reset the cla-rejected label, or should I submit a new PR?

dougm:
Glad your CLA has been accepted @sigma , I went ahead and removed the label manually.

PR #1254, opened 2018-10-10
Include empty VirtualMachineConfigSpec annotation

bill-rich:
Include annotations in VirtualMachineConfigSpec even if they are empty.
Empty annotations are needed in order to clear an existing annotation.

### DIFF
--- a/vim25/types/types.go
+++ b/vim25/types/types.go
@@ -50136,7 +50136,7 @@ type VirtualMachineConfigSpec struct {
 	LocationId                   string                            `xml:"locationId,omitempty"`
 	GuestId                      string                            `xml:"guestId,omitempty"`
 	AlternateGuestName           string                            `xml:"alternateGuestName,omitempty"`
-	Annotation                   string                            `xml:"annotation"`
+	Annotation                   *string                           `xml:"annotation"`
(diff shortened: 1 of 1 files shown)


bill-rich:
Hey @dougm! Thanks for the feedback on this. Just want to make sure I understand the background on it. The idea is that with `string` someone may use `VirtualMachineConfigSpec` and have `Annotation` set to an empty string when they do not want to update the annotation. If we make it a pointer, they could set it to `nil` for no update and "" to remove the annotation. Is that correct?

dougm:
> set it to nil for no update and "" to remove the annotation

@bill-rich yes, exactly!

bill-rich:
Sorry for the delay in getting back to this. I got the string switched to a pointer and ran through all the scenarios (nil, "", or "something") to make sure they all work as expected. Better late than never (hopefully).

bill-rich:
Hey @dougm! I was looking through my open PRs and realized I never addressed this. I was going through and looking at it, and I'm ok with the `SetAnnotation` method, but I also don't want to push a hack on you. Especially for something as trivial as not being able to set the annotation. For now, I'll close this out and make a note on the Terraform side with a workaround. Thanks for all the help (and patience) on this!

PR #1260, opened 2018-10-12
vcsim: add fault message to PropertyCollector RetrieveProperties

ausevj:
Found inconsistent Fault message comparing to real vCenter. When trying to retrieve object which is not found, I should get the message "The object has already been deleted or has not been completely created". 

Steps to test:
1. Create ResourcePool
2. Destroy ResourcePool
3. Try RetrieveProperties with deleted pool identifier


ausevj:
@dougm we had a test that looked for this fault string (don't know why the fault type was not used), figured it's an easy change to implement, so my use case is for another program.

PR #1267, opened 2018-10-17
Add vslm package and govc disk commands

### LINKED ISSUE #1238: Add support for First Class Disk (FCD)
The generated code for FCD (VcenterVStorageObjectManager) is already in place, but no abstractions to use it. The abstractions need to be added, and support through vcsim for testing.

dougm:
- vslm.ObjectManager provides a higher level API for First Class Disk methods

- New govc disk.* and disk.snapshot.* commands manage FCDs and their snapshots

Fixes #1238

### DIFF
--- a/govc/disk/create.go
+++ b/govc/disk/create.go
@@ -1,0 +1,103 @@
+/*
+Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+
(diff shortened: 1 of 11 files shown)


dvonthenen:
LGTM

PR #1292, opened 2018-11-02
vcsim: update VM configureDevices method to change VM properties with UpdateObject

### LINKED ISSUE #1288: vcsim: Map.Update on objects that have not been added to the registry yet
Was going to update `VirtualMachine` methods where properties are set directly to use the new method `Map.Update` instead. Trying to update a property VCSIM fails due to VM not being created yet.

Example:
```diff
govmomi/simulator/virtual_machine.go
@@ -694,7 +707,15 @@ func (vm *VirtualMachine) configureDevices(spec *types.VirtualMachineConfigSpec)
                }
        }

-       vm.Config.Hardware.Device = []types.BaseVirtualDevice(devices)
+       Map.Update(vm, []types.PropertyChange{
+               {Name: "config.hardware.device", Val: []types.BaseVirtualDevice(devices)},
+       })

        return nil
 }
```
always fails when creating VM, because VM is not added to the registry yet. The type in this line is empty string:
https://github.com/vmware/govmomi/blob/ef8efd52a8412f4f30c4c1e99c93e8128141ebef/vim25/mo/retrieve.go#L71

I can check if VM is created, setting property directly if it's not and updating if it is (first time configuring devices - creating VM - it sets properties, when adding additional device - it sends update), but this is not a clean solution when many properties need updating:
```diff
govmomi/simulator/virtual_machine.go
@@ -694,7 +707,15 @@ func (vm *VirtualMachine) configureDevices(spec *types.VirtualMachineConfigSpec)
                }
        }

-       vm.Config.Hardware.Device = []types.BaseVirtualDevice(devices)
+       if vm.Self.Value != "" { // moid isn't set until CreateVM is done
+               Map.Update(vm, []types.PropertyChange{
+                       {Name: "config.hardware.device", Val: []types.BaseVirtualDevice(devices)},
+               })
+       } else {
+               vm.Config.Hardware.Device = []types.BaseVirtualDevice(devices)
+       }

        return nil
 }
```

Wondering if there is a way to rework `Map.Update` (and/or any method it uses) to set properties if object has not been created yet and do not send the updates, or is there any better solution?

ausevj:
VM is now added to registry during `NewVirtualMachine` call. `folder.putChild(...)` is called at the start of the function so it always completes, if VM creation fails - it is up to callers to cleanup (`folder.removeChild`). This might be confusing as caller needs to know that VM is added to folder inside and it's up to him to catch a fault and cleanup. This is done to be safer - if cleanup was up to `NewVirtualMachine` it would need to catch all errors (and cleanup then) - any error not caught would result it no cleanup.

Fixes #1288

### DIFF
--- a/simulator/folder.go
+++ b/simulator/folder.go
@@ -275,6 +275,7 @@ func hostsWithDatastore(hosts []types.ManagedObjectReference, path string) []typ
 func (c *createVM) Run(task *Task) (types.AnyType, types.BaseMethodFault) {
 	vm, err := NewVirtualMachine(c.Folder.Self, &c.req.Config)
 	if err != nil {
+		c.Folder.removeChild(vm)
 		return nil, err
 	}
 
@@ -312,10 +313,9 @@ func (c *createVM) Run(task *Task) (types.AnyType, types.BaseMethodFault) {
 
 	err = vm.create(&c.req.Config, c.register)
 	if err != nil {
+		c.Folder.removeChild(vm)
 		return nil, err
 	}
-
-	c.Folder.putChild(vm)
 
 	host := Map.Get(*vm.Runtime.Host).(*HostSystem)
 	Map.AppendReference(host, &host.Vm, vm.Self)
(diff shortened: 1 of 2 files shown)


ausevj:
@dougm I pushed empty change amend to trigger Travis CI. #1297 changes fixed `vm.clone` issue. Ready for another review

PR #1298, opened 2018-11-08
Tags support for First Class Disks

dougm:
- Add tag related methods to vslm.ObjectManager

- Add vcsim support for vslm.ObjectManager tag methods

- Add govc disk.ls -c and -t options to list tagged disks

- Add govc disk.ls -T option to list tags attached to disks

- Add govc disk.tags.attach and disk.tags.detach commands

### DIFF
--- a/govc/disk/ls.go
+++ b/govc/disk/ls.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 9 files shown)


dvonthenen:
Will take a look at this when I land at home... just about to take off for my connecting flight!

PR #1299, opened 2018-11-09
vcsim: update VM snapshot methods to change VM properties with UpdateObject

ausevj:
Updated VM Snapshot methods (Create, Remove, RemoveAll, Revert) to change VM properties with UpdateObject instead of setting directly.

### DIFF
--- a/simulator/snapshot.go
+++ b/simulator/snapshot.go
@@ -29,20 +29,30 @@ type VirtualMachineSnapshot struct {
 
 func (v *VirtualMachineSnapshot) RemoveSnapshotTask(req *types.RemoveSnapshot_Task) soap.HasFault {
 	task := CreateTask(v, "removeSnapshot", func(t *Task) (types.AnyType, types.BaseMethodFault) {
+		var changes []types.PropertyChange
+
 		Map.Remove(req.This)
 
 		vm := Map.Get(v.Vm).(*VirtualMachine)
 		Map.WithLock(vm, func() {
 			if vm.Snapshot.CurrentSnapshot != nil && *vm.Snapshot.CurrentSnapshot == req.This {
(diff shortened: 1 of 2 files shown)


simulator/snapshot.go
```diff
@@ -29,20 +29,30 @@ type VirtualMachineSnapshot struct {
...
+		var changes []types.PropertyChange
+
 		Map.Remove(req.This)
 
 		vm := Map.Get(v.Vm).(*VirtualMachine)
 		Map.WithLock(vm, func() {
 			if vm.Snapshot.CurrentSnapshot != nil && *vm.Snapshot.CurrentSnapshot == req.This {
-				parent := findParentSnapshotInTree(vm.Snapshot.RootSnapshotList, req.This)
-				vm.Snapshot.CurrentSnapshot = parent
+				var parentSnapshot types.ManagedObjectReference
```
dougm: We should allow `CurrentSnapshot` to be nil, rather than an in invalid ManagedObjectReference.  How about this instead:
```go
				current := types.PropertyChange{Name: "snapshot.currentSnapshot", Val: nil}
				parent := findParentSnapshotInTree(vm.Snapshot.RootSnapshotList, req.This)
				if parent != nil {
					current.Val = *parent
				}
				changes = append(changes, current)
```

dougm: btw, we should be able to use a pointer type for `Val`, I'm going to look at fixing that.  The code was originally written to inflate SOAP -> Go on the client side, before looking at the field types those values would be applied to.

dougm: ok, with #1300 this can be simplified to:
```go
			parent := findParentSnapshotInTree(vm.Snapshot.RootSnapshotList, req.This)
			changes = append(changes, types.PropertyChange{Name: "snapshot.currentSnapshot", Val: parent})
```

PR #1300, opened 2018-11-09
Allow pointer values in mo.ApplyPropertyChange

dougm:
The code was originally written for client side SOAP -> Go inflation,
where struct types were never considered pointers on the wire.
Now that we are using this method in-memory via simulator.Registry.Update,
it is possible to use pointer values.


dougm:
@ausevj can you review?

ausevj:
@dougm Looks good

PR #1306, opened 2018-11-13
vcsim: populate VM guest.net field

dougm:
- Add a guest.net entry for each network device

- Propagate SET.guest.ipAddress to guest.net[0].ipAddress

### DIFF
--- a/simulator/virtual_machine.go
+++ b/simulator/virtual_machine.go
@@ -269,6 +269,20 @@ func (vm *VirtualMachine) apply(spec *types.VirtualMachineConfigSpec) {
 			continue
 		}
 		changes = append(changes, types.PropertyChange{Name: key, Val: val.Value})
+
+		switch key {
+		case "guest.ipAddress":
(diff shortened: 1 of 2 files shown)


simulator/virtual_machine.go
```diff
@@ -269,6 +269,20 @@ func (vm *VirtualMachine) apply(spec *types.VirtualMachineConfigSpec) {
 			continue
 		}
 		changes = append(changes, types.PropertyChange{Name: key, Val: val.Value})
+
+		switch key {
+		case "guest.ipAddress":
+			ip := val.Value.(string)
+			vm.Guest.Net[0].IpAddress = []string{ip}
```
akutz: Hi @dougm,

Looks good. Was hoping to see `if key =~ /guest.net.(.+)/` and translate anything there to `guest.net[0].$1`. Will check if this works as is with IP address alone.

dougm: We can still do that if needed.

PR #1310, opened 2018-11-14
govc: add category option to relevant tags commands

### LINKED ISSUE #1282: govc: attach specific tag when tags with the same name exist in multiple categories
Hi,

Is it possible to add category flag, at least optionally, to the command `govc tags.attach`?
As in my case we have few tags with the same name in different categories.

Thanks.

dougm:
Fixes #1282


dougm:
@t3mi can you review?

PR #1314, opened 2018-11-16
vcsim: support base types in property filter

### LINKED ISSUE #1251: vcsim: Updates are not sent when PropertyFilter contains type that is extended by updated object
This should be changed to check if updated object is a subtype of the one in PropertyFilter:
https://github.com/vmware/govmomi/blob/9fb1bd9464057df0f3229d18626eee4540871f41/simulator/property_filter.go#L49-L53

**e.g.** If PropertyFilter contains type `ComputeResource` and I make changes to `ClusterComputeResource`, the types should match.
Looking at the lines above if `p.Type == "ComputeResource" && ref.Type == "ClusterComputeResource"` should be a valid match. Should be the same with all types/subtypes, like DVP and Network (not tested).

**Reproduction steps** (WaitForUpdate script can be taken from #1235):
1. Create a filter that looks for `ComputeResource` changes
2. `WaitForUpdate`
3. Make any change to `ClusterComputeResource`

dougm:
Fixes #1251

### DIFF
--- a/govc/test/vcsim.bats
+++ b/govc/test/vcsim.bats
@@ -111,3 +111,23 @@ load test_helper
   run govc vm.create bar.yakity
   assert_success
 }
+
+@test "vcsim issue #1251" {
+  vcsim_env
+
+  govc object.collect -type ComputeResource -n 1 / name &
+  pid=$!
+
+  run govc object.rename /DC0/host/DC0_C0 DC0_C0b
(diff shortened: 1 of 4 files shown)


dougm:
@ausevj can you review?

ausevj:
Looks good, worked for me

PR #1318, opened 2018-11-20
Changes pbm simulator

davinderkvnera:
1. Changing the pbm client's path as java client is expecting /pbm.
 2. Added PbmRetrieveServiceContent method in the unauthorized list.


davinderkvnera:
I will reopen again after joining the vmware org in github.

PR #1322, opened 2018-11-26
Added Support for PutUsbScanCodes on each VM

king-jam:
This adds this capability to govc: https://www.virtuallyghetto.com/2017/09/automating-vm-keystrokes-using-the-vsphere-api-powercli.html

### DIFF
--- a/govc/vm/putusbscancodes.go
+++ b/govc/vm/putusbscancodes.go
@@ -1,0 +1,201 @@
+/*
+Copyright (c) 2014-2016 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vm
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"strconv"
+
+	"github.com/vmware/govmomi/govc/cli"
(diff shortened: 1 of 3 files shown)


govc/vm/putusbscancodes.go
```diff
@@ -0,0 +1,316 @@
...
+func (cmd *putusbscancode) Usage() string {
+	return ""
+}
+
+func (cmd *putusbscancode) Description() string {
+	return `Send Keystroke to VM.
+
+Examples:
+ Default Scenario
+  govc vm.putusbscancode $vm -r 1376263 (writes an 'r' to the console)
```
dougm: Tried it out, the examples should be `-vm $vm`.  Or to support both, can do this (and add "VM" to Usage): https://github.com/vmware/govmomi/blob/cf359b62d1e6a76bcfeee0d5cc1bfb8f7db685b7/govc/vm/console.go#L82-L91

govc/vm/putusbscancodes.go
```diff
@@ -0,0 +1,316 @@
...
+  govc vm.putusbscancode $vm -c 0x15 (writes an 'r' to the console)
```
dougm: Can you make these a comment instead of parens `# writes an 'r' to the console`

king-jam:
@dougm Updated per your feedback.

I didn't end up using the console.go example because I encountered an issue with the VM Search function reporting `no arguments` when I provided the `-vm` flag or if I defined  `GOVC_VM`. Bug?

yusufozturk:
Hi all,

First of all, thanks for this extension in the API. @king-jam @dougm 

Our automation process uses PutUsbScanCodes to wake up virtual machines so we can capture their console screens and put it into the AI engine so we can process if there is any anomality on the virtual machine screen.

Everything works great except everytime when we run PutUsbScanCodes, it creates an event on vCenter. There are thousand of virtual machines and we do this process every 5 minutes due to screening and compliance checks so you can image the mess on the vcenter event console.

image

Is there any way to supress the event? If there is no way to supress this event, do you think we can do this via another way?

Thanks,

Yusuf

PR #1326, opened 2018-12-06
govc: add device name match support to device.ls and device.remove

dougm:
The device.info command already supports name matching, extend to the ls and remove commands.


dougm:
@maplain can you review?

maplain:
cool. it's nice to have it there. lgtm.

dougm:
Thanks @maplain

PR #1327, opened 2018-12-06
ssoadmin:create local group and add users to group

AndrewDi:
Make ssoadmin to support `CREATE GROUP` and `Add Users To Group`.

### DIFF
--- a/ssoadmin/client.go
+++ b/ssoadmin/client.go
@@ -160,6 +160,28 @@ func (c *Client) DeletePrincipal(ctx context.Context, name string) error {
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


ssoadmin/client.go
```diff
@@ -160,6 +160,39 @@ func (c *Client) DeletePrincipal(ctx context.Context, name string) error {
 	return err
 }
 
+func (c *Client) AddUsersToGroup(ctx context.Context, groupName string, userIDs []types.PrincipalId) error {
```
dougm: How about we make this a variadic param, `userIDs ...types.PrincipalId`?  That way it can be called with 1 or more IDs and the `AddUserToGroup` method below can be removed.

AndrewDi: That's great!

PR #1351, opened 2019-01-18
vcsim: avoid zero IP address in GOVC_URL output

dougm:
If the '-httptest.serve' flag is specified without an address,
the default GOVC_URL output will be unusable.  Attempt to replace
with a valid IPv4 address.

- Add vcsim '-l' flag as an alias for '-httptest.serve'

- Fix vcsim_stop race; need to wait for process to exit

### DIFF
--- a/simulator/simulator.go
+++ b/simulator/simulator.go
@@ -523,6 +523,37 @@ func (*Service) ServiceVersions(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprint(w, versions)
 }
 
+// defaultIP returns addr.IP if specified, otherwise attempts to find a non-loopback ipv4 IP
+func defaultIP(addr *net.TCPAddr) string {
+	if !addr.IP.IsUnspecified() {
+		return addr.IP.String()
+	}
+
+	nics, err := net.Interfaces()
+	if err != nil {
+		return addr.IP.String()
+	}
+
+	for _, nic := range nics {
+		if nic.Name == "docker0" || strings.HasPrefix(nic.Name, "vmnet") {
+			continue
+		}
+		addrs, aerr := nic.Addrs()
+		if aerr != nil {
+			continue
+		}
+		for _, addr := range addrs {
+			if ip, ok := addr.(*net.IPNet); ok && !ip.IP.IsLoopback() {
+				if ip.IP.To4() != nil {
+					return ip.IP.String()
+				}
+			}
+		}
+	}
+
+	return addr.IP.String()
+}
+
 // NewServer returns an http Server instance for the given service
 func (s *Service) NewServer() *Server {
 	s.RegisterSDK(Map)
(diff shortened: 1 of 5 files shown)


vcsim/README.md
```diff
@@ -88,9 +88,9 @@ The default vcsim listen address is `127.0.0.1:8989`.  Use the `-httptest.serve`
 
 
 ``` shell
-vcsim -httptest.serve=10.118.69.224:8989 # specific address
+vcsim -l 10.118.69.224:8989 # specific address
 
-vcsim -httptest.serve=:8989 # any address
+vcsim -l :8989 # any address
```
alexellis: This is nice :+1:

simulator/simulator.go
```diff
@@ -523,6 +523,37 @@ func (*Service) ServiceVersions(w http.ResponseWriter, r *http.Request) {
...
+		return addr.IP.String()
+	}
+
+	nics, err := net.Interfaces()
+	if err != nil {
+		return addr.IP.String()
+	}
+
+	for _, nic := range nics {
+		if nic.Name == "docker0" || strings.HasPrefix(nic.Name, "vmnet") {
```
alexellis: I wonder if you could use a stub or Go interface to be able to unit test this logic?

vcsim/main.go
```diff
@@ -63,6 +63,7 @@ func main() {
 	cert := flag.String("tlscert", "", "Path to TLS certificate file")
 	key := flag.String("tlskey", "", "Path to TLS key file")
 	env := flag.String("E", "-", "Output vcsim variables to the given fifo or stdout")
+	listen := flag.String("l", "127.0.0.1:8989", "Listen address for vcsim")
```
alexellis: How does listen differ from httptest.serve?

vcsim/main.go
```diff
@@ -110,12 +111,10 @@ func main() {
...
 	f := flag.Lookup("httptest.serve")
-	listen := f.Value.String()
-	if listen == "" {
-		listen = "127.0.0.1:8989"
-		_ = f.Value.Set(listen)
+	if f.Value.String() == "" {
+		_ = f.Value.Set(*listen) // propagate -listen unless -httptest.serve is specified
 	}
-	if err = updateHostTemplate(listen); err != nil {
+	if err = updateHostTemplate(*listen); err != nil {
```
alexellis: Is there any risk of a nill dereference?

vcsim/README.md
embano1: Very user friendly!

simulator/simulator.go
dougm: The new vcsim.bats test covers this well enough I think, the return value is just used in the `GOVC_URL` output.

vcsim/main.go
dougm: It's just an alias, they behave the same.  But @lamw had suggested a while back that `httptest.serve` wasn't obvious.  Decided to address that while I was at it.

vcsim/main.go
dougm: No, `listen` is returned by `flag.String` in this case.

alexellis:
FYI @ivanayov

PR #1368, opened 2019-02-05
Fix #1173

### LINKED ISSUE #1173: Incomplete list of VMs returned with ls
I'm not sure if this is a case of RTFM (or lack thereof) but it seems that when I try and get a list of all VMs, the number of machines returned is not the total number of machines in the Datacentre.

To reproduce (tried it on different instances / datacentres)
```
[~] govc ls -l '*' | grep "VirtualMachine" |wc -l
     170
```
(should be 228, vcentre 6.5)
```
[~] govc ls -l '*' | grep "VirtualMachine" |wc -l
     96
```
(should be 124, vcentre 5.5)

Is this a govc issue? Or a human error?

Alternatively, 

`[~] govc ls /<DC name>/vm | wc -l` returns 177 (for the first cluster)

amandahla:
I changed to get all VMs but it's a little bit slow because I'm retrieving the config.template property for each VM otherwise will count Templates as VM too.

### DIFF
--- a/govc/datacenter/info.go
+++ b/govc/datacenter/info.go
@@ -29,6 +29,8 @@ import (
 	"github.com/vmware/govmomi/govc/flags"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/view"
+	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 )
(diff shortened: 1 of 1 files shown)


amandahla:
@dougm Much faster, thanks! Do you think is necessary to change others too to not use finder?

dougm:
@amandahla looking good, can you also remove "WIP" from the commit message?

There might be other places where `ContainerView` would be better.  In general, ContainerView is best when you want all instances, Finder is better for specific/subset of instances.  I think Finder could also be optimized to use ContainerView underneath in the `*` pattern case at least.

amandahla:
@dougm sure, done!

Thanks!

amandahla:
@dougm no problem, thanks for the comments.

amandahla:
Glad to help! :-)

PR #1302, opened 2018-11-12
build: Refactored Travis-CI to use containers

akutz:
This patch refactors the Travis-CI builds to use containers when possible. This will shorten the build queue times dramatically. Only when deploying build artifacts for tagged releases will Travis-CI require the VM infrastructure.

The phrase `/ci-deploy` may also be added to the commit message in order to activate the deploy stage for non-release builds.

Please see this [forked build](https://travis-ci.com/akutz/govmomi/builds/91010365) for an example of the new config in action. This PR's [build](https://travis-ci.org/vmware/govmomi/builds/453784582) also demonstrates the new config, but this link will change every time the PR is updated.

PR #1370, opened 2019-02-07
Fix travis deploy

dougm:
The on tags section was removed in #1302

Causing deploy to fail:
 Skipping a deployment with the script provider because this branch is not permitted: v0.20.0
 https://travis-ci.org/vmware/govmomi/jobs/489881920#L648

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ jobs:
     - <<:            *lint-stage
       env:           LINTER=goimports
 
-    # The "build" stage verifies the program can be built against the 
+    # The "build" stage verifies the program can be built against the
     # various GOOS and GOARCH combinations found in the Go releaser
     # config file, ".goreleaser.yml".
     - &build-stage
@@ -89,6 +89,9 @@ jobs:
       - provider:     script
         skip_cleanup: true
         script:       curl -sL http://git.io/goreleaser | bash
+        on:
+          tags: true
+          condition: $TRAVIS_OS_NAME = linux
       addons:
         apt:
           update:     true


akutz:
Hi @dougm,

Hmm. I *did* remove that, but it was because I added a dedicated deploy stage. I see the error, but I'm not sure the correct solution going forward is to revert to the former Travis-CI config.

That said, if you want to do that in order to push a build, and then we can figure out the error caused by the deploy stage change, that's fine with me. I certainly don't want to be a blocker here.

akutz:
Hi @dougm,

Based on [this](https://github.com/travis-ci/travis-ci/issues/8289) issue, I recommend the following:

1. Remove the line `if:            tag IS present OR commit_message =~ /\/ci-deploy/` from the `deploy` stage
2. Add the same `on` condition in this PR, but to the `deploy` section in the `deploy` stage

.travis.yml
```diff
@@ -89,6 +89,10 @@ jobs:
       - provider:     script
         skip_cleanup: true
         script:       curl -sL http://git.io/goreleaser | bash
+        on:
+          tags: true
+          condition: $TRAVIS_OS_NAME = linux
+          go: '1.11'
```
akutz: Hi @dougm,

I think you can drop the Go version since that is set globally.

dougm:
Thanks @akutz , I'm not sure if it's the right fix either.  We can take a look in the morning.

I was going to do a manual release, but that isn't working either for me (using linux/GNU Make 4.1):

```
% ./release.sh release
Building govc...
++ go env GOPATH
+ make -C /home/vmware/src/github.com/vmware/govmomi/. -j build-all
make: Entering directory '/home/vmware/src/github.com/vmware/govmomi'
make: *** No rule to make target 'build-all'.  Stop.
make: Leaving directory '/home/vmware/src/github.com/vmware/govmomi'
```

dougm:
#1375

PR #1374, opened 2019-02-07
Fix staticcheck issues

mjtrangoni:
Hi @dougm,

I fixed some [staticcheck](https://github.com/dominikh/go-tools) issues here. If you agree with this, I can fix the rest. This check should also be added to TravisCI in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 secrets.yml
 dist/
 .idea/
+
+# Ignore editor temp files
+*~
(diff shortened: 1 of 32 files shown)


mjtrangoni:
@dougm PTAL. I dropped the `vim25/xml` changes.

govc/vm/info.go
```diff
@@ -101,12 +101,10 @@ func (cmd *info) Run(ctx context.Context, f *flag.FlagSet) error {
...
-	if err != nil {
-		if _, ok := err.(*find.NotFoundError); ok {
-			// Continue with empty VM slice
-		} else {
-			return err
-		}
+	if _, ok := err.(*find.NotFoundError); ok {
+		// Continue with empty VM slice
+	} else {
+		return err
```
dougm: We don't want to return `err` here if it is `nil`, looks like this is what's causing CI to fail - at least if fails locally for me unless this hunk is reverted.

mjtrangoni: @dougm Good catch! I reverted it for the moment. Thanks!

mjtrangoni:
@dougm PTAL. I fixed more trivial issues. 
The current status is, 
```
$ staticcheck ./... | wc -l 
40
```

PR #1375, opened 2019-02-07
(WIP) Release process rehaul

frapposelli:
> DO NOT MERGE this PR, upstream work in `goreleaser` is not complete yet.

This PR intends to rehaul the release process for govc, adding vcsim to the list of artifacts published at every release, and keeping the door open for more (like `toolbox`).

Changes:
- Move to a single, autogenerated, changelog. the changelog creation is automated with `git-chlog` keeping the same standard structure as before. The old changelogs are preserved and point to the new one.
- `build.sh` and `release.sh` are removed, replaced by `goreleaser` and `git-chlog`.
- `RELEASE.md` was added, to explain and document the steps required to cut a govmomi release.
- `vcsim` was added as published artifact (binary, brew and docker image)
- Both `vcsim` and `govc` are compiled for `arm` and `arm64` platforms. 

/cc @dougm @akutz @lamw


vcsim/main.go
```diff
@@ -42,6 +42,12 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
+// vcsim does not track versions like govc does, it only shows if a binary was
+// built during a tag event or from a git checkout.
+const Version = "dev"
```
dougm: We could also move `govc/flags.Version` somewhere common and use it for both vcsim and govc.

CONTRIBUTORS
```diff
@@ -27,15 +27,15 @@ bastienbc <bastien.barbe.creuly@gmail.com>
 Bob Killen <killen.bob@gmail.com>
 Brad Fitzpatrick <bradfitz@golang.org>
 Bruce Downs <bruceadowns@gmail.com>
-Cédric Blomart <cblomart@gmail.com>
```
dougm: btw, this diff seems to be a difference with `sort` between Linux and Mac.

frapposelli: yeah, I noticed as well, will investigate.

vcsim/main.go
frapposelli: good point, maybe it's time to move version to a top-level package

CONTRIBUTORS
frapposelli: I standardized `scripts/contributors.sh` on GNU sort, but running it on Mac does not change the outcome... interesting...

vcsim/main.go
frapposelli: made the move, now it's shared amongst all the packages (including toolbox)

frapposelli:
Pushed an update with some improvements and added `toolbox` to the list of binaries built.

@akutz agree on your suggested changes, will make a separate commits as I need to take a second look at the current Makefile, I focused mostly on `goreleaser` as it is compiling the binaries independently, without calling the Makefile.

embano1:
Closing this as #2340 and following PRs will fix it using Github Actions.

PR #1377, opened 2019-02-10
Add content library support

### LINKED ISSUE #735: vm.clone: add support for 'Content Library' as template source?
Hi,

I literally just found out about a feature that has been added in vCenter 6.0 called 'Content Library', because one of the environments I work in started adapting this new feature for exclusively storing their VM templates in.

How much effort would be involved to support a 'Content Library' as a template source for cloning new VMs?

I haven't done any research yet on this topic, and I don't know how many administrators have adapted this feature. But I thought I'd put the word out and see what other people think.

markpeek:
Content libraries are a centralized repository for managing and distributing files (such as VM's, templates, ISO images) on and across vcenter servers. This change by @markpeek and @akutz implements support for content library api's and govc commands to work with content libraries, items, and deploying templates.

Note: there are more changes planned to improve this support. Merging this (large) change would be helpful to allow more people to use, test, and contribute to the further development.

Resolves #735.


vapi/library/finder/finder.go
```diff
@@ -0,0 +1,309 @@
...
+	// Get the library item files.
+	return f.findLibraryItemFiles(ctx, items, parts[2])
+}
+
+// FindResult is the type of object returned from a Find operation.
+type FindResult interface {
+
+	// GetParent returns the parent of the find result. If the find result
+	// is a Library then this function will return nil.
+	GetParent() FindResult
```
dougm: Doesn't matter too much, but: https://golang.org/doc/effective_go.html#Getters
If the `Get` prefix is go avoid clashing with field names, just leave as-is.

vapi/library/library_item_updatesession_file.go
```diff
@@ -0,0 +1,126 @@
...
+func GetContentLengthAndFingerprint(
+	ctx context.Context, uri string) (int64, string, error) {
+	resp, err := http.Head(uri)
+	if err != nil {
+		return 0, "", err
+	}
+	if resp.TLS == nil || len(resp.TLS.PeerCertificates) == 0 {
+		return resp.ContentLength, "", nil
+	}
+	fingerprint := &bytes.Buffer{}
```
dougm: Can use `soap.ThumbprintSHA1` here.

akutz:
Hi @dougm,

> Mark, we discussed removing the `library.item.update.*` commands as they were for testing, rather than end user, right?

I can't speak for @markpeek, but he and I did discuss some of the commands and whether they should stay or go. I think he left them in because I said that there's no other good way to access that data or perform those actions unless someone writes test code to do it. It's kind of useful to have access to the data/actions in the library API as there's no other good way to interact with those calls.

vapi/library/finder/finder.go
```diff
@@ -0,0 +1,309 @@
...
+// of a wildcard character in the search string results in a full listing of
+// that part of the path's server-side items.
+//
+// Path parts that do not use wildcard characters rely on server-side Find
+// functions to find the path token by its name. Ironically finding one
+// item with a direct path takes longer than if a wildcard is used because
+// of the multiple round-trips. Direct paths will be more performant on
+// systems that have numerous items.
+type Finder struct {
+	M *library.Manager
```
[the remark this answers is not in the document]
akutz: Hi @dougm,

I exported it in case someone wanted to create a `Finder` without the `NewFinder` function. The only reason it's named `M` is because I loathe long lines of code, but sure, it can be renamed.

vapi/library/finder/finder.go
akutz: Hi @dougm,

While the docs don't state so directly, I believe the suggestion is more applicable to `struct` types rather than `interface` types. Personally I think I should have made `FindResult` a `struct`, except originally it was an `interface` because there were more than one `findResult` `struct` variants.

Anyway, I made the functions have `Get` prefixes because as an interface I cannot know or assume that the structs that may implement the interface won't have conflicting, exported fields or functions. I thought `Get` prefixes would help avoid possible collisions.

dougm:
ok, let's leave the update commands for now, I will take a closer look when implementing vcsim support.

vapi/library/library_item_updatesession_file.go
akutz: Hi @dougm,

I suppose, but I don't think it's a big deal either way.

akutz:
Hi @dougm,

Your suggestion for removing the commands did get me thinking. What about an environment variable `GOVMOMI_ADVANCED=true` that causes additional commands to be registered? Otherwise certain commands will not appear to normal users?

vapi/library/finder/finder.go
akutz: Hmm, now I remember the reason I didn't make `FindResult` a `struct`, even after removing any variants of the `findResult` type. Maybe I have it wrong, but I thought it is illegal to have a struct type reference itself in a field. However, I just tested that ([playground link](https://play.golang.org/p/dyh9dSZM8AX), and it doesn't appear to be the case.

dougm:
> `GOVC_ADVANCED=true` 

Yeah that'd be fine too.

dougm:
I think it'll be easier to work on the suggestions after merging.  I'll take care of the `flags.*` suggestions and get going with vcsim support.

@markpeek just let me know what you think about rename `vcenter.deploy` to `library.deploy`.  If we can do that pre-merge, that'd be better for git history, but not a big deal either way.  Thanks again!

markpeek:
Oops...was in meetings and going to respond. Looks like it got merged. :-) Thanks!

Yes, will address the feedback in follow up PR's.

dougm:
@markpeek yeah easier having it merged.  I am working on the DatastoreFlag suggestion now.  Let's also coordinate via slack or other GH issues, who's working on what so we don't duplicate :)

govc/vcenter/deploy.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+type deploy struct {
+	*flags.ClientFlag
+	*flags.DatacenterFlag
+	datastore string
+}
+
+func init() {
+	cli.Register("vcenter.deploy", &deploy{})
```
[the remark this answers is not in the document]
lamw: 👍 

You're deploying a library item out of Content Library, not vCenter and actually took me awhile to find as I was assuming all **library** operations would be prefix with library :)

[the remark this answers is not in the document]
markpeek: Per @dougm comment above and @lamw new comment, likely we should rename this command. The naming was following the API naming [here](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/1cd28284-3b72-4885-9e31-d1c6d9e26686/71ef7304-a6c9-43b3-a3cd-868b2c236c81/doc/operations/com/vmware/vcenter/ovf/library_item.deploy-operation.html).

ooOOJavaOOoo:
Any chance https://github.com/terraform-providers/terraform-provider-vsphere/issues/291 will get some attention with this working?

PR #1379, opened 2019-02-11
Remove nested progress.Tee usage

dougm:
The call to progress.NewReader already creates a Tee.
This nesting caused prevented progress from exiting when an error occurs during import.vmdk upload.


dougm:
Great, thanks @tdhite

PR #1380, opened 2019-02-11
govc library: use govc/flags for Datastore and ResourcePool

dougm:


govc/library/item/upload.go
```diff
@@ -63,8 +64,8 @@ func (cmd *upload) Description() string {
 	return `List library items.
 
 Examples:
-  govc library.item.upload library_name item_name
-  govc library.item.upload library_name item_name -json | jq .`
+  govc library.item.upload library_name file_name
```
akutz: Hi @dougm,

I believe `item_name` is correct here as you need to know both the name of the library and item to which the file(s) is/are being uploaded.

vapi/library/library.go
```diff
@@ -80,8 +80,11 @@ func (c *Manager) FindLibrary(
 }
 
 // CreateLibrary creates a new library with the given Type, Name,
-// Description, and CategoryID.
+// Description, and CategoryID.  Type defaults to "LOCAL".
```
akutz: Hi @dougm,

I appreciate you catching this and setting a sane default, but I'm not sure about this long-term. The goal is to introduce `SUBSCRIBED` at a later date, and I'm not sure if it makes sense to default to a type here. @markpeek, what do you think? On the other hand, we could let the _defaults to LOCAL_ logic live in the CLI and leave this _as-is_ for now.

govc/library/item/upload.go
```diff
@@ -106,10 +107,9 @@ func (cmd *upload) Run(ctx context.Context, f *flag.FlagSet) error {
 		}
 
 		// Create a library item
-		cmd.item.Name = filename
+		cmd.item.Name = filepath.Base(filename)
```
dougm: @akutz realized `library.Finder` (library.{ls,info}) doesn't work if filename includes `/`'s.  But I think we want to use the basename here in any case.

akutz: Hi @dougm,

That's correct, the finder tokenizes on `/`. If you want to fix this you could just change it so that if `len(parts) > 3` then `parts[2] = strings.Join("", parts[2:])`

vapi/library/library.go
dougm: I agree, I'll remove this and just set in the cli command.

govc/library/item/upload.go
dougm: But this command creates the item for you, from the local file_name.

akutz: Hi @dougm,

Yep. I wish I'd realized that sooner :)

govc/library/item/upload.go
akutz: Hi @dougm,

Please see https://github.com/vmware/govmomi/pull/1381

dougm: Thanks @akutz , just wanted to mention it for now.  We can include the fix with the vcsim/bats tests PR

dougm:
@akutz I removed the LOCAL default, and added 1 more commit, can you review?

dougm:
Thanks @akutz @markpeek

PR #1382, opened 2019-02-12
support customize vm folder in ovf deploy

figo:
use both folder to support customize
vm folder  in content library ovf deploy

cc @dougm

### DIFF
--- a/govc/vcenter/deploy.go
+++ b/govc/vcenter/deploy.go
@@ -31,6 +31,7 @@ import (
 type deploy struct {
 	*flags.DatastoreFlag
 	*flags.ResourcePoolFlag
+	*flags.FolderFlag
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


govc/vcenter/deploy.go
```diff
@@ -127,10 +155,17 @@ func (cmd *deploy) Run(ctx context.Context, f *flag.FlagSet) error {
 			DeploymentSpec: vcenter.DeploymentSpec{
 				Name:               vmName,
 				DefaultDatastoreID: datastoreID,
-				AcceptAllEULA:      true,
+				NetworkMapping: []vcenter.NetworkID{
+					vcenter.NetworkID{
+						Key:   "Network",
```
dougm: Should `Key` value be `network.Reference().Type`, when network is a DVPG for example?

figo:
@dougm i removed the networkMapping for now, need more investigation on that part.

PR #1383, opened 2019-02-13
Fix year in changelog

rHermes:


frapposelli:
Thanks @rHermes !

dougm:
Thanks @rHermes , we have the same bug in the govc/changelog if you want to fix that too :)

rHermes:
I will submit a pull request as soon as I get home! Thanks for the warm welcome!

PR #1393, opened 2019-02-26
Fix a race in NewServer().

benjaminp:
It's important to do all global configuration before starting the server. Otherwise, a request could access settings while they're being configured.


benjaminp:
There's fundamentally no way to find the final certificate of a `httptest` server before it's been started. Hence, I've replaced the `vcsim.server.cert` property with a callback. Another option might be to just manually create a certificate for `localhost` rather than relying on `httptest` to do it.

PR #1395, opened 2019-03-01
Omit namespace tag in generated method body response types

dougm:
To workaround a property collector issue where a client needs to invoke a PropertyCollector method with a namespace other than vim25.
In this case, the body Response type namespace will be that of the request (e.g. "urn:vsan"), which Go fails to decode when the Res
field tag has the urn:vim25 namespace.
The response body type is already given as an input to RoundTrip(), so there's no need to validate the namespace on the client side
and decoding works as expected.
Added custom encoding to the simulator to add namespace to the response body type.

### DIFF
--- a/gen/vim_wsdl.rb
+++ b/gen/vim_wsdl.rb
@@ -766,10 +766,13 @@ class Operation
 
   def dump(io)
     func = ucfirst(name)
+    if namespace != "vim25"
+      tag = "urn:#{namespace} "
+    end
     io.print <<EOS
   type #{func}Body struct{
     Req *#{go_input} `xml:"urn:#{namespace} #{input},omitempty"`
-    Res *#{go_output} `xml:"urn:#{namespace} #{output},omitempty"`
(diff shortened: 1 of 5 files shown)


divyenpatel:
@dougm changes looks good. 

I have a question.

We seem to be omitting namespace for vim25 and pbm methods. Do we need to do the same for ssoadmin, internal and lookup?

Also, I have noticed some panics in the test logs, Although tests seems passed. Is it related to this change?

https://travis-ci.org/vmware/govmomi/jobs/500607970#L3024
https://travis-ci.org/vmware/govmomi/jobs/500607970#L3058

dougm:
Thanks @divyenpatel 
Good catch on the panic.  It was related to this PR, which uncovered a bug in the test (not checking error) and but in the simulator `RefreshStorageInfo` method (not setting Res field).  Updated to include that fix.

We really only need the change for the vim25 methods, but the generator update removed pbm's too.  The other packages were generated originally, but are now maintained by hand, left those as-is for now.  I suppose the generator could be changed to just omit for the vim25 namespace.

PR #1397, opened 2019-03-05
Make PostEvent TaskInfo param optional

dougm:


dougm:
Note: the PostEvent wrapper was added in #214

PR #1402, opened 2019-03-08
vcsim: fork httptest server package

### LINKED ISSUE #1400: vcsim failing with runtime error
Hi @dougm,

vcsim is failing with a runtime exception:

```shell
# vcsim
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xba7ca3]

goroutine 1 [running]:
main.main()
	/go/src/github.com/vmware/govmomi/vcsim/main.go:114 +0xd13
```

The line mentioned by the error is the second line in the following snippet:

```go
	f := flag.Lookup("httptest.serve")
	serve := f.Value.String()
```

It looks like wither `f` or `f.Value` is `nil` when accessed on the second line.

This is with `vcsim` compiled from `v0.20.0` for linux-amd64.

dougm:
As of Go 1.12, the "httptest.serve" flag is not registered unless it is present in os.Args.
There is no other way to set the httptest.serveFlag variable, breaking vcsim's default setting
of port 8989.

The simulator/internal/{server,testcert}.go files are a copy from Go's 1.11 httptest package.

We can remove "httptest.serve" and replace with vcsim's "l" flag, along with removing other
workarounds for the httptest package in a future PR.

Fixes #1400

### DIFF
--- a/simulator/internal/server.go
+++ b/simulator/internal/server.go
@@ -1,0 +1,354 @@
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Implementation of Server
+
+package internal
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"flag"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+	"time"
+)
+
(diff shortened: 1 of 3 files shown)


dougm:
Thanks @akutz , I'll include a README with a follow up change that'll also remove the `-httptest.serve` flag and replace with vcsim's `-l'.  There's a workaround in there for another httptest quirk that I plan to address then too.

Issue #1399, opened 2019-03-07
vcsim: Report back local datastore type as OTHER


exsilium:
Hi, firstly.. thank you for all the hard work and effort that goes into maintaining govmomi and its related subprojects. When implementing solutions on top of vSphere Web Services SDK, `vcsim` is a great tool to test against.

One issue that I encountered with out-of-the-box run is that when doing the inventory of the assets within a DC, the current local Datastore implementation returns a FileSystemType "local". This is an issue, because it's not part of vSphere Web Services API.

As it is a custom made solution within the simulator, i'd still suggest the FileSystemType reported back to be set as "OTHER", which would make it compatible with the SDK API enums.

Source: Enum - HostFileSystemVolumeFileSystemType(vim.host.FileSystemVolume.FileSystemType) - https://code.vmware.com/apis/358/vsphere

```
diff --git a/simulator/host_datastore_system.go b/simulator/host_datastore_system.go
index e6a2dcf..c3f7291 100644
--- a/simulator/host_datastore_system.go
+++ b/simulator/host_datastore_system.go
@@ -98,7 +98,7 @@ func (dss *HostDatastoreSystem) CreateLocalDatastore(c *types.CreateLocalDatasto
                Path: c.Path,
        }
 
-       ds.Summary.Type = "local"
+       ds.Summary.Type = "OTHER"
 
        if err := dss.add(ds); err != nil {
                r.Fault_ = err
```

dougm:
Thanks @exsilium , could you open a pull request with this change?  There's also a generated constant that you can use `types.HostFileSystemVolumeFileSystemTypeOTHER`

dougm:
Fixed by #1405

PR #1405, opened 2019-03-11
Report local Datastore back as type OTHER

exsilium:
Hi. This PR is addressing the issue reported in #1399 

- OTHER is a valid and compatible FileSystemType in vSphere Management Web SDK

Thanks and Cheers! :beers:


exsilium:
CLA Signed

PR #1406, opened 2019-03-12
Add error check for deferred functions

mjtrangoni:
Hi @dougm,

I fixed here some errcheck issues. PTAL, the log types could be others than **Fatal**. See,
```
$ golangci-lint run --skip-dirs '^vim25/xml/*' --disable-all --enable=errcheck
```

### DIFF
--- a/event/processor.go
+++ b/event/processor.go
@@ -115,7 +115,9 @@ func (p *eventProcessor) run(ctx context.Context, tail bool) error {
 		return err
 	}
 
-	defer list.Destroy(context.Background())
+	defer func() {
+		_ = list.Destroy(context.Background())
+	}()
 
 	ref := list.Reference()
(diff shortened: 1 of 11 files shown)


event/processor.go
```diff
@@ -115,7 +116,11 @@ func (p *eventProcessor) run(ctx context.Context, tail bool) error {
 		return err
 	}
 
-	defer list.Destroy(context.Background())
+	defer func() {
+		if err := list.Destroy(context.Background()); err != nil {
+			log.Fatal(err)
```
dougm: We don't want to `log.Fatal` as that will terminate the program.  Using `log.Print` is fine.  May want to add some context to the message `"destroy event view: %s"`

mjtrangoni:
@dougm PTAL

PR #1412, opened 2019-03-14
Fix staticcheck issues value of `XXX` is never used

mjtrangoni:
Hi @dougm,

I fixed here some trivial staticcheck issues related with never used variables.

See,
```
$ golangci-lint run --skip-dirs '^vim25/xml/*' --disable-all --enable=staticcheck
```

### DIFF
--- a/object/virtual_machine.go
+++ b/object/virtual_machine.go
@@ -263,7 +263,7 @@ func (v VirtualMachine) WaitForNetIP(ctx context.Context, v4 bool, device ...str
 	p := property.DefaultCollector(v.c)
 
 	// Wait for all NICs to have a MacAddress, which may not be generated yet.
-	err := property.Wait(ctx, p, v.Reference(), []string{"config.hardware.device"}, func(pc []types.PropertyChange) bool {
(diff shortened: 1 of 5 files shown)


object/virtual_machine.go
```diff
@@ -263,7 +263,7 @@ func (v VirtualMachine) WaitForNetIP(ctx context.Context, v4 bool, device ...str
 	p := property.DefaultCollector(v.c)
 
 	// Wait for all NICs to have a MacAddress, which may not be generated yet.
-	err := property.Wait(ctx, p, v.Reference(), []string{"config.hardware.device"}, func(pc []types.PropertyChange) bool {
+	_ = property.Wait(ctx, p, v.Reference(), []string{"config.hardware.device"}, func(pc []types.PropertyChange) bool {
```
dougm: Looks like this error should be returned rather than ignored.

mjtrangoni: @dougm I did it, and it brokes the test suite. See [here](https://travis-ci.org/vmware/govmomi/jobs/506161754).  Should I TAL at the tests?

dougm: @mjtrangoni do you want to try again (returning err here) and see if CI is ok with the change?

mjtrangoni:
@dougm Thanks!

PR #1445, opened 2019-05-01
govc: refactor library commands

dougm:
Major user-facing changes as we discussed:

- Making use of library.Finder in all commands

- Renaming of some commands

- Removal / Collapsing of some commands

The library command set with this PR is now:
```console
% govc -h | grep library
  library.create
  library.deploy
  library.import
  library.info
  library.ls
  library.rm
  library.session.ls
  library.session.rm
```


lamw:
👍 Ship it!

dougm:
Thanks for the reviews everyone!

PR #1455, opened 2019-05-16
Add PerformanceManager example

amandahla:
Example using simulator to see how to collect cpu.used.summation performance counter from VM.
Reference:
https://pubs.vmware.com/vsphere-6-0/index.jsp?topic=%2Fcom.vmware.wssdk.pg.doc%2FPG_Performance.18.3.html

### DIFF
--- a/examples/perfmanager/main.go
+++ b/examples/perfmanager/main.go
@@ -1,0 +1,104 @@
+/*
+Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This example program shows how to collect performance metrics from virtual machines using govmomi.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/vmware/govmomi/examples"
+	"github.com/vmware/govmomi/performance"
+	"github.com/vmware/govmomi/view"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func main() {
+	examples.Run(func(ctx context.Context, c *vim25.Client) error {
+		// Get virtual machines references
+		m := view.NewManager(c)
+
+		v, err := m.CreateContainerView(ctx, c.ServiceContent.RootFolder, nil, true)
+		if err != nil {
+			return err
+		}
+
+		defer v.Destroy(ctx)
+
+		vmsRefs, err := v.Find(ctx, []string{"VirtualMachine"}, nil)
+		if err != nil {
+			return err
+		}
+
+		// Create a PerfManager
+		perfManager := performance.NewManager(c)
+
+		// Retrieve counters name list
+		counters, err := perfManager.CounterInfoByName(ctx)
+		if err != nil {
+			return err
+		}
+
+		var names []string
+		for name := range counters {
+			names = append(names, name)
+		}
+
+		// Create PerfQuerySpec
+		spec := types.PerfQuerySpec{
+			MaxSample:  1,
+			MetricId:   []types.PerfMetricId{{Instance: "*"}},
+			IntervalId: 300,
+		}
+
+		// Query metrics
+		sample, err := perfManager.SampleByName(ctx, spec, names, vmsRefs)
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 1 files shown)


dougm:
Thanks @amandahla .  Did you look at using the `performance.Manager` helpers?  This would simplify the example.  The govc/metric commands make use of it too.

examples/perfmanager/main.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Fatal(err)
+	}
+
+	s := model.Service.NewServer()
+	defer s.Close()
+
+	// Connect and login to ESX or vCenter
+	// Using simulator instead of real ESX/vCenter
+	//c, err := examples.NewClient(ctx)
+	c, err := govmomi.NewClient(ctx, s.URL, true)
```
dougm: I'd also prefer to stay consistent with the other examples, and use `examples.NewClient` instead.  I do like the idea of defaulting to vcsim if the `url` flag isn't set.

amandahla:
I have used your code from #1458 then change a few things like using finder instead of view to get VMs references.
What do you think?
I have noticed now that I forgot to put a "// Output:" comment like you referenced [here](https://golang.org/pkg/testing/#hdr-Examples). I can do that using data from simulator, I guess.

dougm:
Can you rebase now that #1458 is merged?

The "// Output:" comment is only useful when using `go test` and following the rest of the conventions there.  In the case of gomvomi/examples, `go test` is not required to run and the output can differ if `-url` flag is given.

amandahla:
@dougm Sure, here it is.

examples/perfmanager/main.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/performance"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func main() {
+	examples.Run(func(ctx context.Context, c *vim25.Client) error {
+		// Find virtual machine list
+		f := find.NewFinder(c, true)
```
dougm: `true` should be `false` here, as this usage is just to find the vms and not collect any properties.

examples/perfmanager/main.go
```diff
@@ -0,0 +1,111 @@
...
+
+		f.SetDatacenter(dc)
+
+		vms, err := f.VirtualMachineList(ctx, "*")
+		if err != nil {
+			return err
+		}
+
+		// Convert virtual machines into list of references
+		var vmsRefs []types.ManagedObjectReference
```
dougm: I think using `ContainerView` instead of `Finder` would be better for this example:

* `ContainerView.Find` already returns `[]types.ManagedObjectReference`

* The `kind` param to `Find` could be changed from "VirtualMachine" to another object type (such as "HostSystem") and it looks like your example would still work.  What do you think?

```go
		m := view.NewManager(c)

		v, err := m.CreateContainerView(ctx, c.ServiceContent.RootFolder, nil, true)
		if err != nil {
			return err
		}

		defer v.Destroy(ctx)

		refs, err := v.Find(ctx, []string{"VirtualMachine"}, nil)
		if err != nil {
			return err
		}
```

examples/examples.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright (c) 2017 VMware, Inc. All Rights Reserved.
-
```
dougm: Can you revert these whitespace changes here and the other 2 files below?

examples/perfmanager/main.go
```diff
@@ -0,0 +1,111 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This example program shows how the `view` and `property` packages can
+be used to navigate a vSphere inventory structure using govmomi.
```
dougm: Comment needs an update :)

examples/perfmanager/main.go
```diff
@@ -0,0 +1,111 @@
...
+			for _, v := range metric.Value {
+				counter := counters[v.Name]
+				units := counter.UnitInfo.GetElementDescription().Label
+
+				instance := v.Instance
+				if instance == "" {
+					instance = "-"
+				}
+
+				if v.ValueCSV() != "" {
```
dougm: Better to just call `ValueCSV` once as it does some work.  You could assign to a var first or change the check to:
```go
if len(v.Value) != 0 {
```

examples/perfmanager/main.go
amandahla: Great, much better. I didn't know that. :)

examples/examples.go
amandahla: I reset to the previous commit e just sent now the perfmanager example.

examples/perfmanager/main.go
amandahla: Done.

examples/perfmanager/main.go
amandahla: Changed. :)

PR #1458, opened 2019-05-18
Default to running against vcsim in examples

dougm:
- Add simulator.Model.Do method, generally useful outside of examples

- Add examples.Run method, which defaults to running the example against vcsim

### DIFF
--- a/examples/examples.go
+++ b/examples/examples.go
@@ -20,11 +20,13 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"log"
 	"net/url"
 	"os"
 	"strings"
 
 	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/simulator"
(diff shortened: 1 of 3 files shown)


dougm:
@amandahla here's the idea I was referring to in your PR #1455 , what do you think?
We could change all existing examples to use `examples.Run`.
New examples could also use the `simulator.Model.Do` func directly in the case of create/destroy operations, or examples that depend on inventory structure (such as multiple clusters).

We could also use `simulator.Model.Do` in [godoc style examples](https://golang.org/pkg/testing/#hdr-Examples), though currently there are cyclic imports in some cases (such as with Finder).

amandahla:
@dougm That's great, much easier to understand, thanks. I'll rewrite the one in #1455

dougm:
Thanks for reviewing @amandahla

PR #1465, opened 2019-05-23
Fix pbm field type lookup

dougm:
When fields are encoded with xsi type, vim25/xml typeToString() does not have an option to include namespace prefix.
Workaround this by re-trying the lookup with the namespace prefix.

### DIFF
--- a/pbm/simulator/simulator_test.go
+++ b/pbm/simulator/simulator_test.go
@@ -211,9 +211,19 @@ func TestSimulator(t *testing.T) {
 	t.Logf("VSAN Profile: %q successfully created", vsanProfileID.UniqueId)
 
 	// 6. Verify if profile created exists by issuing a RetrieveContent request.
-	_, err = pc.RetrieveContent(ctx, []types.PbmProfileId{*vsanProfileID})
+	profiles, err := pc.RetrieveContent(ctx, []types.PbmProfileId{*vsanProfileID})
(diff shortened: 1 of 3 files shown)


pbm/simulator/simulator_test.go
```diff
@@ -211,10 +211,20 @@ func TestSimulator(t *testing.T) {
...
-	_, err = pc.RetrieveContent(ctx, []types.PbmProfileId{*vsanProfileID})
+	profiles, err := pc.RetrieveContent(ctx, []types.PbmProfileId{*vsanProfileID})
 	if err != nil {
 		t.Fatal(err)
 	}
+	for _, profile := range profiles {
+		if cap, ok := profile.(*types.PbmCapabilityProfile); ok {
+			_, ok = cap.Constraints.(*types.PbmCapabilitySubProfileConstraints)
+			if !ok {
+				t.Errorf("cap=%T", cap.Constraints)
```
subramanian-neelakantan: Is it sufficient to log an error msg, or should we fail the test here with a t.Fatal? Your call.

dougm: `t.Error` will cause go test to fail, but the test continues to run.  `t.Fatal` aborts the test, I tend to only use it if a `panic()` would ensue.  It's `t.Logf` that just logs and does not fail the test.

PR #1473, opened 2019-05-28
vcsim: use stable UUIDs for inventory objects

### LINKED ISSUE #1470: vcsim should support run-run repeat-ability for the model objects
Vcsim  creates model on the fly i.e, when the simulator is started. This means it generates new model objects with new UUIDs each time the simulator is started. 

For a vcsim client, this is a major issue. Because if it receives new resources each time , it becomes hard to use for simulation purposes especially for scale testing. It confuses the vcsim client as it appears that the resources are changed ( since it generates new UUIDs each time) and we need to safeguard the simulator against restarts/crashes to avoid this. 

Is there a workaround where we get the same result each time we run the simalator with same config params(like vm count,dc count etc)?

dougm:
Fixes #1470

### DIFF
--- a/govc/test/vcsim.bats
+++ b/govc/test/vcsim.bats
@@ -128,6 +128,13 @@ load test_helper
 @test "vcsim vm.create" {
   vcsim_env
 
+  # VM uuids are stable, based on path to .vmx
+  run govc object.collect -s vm/DC0_H0_VM0 config.uuid config.instanceUuid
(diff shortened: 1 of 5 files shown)


dougm:
@suhassankyaan can you review?

suhassankyaan:
@dougm . Changes looks good, tested this and now we have stable UUIDs for inventory objects.
I was thinking to seed the random (random.Seed) but this looks a better approach as it returns the version 5 UUID. 

Thanks for fixing it sooner, appreciate it!

suhassankyaan:
Only after commit, i guess vcsim binary will be refreshed right? 

I tested this change with "go run main.go" command under simulator
But vcsim still generated random uuids.

PR #1479, opened 2019-06-03
Add library export support

dougm:
- Add govc library.export command

- Add library Download Session support

- Rename library 'UpdateSession' to 'Session', same structure for download sessions

- Move vcenter.LocalizableMessage to rest package

### DIFF
--- a/govc/library/export.go
+++ b/govc/library/export.go
@@ -1,0 +1,199 @@
+/*
+Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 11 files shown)


govc/library/export.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (cmd *export) Process(ctx context.Context) error {
+	if err := cmd.ClientFlag.Process(ctx); err != nil {
+		return err
+	}
+	return cmd.OutputFlag.Process(ctx)
+}
+
+func (cmd *export) Run(ctx context.Context, f *flag.FlagSet) error {
+	return cmd.WithRestClient(ctx, func(c *rest.Client) error {
```
markpeek: Possible improvement would be to add a comment about the various modes and calling conventions. Looks like there can be a number of names specified, items/files, and along with a "-" for stdout.

dougm: Good call @markpeek , I've updated the docs.

PR #1480, opened 2019-06-04
Add http source support to library.import

dougm:


govc/test/library.bats
```diff
@@ -121,7 +121,11 @@ load test_helper
   assert_success
   library_id="$output"
 
-  run govc library.import my-content "$GOVC_IMAGES/$TTYLINUX_NAME.ova"
+  # copy ova to datastore so we can test library.import with an http source
+  run govc datastore.upload "$GOVC_IMAGES/$TTYLINUX_NAME.ova" "$TTYLINUX_NAME.ova"
```
akutz: Hi @dougm,

I guess I don't understand this part of the test since the use case is for an OVA imported to the CL from an HTTP source. Why is the datastore involved?

govc/library/import.go
```diff
@@ -90,15 +90,21 @@ func (cmd *item) Run(ctx context.Context, f *flag.FlagSet) error {
 	ext := filepath.Ext(base)
 	mf := strings.Replace(base, ext, ".mf", 1)
 	kind := ""
-	archive := &importx.ArchiveFlag{Archive: &importx.FileArchive{Path: file}}
+	client, err := cmd.Client()
+	if err != nil {
+		return err
+	}
+	opener := importx.Opener{Client: client}
```
akutz: Hi @dougm,

I suppose I don't really understand what the `importx` package is doing, because after looking it sure looks like [this](https://github.com/vmware/govmomi/blob/9034a7e2ec6de3d37718e89a2d7dec6657879e34/govc/importx/archive.go#L186) downloads the OVA locally?

The request was to import the OVA directly from the HTTP source *into* the library similar to [what sk8 does](https://github.com/akutz/simple-k8s-test-env/blob/feature/sk8-app/pkg/provider/vsphere/govmomi/client/library.go).

At no point does the client need to access or fetch anything from the remote OVA. The client simply informs the library to download the OVA into a new/existing library item.

govc/test/library.bats
dougm: This just puts the .ova file behind an http endpoint we can use to test `library.import http://...` with.

govc/library/import.go
dougm: That `Download` method does not save to disk, it just opens a stream.  There is another `DownloadFile` method that does save to disk, but that isn't used in this case.

dougm: btw, if there's a way for content library to pull directly from the http endpoint, I can look into that too.. but this Opener thing was already there for import.{ova,ovf} commands, it just wasn't hooked up to library.import.

dougm: Ah yeah, I see now, AddLibraryItemFileFromURI.. let me see about hooking that up instead of Opener

akutz: Hi @dougm,

Yep. I believe I wrote some helper code around it? Or is that the helper code? Just looked, that **is** the helper code. Please note that that calling that not block the client on waiting for the upload to complete. To do that you must invoke another helper function, [`WaitOnLibraryItemUpdateSession`](https://github.com/vmware/govmomi/blob/d020dc1638f5e609887634e720040394db7a2ac2/vapi/library/library_item_updatesession.go#L96-L116).

akutz: Hi @dougm,

FYI, some downsides to this approach -- it does not provide a quick means of determining whether or not a library item needs to be updated from the remote OVA. Consider this, the library item was already created as a result of importing an OVA. When you import an OVA via `AddLibraryItemFileFromURI`, the content library only keeps the contents of the OVA. Because you're updating the library item from another remote OVA, you don't have a great way to determine whether or not the library item's existing contents match those found in the remote OVA.

To get around this I try to upload a checksum file generated from the OVA into the same content library item as the OVA. That way I can check to see if it exists, and then use that to see if it matches the checksum file associated with the remote OVA (assuming it has one). This doesn't actually checksum the OVA at any point, so it's still not perfect.

dougm: Whoops, I removed `WaitOnLibraryItemUpdateSession` since it wasn't referenced in govmomi or sk8, but I had only searched in vmware/sk8, hadn't seen the branch in akutz - I'll add it back.

Trying `AddLibraryItemFileFromURI` for the first time, looks like no files get added when pointed at a `.ova`, but works as expected with a `.ovf` (with .vmdk(s) in the same dir) - does pulling `.ova` work for you?

Regarding the checksum, if there is manifest `.mf` next to the `.ovf`, the imported files include the checksum info:
```console
% govc library.info -json my-content/ttylinux/*.vmdk | jq .                                                                                                                                                                      
[
  {
    "cached": true,
    "checksum_info": {
      "algorithm": "SHA1",
      "checksum": "ed64564a37366bfe1c93af80e2ead0cbd398c3d3"
    },
    "name": "ttylinux-pc_i486-16.1-disk1.vmdk",
    "size": 29360128,
    "version": "1"
  }
]
```

dougm: Looks like the issue was the `.ova` I was using has a manifest with an entry for the .vmdk, but not the .ovf and the command only checks the returned `error`, not the `State`.

General question: should we have the APIs check for `State == "ERROR"` and return an `error`?
We do this in the `task.Wait` method for example, in the case of `TaskInfoStateError`.

dougm:
@akutz updated with library.import `-pull` option and vcsim support for pull.

govc/library/import.go
```diff
@@ -158,6 +169,15 @@ func (cmd *item) Run(ctx context.Context, f *flag.FlagSet) error {
 			LibraryItemID: cmd.ID,
 		})
 
+		if cmd.pull {
+			_, err = m.AddLibraryItemFileFromURI(ctx, session, filepath.Base(file), file)
+			if err != nil {
+				return err
+			}
+
+			return m.WaitOnLibraryItemUpdateSession(ctx, session, 3*time.Second, nil)
```
akutz: Hi @dougm,

While you *can* use the helper function to wait, I wonder if you want to implement the progress sink here since?

dougm: Thanks for the review @akutz .  Good idea, `Session` does have a `ClientProgress` we can use for this.  Though we also need the `error` wrapper for `State` for the `progress.Report.Error()` impl.  So I'll merge this as-is, and see about including progress here as part of the error related changes we discussed.

akutz: sgtm, thanks again!

PR #1486, opened 2019-06-07
govc: add vm.change hot-add options

### LINKED ISSUE #1485: Changing hotplug for vcpu or memory does not work
`govc version`: 0.20.0
vSphere Client version 6.7.0.20000
Hypervisor: VMware ESXi, 6.7.0, 10302608

I want to disable vcpu and memory hotplug in a vm using:

`govc vm.change -vm myvm -e vcpu.hotadd=false -e mem.hotadd=true` (tried different combinations, i.e. only changing mem, only changing vcpu)

But nothing changes:

`govc object.collect -s /myfolder/myvm config.cpuHotAddEnabled` returns true (also for mem).

I can change the setting using the vSphere GUI, or using Ansible.

This screenshot shows the events in vSphere when using `govc`:
govc

*Note*: "Modified" is blank

This screenshot shows the events in vSphere when using Ansible:
ansible

Here "Modified" shows: "config.cpuHotAddEnabled: true -> false;"

dougm:
Fixes #1485

### DIFF
--- a/govc/vm/change.go
+++ b/govc/vm/change.go
@@ -99,6 +99,8 @@ func (cmd *change) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.Tools = &types.ToolsConfigInfo{}
 	f.Var(flags.NewOptionalBool(&cmd.Tools.SyncTimeWithHost), "sync-time-with-host", "Enable SyncTimeWithHost")
 	f.Var(flags.NewOptionalBool(&cmd.VPMCEnabled), "vpmc-enabled", "Enable CPU performance counters")
+	f.Var(flags.NewOptionalBool(&cmd.MemoryHotAddEnabled), "memory-hot-add-enabled", "Enable memory hot add")
+	f.Var(flags.NewOptionalBool(&cmd.CpuHotAddEnabled), "cpu-hot-add-enabled", "Enable CPU hot add")
 }
 
 func (cmd *change) Description() string {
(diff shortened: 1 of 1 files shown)


dougm:
@henrik-farre I'll go ahead and merge this, can you let me know if it works as expected for you?

henrik-farre:
> @henrik-farre I'll go ahead and merge this, can you let me know if it works as expected for you?

@dougm : it works, I can enable and disable cpu/mem hotplug :)

But I'm not sure the `<nil>` part of `-memory-hot-add-enabled=<nil>` and `-cpu-hot-add-enabled=<nil>` makes much sense when you can use =true/false. The examples also uses the options without true/false:

```
Examples:
...
  # Enable both cpu and memory hotplug on a guest:
  govc vm.change -vm $vm -cpu-hot-add-enabled -memory-hot-add-enabled
...
```
But to disable it I had to use `-cpu-hot-add-enabled=false`

dougm:
Great, glad it works for you, not sure why the `-e` options didn't.

In general, Go's boolean flags when given default to `true`, so these are the same:

* `-cpu-hot-add-enabled`

* `-cpu-hot-add-enabled=true`

* `-cpu-hot-add-enabled true`

But to disable, `false` must be used.  We could change `<nil>` to something else, any suggestions?  We want to indicate there is no default value for these flags.. unlike Go's built in boolean flag which would always have a `true` or `false` default.

https://github.com/vmware/govmomi/blob/ad1ad8c73e048186b147138602ae6fc43b3e4f0f/govc/flags/optional_bool.go#L44

henrik-farre:
Hm, it's confusing for someone who doesn't know go :)

Maybe <nil|false> or is that worse?

dougm:
Understood, `<nil>` is also Go's standard string version of a nil pointer.  I thought about similar, `true|false`, it just isn't consistent with Go's generated help for other flags, where the default value is displayed.  I may just add an explanation to the opening of USAGE.md, as this flag is used in a number of other commands.  Does that sound ok to you?

henrik-farre:
Yes :)

Issue #1352, opened 2019-01-22
credentials via container secrets


rockaut:
In swarm and kubernetes it's possible to securely manage credentials directly and map them to containers as files. It would be great if we can use these mechanisms somehow.

Maybe through an ENV variable like GOVC_SECRETFILE which overrides GOVC_PASSWORD. Also it would be nicer if govc env command DON'T actually shows the sensitive password variables.

dougm:
Not sure about taking on the dependencies to support this in govc.  Can you provide more specific doc/code pointers to the mechanisms you're referring to?

Another option might be integration with desktop mechanisms such as OSX Keychain and/or Linux (GNOME?) Keyring, etc.

dougm:
ok, I had something like NaCl in mind (https://godoc.org/golang.org/x/crypto/nacl/secretbox) regarding the deps.  Didn't realize these types of secrets were exposed as literals within a file.  In that case, rather than add a new flag+env var, how about if `GOVC_PASSWORD` is set to a file we just source the contents?

We can also mask the `govc env` output, but there are scripts that depend on the current behavior.

rockaut:
You mean like ```GOVC_PASSWORD=/var/secrets/password```? Sounds great and should do the trick. But please also for GOVC_USERNAME.

Ad env. I understand that there might be dependency on that currently. Should i extract it from here to another new issue to discuss there? I just wanted to bring up if others see that as problem too.

dougm:
@rockaut haven't forgotten about this, we are discussing similar for vcsim in #1489 
No need to open a separate issue.

PR #1489, opened 2019-06-12
vcsim: add support to override credentials

### LINKED ISSUE #1487: Request for vcsim - validation of credentials
I'm adding support for secure Kubernetes secrets to the vcenter-connector [1], but have no access to vCenter to test it out. I wanted to use vcsim [2], but realised that it doesn't allow overriding of credentials for the simulated server, or validate the password.

I can understand this may be OK for MVP, but I think it would help people making use of vcsim if it A) validated the credentials it prints out in debug B) allowed for them to be provided via a flag.

Thanks again for your work :beers:  

[1] https://github.com/openfaas-incubator/vcenter-connector

[2] https://github.com/vmware/govmomi/tree/master/vcsim

dougm:
Fixes #1487

### DIFF
--- a/govc/flags/client.go
+++ b/govc/flags/client.go
@@ -234,12 +234,22 @@ func (flag *ClientFlag) Register(ctx context.Context, f *flag.FlagSet) {
 
 func (flag *ClientFlag) Process(ctx context.Context) error {
 	return flag.ProcessOnce(func() error {
-		if err := flag.DebugFlag.Process(ctx); err != nil {
+		err := flag.DebugFlag.Process(ctx)
+		if err != nil {
 			return err
 		}
 
 		if flag.url == nil {
 			return errors.New("specify an " + cDescr)
+		}
+
+		flag.username, err = session.Secret(flag.username)
+		if err != nil {
+			return err
+		}
+		flag.password, err = session.Secret(flag.password)
+		if err != nil {
+			return err
 		}
 
 		// Override username if set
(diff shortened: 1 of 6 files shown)


dougm:
@alexellis does this suffice?  We could also add a separate flag, but with this PR you can include username:password with the existing `-l` flag, for example: `vcsim -l nobody:nothing@127.0.0.1:8989`

alexellis:
I might be being a bit pedantic, but I'd separate out the listening from the credentials and also allow them to be taken in via file (for use in Kubernetes).

Ideally:

```
--credentials nobody:nothing
--credentials-from /var/run/secret/basic-auth
```

More verbose might be:

```
--user nobody
--password nothing
--user-from /var/run/secret/basic-auth-user
--password-from  /var/run/secret/basic-auth-password
```

dougm:
I'm ok with adding more flag(s), using `-l` just avoided have to decide how many flags and their names :)

What do you think about `--username` and `--password`, where if the value is a file (checking via Stat), we source from the file?  Else use the literal value of the flags.

Was also thinking about this for govc, with #1352 where the client secrets can be specified via `GOVC_USERNAME` and `GOVC_PASSWORD`.

Either way, would like to keep it consistent between govc and vcsim, I'm open having separate `-from` flags for both the username and password.

alexellis:
Sounds workable, thanks :+1:

dougm:
@alexellis ok updated, thanks for reviewing.

PR #1492, opened 2019-06-14
Support external PSC lookup service

### LINKED ISSUE #1490: govc sso.ls do not support VCSA with external PSC
When use govc sso.ls on VCSA with external PSC, error 404 will return.

dougm:
Fixes #1490


dougm:
@AndrewDi this may work for your case, can you test it out?

AndrewDi:
@dougm Works fine for me.

PR #1498, opened 2019-06-19
Fix tls config in soap.NewServiceClient

dougm:
Use the same transport when creating a service client to preserve options
such as TLSClientConfig.


sachu:
Thank you for fixing this, @dougm.

PR #1501, opened 2019-06-20
tags API: add methods for association of multiple tags/objects

### LINKED ISSUE #1429: Expand Finder to leverage tags to search for Objects
Maybe the functionality is there, but combing through the code, I can't see how it would be done.  Could `finder` be leveraged to find objects based on tags?  Thinking this would be helpful so that you could find cluster/network/datastore resources based on a given set of tags.

dougm:
Add support for multiple tags/objects and '-l' flag to govc tags.attached.ls command

Fixes #1429

### DIFF
--- a/govc/tags/association/ls.go
+++ b/govc/tags/association/ls.go
@@ -21,16 +21,20 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"strings"
+	"text/tabwriter"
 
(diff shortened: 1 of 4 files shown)


dougm:
@donwlewis can you review?

donwlewis:
Nice!  Looks like exactly what I was looking for.  Thanks!

dougm:
Cool, thanks for revewing @donwlewis

PR #1502, opened 2019-06-21
govc: add sso.group commands

### LINKED ISSUE #1483: Add support to create/list sso groups and add sso users to the existing groups
To use govc to manage sso user/groups, we want something like this:
sso.group.ls
sso.group.create
sso.group.rm
sso.group.update  //which will add user to the existing groups
sso.group.permission.set //this may be already available by sso.permission.set, just double check

cc @dougm and thanks a lot!

dougm:
Fixes #1483

### DIFF
--- a/govc/sso/group/rm.go
+++ b/govc/sso/group/rm.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+Copyright (c) 2019 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package user
+package group
 
 import (
 	"context"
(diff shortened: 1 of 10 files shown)


dougm:
@apalcloud @AndrewDi can you review?

apalcloud:
> @apalcloud @AndrewDi can you review?

Thanks @dougm. A quick question, does govc support add/remove AD/Windows domain groups/users?

govc/sso/group/rm.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (cmd *rm) Usage() string {
+	return "NAME"
+}
+
+func (cmd *rm) Description() string {
+	return `Remove SSO group.
+
+Examples:
+  govc sso.user.rm NAME`
```
AndrewDi: rename `govc sso.user.rm NAME` to `govc sso.group.rm NAME`

dougm: Good catch, fixed.

AndrewDi:
LGTM

dougm:
@apalcloud no, the API govc uses doesn't support any admin operations for external directories.

apalcloud:
> @apalcloud no, the API govc uses doesn't support any admin operations for external directories.
Ok thx

PR #1505, opened 2019-06-23
Add various govmomi client examples

### LINKED ISSUE #1443: Any API delete datastores in host or datacentre

### LINKED ISSUE #1451: Grab Events for an Object
I am looking for a way of grabbing the events for an object.  Specifically, we want to be able to get a newly cloned virtual machine, and get the status of the customization spec process (whether it passes or fails).  I am looking at the [Events](https://godoc.org/github.com/vmware/govmomi/event#Manager.Events) function, but it doesn't really seem to work as I expected, and the documentation is a little lacking on how to use it.  Also [the following line](https://github.com/vmware/govmomi/blob/445ddf569fe7c9f37530279041a950b16b7e38f0/event/manager.go#L175) indicates that it is probably not something that is ready to be used.  Is there another way of trying to accomplish this task?  Thanks.

### LINKED ISSUE #1478: Unable to create a folder
I'm trying to clone a VM without working with stuff like FolderFlags etc, and I get "A specified parameter was not correct: spec.folder" error

```
func (vmh *VmHandler) VmClone(vmName, vmNewName string, options *VmOptions) bool {

	vmh.UpdateExsiInfo(false)


	for i, vmInfo := range vmh.vmsInfo {
		if vmInfo.Name == vmName {
			if vmInfo.Summary.Runtime.PowerState == "poweredOn" {
				fmt.Printf("VM [%s] is powered on, please turn it off.\n", vmName)
				return false
			}
			vm := vmh.vms[i]

			folder, err :=vmh.finder.FolderOrDefault(vmh.ctx, "/")
			if err != nil {
				fmt.Printf("Unable to get folder. %s\n", err)
				return false
			}

			cloneSpec := types.VirtualMachineCloneSpec{
				PowerOn:  false,
			}
			task, err := vm.Clone(vmh.ctx, folder, vmName+"_", cloneSpec)
			if err != nil {
				fmt.Printf("Unable to clone VM [%s]. %s\n", vmName, err)
				return false
			}

			err = task.Wait(vmh.ctx)
			if err != nil {
				fmt.Printf("Unable to clone VM [%s] (tsk wait). %s\n", vmName, err)
				return false
			}
		}
	}
	fmt.Printf("Unable to clone VM [%s], incorrect VM name.\n", vmName)
	return false
}
```

### LINKED ISSUE #1495: Get the name of host which VM running.
Hello!

I write yet another tool for monitoring of vSphere for Zabbix with some analytics features and I need to get an information about the host which VM running. If I use 'vm.Summary.Runtime.Host' I getting name of a host like 'host-19'  or 'host-181'. How I can 'to resolve' that to a given name, e.g. 'prod-h.myorg.domain'?

Thank you!

### LINKED ISSUE #1497: Is there a plan to create community samples?
Pyvmomi feels much more consumable because of https://github.com/vmware/pyvmomi-community-samples - is there any plan to create something similar to govmomi (or does it already exist somewhere)

dougm:
Using Go's example style for these tests rather than main functions in the examples/ directory.
These examples will be included in godoc, tested by 'go test' and only run against vcsim.

- Add simulator.Example function to make examples more concise

- Make finder.NewFinder 'all' param optional.  If not specified, also attempts default DC lookup

- Make task.WaitForResult progress param optional

- Add vcsim Datastore.Destroy_Task impl

Fixes #1497
Fixes #1495
Fixes #1478
Fixes #1451
Fixes #1443

### DIFF
--- a/event/example_test.go
+++ b/event/example_test.go
@@ -1,0 +1,56 @@
+/*
+Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package event_test
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/vmware/govmomi/event"
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func ExampleManager_Events() {
+	simulator.Example(func(ctx context.Context, c *vim25.Client) error {
+		m := event.NewManager(c)
+
+		vm, err := find.NewFinder(c).VirtualMachine(ctx, "DC0_H0_VM0")
+		if err != nil {
+			return err
+		}
+
+		objs := []types.ManagedObjectReference{vm.Reference()}
+
+		return m.Events(ctx, objs, 10, false, false, func(ref types.ManagedObjectReference, events []types.BaseEvent) error {
+			event.Sort(events)
+			for _, event := range events {
+				fmt.Printf("%T\n", event)
+			}
+			return nil
+		})
+	})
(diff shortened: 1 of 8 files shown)


dougm:
cc @amandahla

amandahla:
Seems great :-)
Do you want help to create others examples? Maybe a example_test.go in performance package to replace https://github.com/vmware/govmomi/blob/master/examples/perfmanager/main.go ?

dougm:
Thanks @amandahla , yes help creating other examples would be great!

ok to leave the existing examples/ as-is for now.  The godoc examples are meant to be more concise, method specific examples that only run against vcsim (where we control the inventory and can be destructive).  Whereas the examples/ directory can be more complete and run against either vcsim or a real env (where we don't control the inventory and don't want to be destructive).

PR #1506, opened 2019-06-24
govcsim: Set datastore status as normal

Akasurde:
While creating datastore maintenance mode is set to None.
Fix add string value 'normal' as the value of datastore maintenance mode.

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

### DIFF
--- a/simulator/host_datastore_system.go
+++ b/simulator/host_datastore_system.go
@@ -108,6 +108,7 @@ func (dss *HostDatastoreSystem) CreateLocalDatastore(c *types.CreateLocalDatasto
 	}
 
 	ds.Summary.Type = string(types.HostFileSystemVolumeFileSystemTypeOTHER)
+	ds.Summary.MaintenanceMode = string(types.DatastoreSummaryMaintenanceModeStateNormal)
 
 	if err := dss.add(ds); err != nil {
 		r.Fault_ = err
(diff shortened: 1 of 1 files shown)


lebonez:
I don't believe this fixes the issue that you referenced for my pull request. The problem in my request is that when we autoselect a datastore for a vm creation it doesn't check if the datastore is in maintenance mode. I was having problems while doing storage maintenance and vms failing to create if it happened to pick a maintenance mode datastore. I agree with setting to normal when creating a datastore as well. Just figured I would let you know. Also I have no idea why my pull request failed CI.

Akasurde:
@lebonez This is a fix for the testcase. Ansible uses govcsim as a backend for emulating Vmware objects and testing them. Once this is fixed in govcsim, we can re-run the CI and get your PR merged. Your fix in Ansible is correct but we need to fix tests and ultimately govcsim. I hope this explanation helps.

lebonez:
Okay thanks for the explanation.

simulator/host_datastore_system.go
```diff
@@ -108,6 +108,7 @@ func (dss *HostDatastoreSystem) CreateLocalDatastore(c *types.CreateLocalDatasto
 	}
 
 	ds.Summary.Type = string(types.HostFileSystemVolumeFileSystemTypeOTHER)
+	ds.Summary.MaintenanceMode = "normal"
```
dougm: Let's use the generated constant for this value: `types.DatastoreSummaryMaintenanceModeStateNormal`

Akasurde:
@dougm Thanks for the review and merge. @lebonez Thanks for the review.

PR #1507, opened 2019-06-25
Add govmomi performance example

amandahla:
Creates a new performance example in Go's example style as the others in PR #1505.

### DIFF
--- a/performance/example_test.go
+++ b/performance/example_test.go
@@ -1,0 +1,106 @@
+/*
+Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package performance_test
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/vmware/govmomi/performance"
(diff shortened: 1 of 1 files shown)


performance/example_test.go
```diff
@@ -0,0 +1,107 @@
...
+				counter := counters[v.Name]
+				units := counter.UnitInfo.GetElementDescription().Label
+
+				instance := v.Instance
+				if instance == "" {
+					instance = "-"
+				}
+
+				if len(v.Value) != 0 && v.Name == "sys.uptime.latest" {
+					//fmt.Printf("%s\t%s\t%s\t%s\t%s\n",name, instance, v.Name, v.ValueCSV(), units)
```
dougm: Let's leave commented code out of the examples.

performance/example_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"fmt"
+
+	"github.com/vmware/govmomi/performance"
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/view"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func ExamplePerformance_ToMetricSeries() {
```
dougm: CI failed here: https://travis-ci.org/vmware/govmomi/jobs/550296363
Needs to be `PerformanceManager` instead of just `Performance`

performance/example_test.go
amandahla: Done!

performance/example_test.go
amandahla: Changed.

dougm: Whoops, CI failed again, it should actually be `ExampleManager_ToMetricSeries`.  Has to be the same name as the type: https://github.com/vmware/govmomi/blob/2f2047990fde46ace7bb9f096fe2f74e4068c036/performance/manager.go#L36

amandahla: I didn't know that, thanks. Here we go now :)

PR #1515, opened 2019-07-03
Set the InventoryPath of the folder object in DefaultFolder

ykakarap:
The `DefaultFolder` is called in `FolderOrDefault` function which is expected to return a folder object of the default folder if the given path the empty.

The `DefaultFolder` object after creating the folder object needs to set the InventoryPath.

This PR fixes the above-mentioned issue.

### DIFF
--- a/find/finder.go
+++ b/find/finder.go
@@ -917,11 +917,10 @@ func (f *Finder) DefaultFolder(ctx context.Context) (*object.Folder, error) {
 	folder := object.NewFolder(f.client, ref.Reference())
 
 	// Set the InventoryPath of the newly created folder object
-	e, err := f.Element(ctx, folder.Reference())
-	if err != nil {
-		return nil, err
-	}
-	folder.SetInventoryPath(e.Path)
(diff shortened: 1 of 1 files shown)


find/finder.go
```diff
@@ -916,6 +916,13 @@ func (f *Finder) DefaultFolder(ctx context.Context) (*object.Folder, error) {
 	}
 	folder := object.NewFolder(f.client, ref.Reference())
 
+	// Set the InventoryPath of the newly created folder object
+	e, err := f.Element(ctx, folder.Reference())
```
dougm: In this case, we can avoid calling `f.Element`, since we have the Datacenter already and the vmFolder name is always "vm".  I think the following should be ok:
```go
folder.SetInventoryPath(path.Join(f.dc.Datacenter.InventoryPath, "vm"))
```
Can you try that out?

ykakarap: I tried this approach and it works. Pushed a new commit with the change.

dougm:
Thanks @ykakarap !

PR #1516, opened 2019-07-07
VSLM FCD Global Object Manager client for 6.7U2+

dsu-igeek:
Support for FCD Global Object Manager added in 6.7U2.  Implements similar API to existing vslm object_manager but no longer requires clients to track datastores for FCDs.  Also adds searching APIs and per-FCD metadata.

### DIFF
--- a/vslm/client.go
+++ b/vslm/client.go
@@ -1,0 +1,59 @@
+/*
+Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vslm
+
+import (
+	"context"
+
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/soap"
+	vim "github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/govmomi/vslm/methods"
+	"github.com/vmware/govmomi/vslm/types"
+)
+
+const (
+	Namespace = "vslm"
+	Path      = "/vslm/sdk"
+)
+
+var (
+	ServiceInstance = vim.ManagedObjectReference{
+		Type:  "VslmServiceInstance",
+		Value: "ServiceInstance",
+	}
+)
+
+type Client struct {
+	*soap.Client
+
+	ServiceContent types.VslmServiceInstanceContent
+}
+
+func NewClient(ctx context.Context, c *vim25.Client) (*Client, error) {
+	sc := c.Client.NewServiceClient(Path, Namespace)
+	req := types.RetrieveContent{
+		This: ServiceInstance,
+	}
+
+	res, err := methods.RetrieveContent(ctx, sc, &req)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{sc, res.Returnval}, nil
+}
--- a/vslm/global_object_manager.go
+++ b/vslm/global_object_manager.go
@@ -1,0 +1,655 @@
+/*
+Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vslm
+
+import (
+	"context"
+	"time"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
+	vim "github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/govmomi/vslm/methods"
+	"github.com/vmware/govmomi/vslm/types"
+)
+
+type Task struct {
+	c *Client
+	object.Common
+	vim.ManagedObjectReference
+}
+
+func NewTask(client *Client, mref vim.ManagedObjectReference) *Task {
+	m := Task{
+		ManagedObjectReference: mref,
+		c:                      client,
+	}
+	return &m
+}
+
+func (this *Task) QueryResult(ctx context.Context) (vim.AnyType, error) {
+	req := types.VslmQueryTaskResult{
+		This: this.ManagedObjectReference,
+	}
+	res, err := methods.VslmQueryTaskResult(ctx, this.c, &req)
+	if err != nil {
+		return nil, err
+	}
+	return &res.Returnval, nil
+}
+
+func (this *Task) QueryInfo(ctx context.Context) (*types.VslmTaskInfo, error) {
+	req := types.VslmQueryInfo{
+		This: this.ManagedObjectReference,
+	}
+	res, err := methods.VslmQueryInfo(ctx, this.c, &req)
+	if err != nil {
+		return nil, err
+	}
+	return &res.Returnval, nil
+}
+
+func (this *Task) Cancel(ctx context.Context) error {
+	req := types.VslmCancelTask{
+		This: this.ManagedObjectReference,
+	}
+	_, err := methods.VslmCancelTask(ctx, this.c, &req)
+	return err
+}
+
+func (this *Task) Wait(ctx context.Context, timeoutNS time.Duration) (vim.AnyType, error) {
+	return this.WaitNonDefault(ctx, timeoutNS, 10000, true, 10000000)
+}
+
+func (this *Task) WaitNonDefault(ctx context.Context, timeoutNS time.Duration, startIntervalNS time.Duration,
+	exponential bool, maxIntervalNS time.Duration) (vim.AnyType, error) {
+	waitIntervalNS := startIntervalNS
+	startTimeNS := time.Now()
+	for time.Now().Sub(startTimeNS) < timeoutNS {
+		info, err := this.QueryInfo(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if info.State == types.VslmTaskInfoStateQueued || info.State == types.VslmTaskInfoStateRunning {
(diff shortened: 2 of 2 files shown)


examples/vslm/main.go
```diff
@@ -0,0 +1,214 @@
...
+	deleteCmd := flag.NewFlagSet("delete", flag.ExitOnError)
+	deleteID := deleteCmd.String("id", "", "id to delete")
+
+	if len(os.Args) < 2 {
+		fmt.Println("expected subcommand create, list, delete")
+		flag.Usage()
+		os.Exit(1)
+	}
+
+	switch (os.Args[1]) {
```
dougm: Have you looked at the govc `disk.*` commands?  https://github.com/vmware/govmomi/blob/master/govc/USAGE.md#diskcreate

Sources are here: https://github.com/vmware/govmomi/tree/master/govc/disk

And tests: https://github.com/vmware/govmomi/blob/master/govc/test/disk.bats

Thinking we should adapt the disk commands to support using the global catalog, but that can be done later.

go.mod
```diff
@@ -10,4 +10,5 @@ require (
 	github.com/kr/pretty v0.1.0
 	github.com/kr/text v0.1.0 // indirect
 	github.com/vmware/vmw-guestinfo v0.0.0-20170707015358-25eff159a728
+	golang.org/x/tools v0.0.0-20190520220859-26647e34d3c0 // indirect
```
dougm: Doesn't look like this change is needed?

go.sum
```diff
@@ -11,3 +11,10 @@ github.com/rasky/go-xdr v0.0.0-20170217172119-4930550ba2e2 h1:lbe6PJ3nOQAUvpx9P3
 github.com/rasky/go-xdr v0.0.0-20170217172119-4930550ba2e2/go.mod h1:Nfe4efndBz4TibWycNE+lqyJZiMX4ycx+QKV8Ta0f/o=
 github.com/vmware/vmw-guestinfo v0.0.0-20170707015358-25eff159a728 h1:sH9mEk+flyDxiUa5BuPiuhDETMbzrt9A20I2wktMvRQ=
 github.com/vmware/vmw-guestinfo v0.0.0-20170707015358-25eff159a728/go.mod h1:x9oS4Wk2s2u4tS29nEaDLdzvuHdB19CvSGJjPgkZJNk=
+golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
```
dougm: Doesn't look like these changes are needed?

vslm/client.go
```diff
@@ -0,0 +1,61 @@
...
+
+type Client struct {
+	*soap.Client
+
+	ServiceContent types.VslmServiceInstanceContent
+}
+
+func NewClient(ctx context.Context, c *vim25.Client) (*Client, error) {
+	sc := c.Client.NewServiceClient(Path, Namespace)
+	log.Print("sc URL = ", sc.URL())
```
dougm: This log should be removed.

vslm/global_object_manager.go
```diff
@@ -0,0 +1,659 @@
...
+limitations under the License.
+*/
+
+package vslm
+
+import (
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
```
dougm: Let's not add this dependency.  There is an errors package coming in Go 1.13: https://crawshaw.io/blog/xerrors

This import is also causing CI to fail, since the package isn't vendored.

vslm/global_object_manager.go
```diff
@@ -0,0 +1,659 @@
...
+func (this *Task) Wait(ctx context.Context, timeoutNS time.Duration) (vim.AnyType, error) {
+	return this.WaitNonDefault(ctx, timeoutNS, 10000, true, 10000000)
+}
+
+func (this *Task) WaitNonDefault(ctx context.Context, timeoutNS time.Duration, startIntervalNS time.Duration,
+	exponential bool, maxIntervalNS time.Duration) (vim.AnyType, error) {
+	waitIntervalNS := startIntervalNS
+	startTimeNS := time.Now()
+	for time.Now().Sub(startTimeNS) < timeoutNS {
+		info, err := this.QueryInfo(ctx)
```
dougm: Is it not possible to use the property collector with a `vslm.Task`?

vslm/global_object_manager.go
```diff
@@ -0,0 +1,659 @@
...
+}
+
+func (this *GlobalObjectManager) SetControlFlags(ct context.Context, controlFlags []string) error {
+	req := types.VslmSetVStorageObjectControlFlags{
+		This:         this.Reference(),
+		ControlFlags: controlFlags,
+	}
+
+	_, err := methods.VslmSetVStorageObjectControlFlags(ct, this.c, &req)
+	if err != nil {
```
dougm: No need for this check, can just `return err` here.

vslm/global_object_manager.go
```diff
@@ -0,0 +1,659 @@
...
+	return nil
+}
+
+func (this *GlobalObjectManager) ClearControlFlags(ct context.Context) error {
+	req := types.VslmClearVStorageObjectControlFlags{
+		This: this.Reference(),
+	}
+
+	_, err := methods.VslmClearVStorageObjectControlFlags(ct, this.c, &req)
+	if err != nil {
```
dougm: No need for this check, can just `return err` here.

vslm/global_object_manager.go
```diff
@@ -0,0 +1,659 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (this *GlobalObjectManager) AttachTag(ctx context.Context, id vim.ID, category string, tag string) error {
```
dougm: `ObjectManager` uses `types.VslmTagEntry` for category+tag, since `ListAttachedTags` returns that type:

https://github.com/vmware/govmomi/blob/35d0b7d3d2f6ddb02af246d2e0285e5c23aa9850/vslm/object_manager.go#L360

`ObjectManager` also uses `id string` everywhere, rather than `vim.ID`.

I'd be open to changing `ObjectManager`'s method signatures if you prefer, but would like to keep them consistent.

vslm/global_object_manager.go
```diff
@@ -0,0 +1,659 @@
...
+
+	res, err := methods.VslmQueryChangedDiskAreas(ctx, this.c, &req)
+	if err != nil {
+		return nil, err
+	}
+
+	return &res.Returnval, nil
+}
+
+func (this *GlobalObjectManager) queryGlobalCatalogSyncStatus(ct context.Context) ([]types.VslmDatastoreSyncStatus, error) {
```
dougm: Are these 2 methods intentionally not exported or should that be `Q`?

vslm/global_object_manager.go
```diff
@@ -0,0 +1,659 @@
...
+
+	res, err := methods.VslmRetrieveVStorageObjectMetadataValue(ct, this.c, &req)
+	if err != nil {
+		return "", err
+	}
+
+	return res.Returnval, nil
+}
+
+func (this *GlobalObjectManager) retrieveObjects(ct context.Context, ids []vim.ID) ([]types.VslmVsoVStorageObjectResult,
```
dougm: Should this method be exported along with the one below?

go.mod
dsu-igeek: Go just kind of stuffed it in there so I assumed it was needed.

vslm/client.go
dsu-igeek: Sure

vslm/global_object_manager.go
dsu-igeek: Forgot those

vslm/global_object_manager.go
dsu-igeek: Yup

vslm/global_object_manager.go
dsu-igeek: Sure - not real happy with Go error handling, so let's add this back in when we can.

vslm/global_object_manager.go
dsu-igeek: I'd prefer to use the ID object rather than a string.  For example, ListObjectsForSpec returns a VslmVsoVStorageObjectQuerySpec which has an array of IDs.  In order to retrieve an object, you'd have to convert the ID to a string and we'll then convert back to an ID.  Can we change those signatures in ObjectManager?  How much stuff will we break?

examples/vslm/main.go
dsu-igeek: Yeah, we can do that.  What's our backwards compatibility story?  The global object manager is only available from 6.7U2+

dougm: In this case, when the `-ds` flag is given, the govc disk commands invoke the `ObjectManager` methods..  If no `-ds` flag is given, they invoke the `GlobalObjectManager` methods.  In the latter case, when run against VC < 6.7U2, the user just gets an error (e.g. MethodNotFound).  But again, we can look at doing this later.

go.mod
dougm: go.{mod,sum} should only change if `go mod` commands are used to update dependencies, but possible your IDE had done that automatically?

vslm/global_object_manager.go
dougm: :+1: for now let's just keep it consistent.

vslm/global_object_manager.go
dougm: I'd be ok with that, it'll break a bunch of the govc disk commands, but those are easy to fix.  It'll also break some of the K8s CSI driver for vSphere, but that won't be much trouble to fix either.  I don't know of anything else using the current `ObjectManager` APIs and they haven't been around too long.

vslm/client.go
dsu-igeek: Reverted this as well.

vslm/global_object_manager.go
dsu-igeek: Fixed

vslm/global_object_manager.go
dsu-igeek: I'll update ObjectManager and our examples and make a separate PR

vslm/global_object_manager.go
dsu-igeek: Should be fixed now

go.mod
codenrhoden: they will also change as needed when running things like `go test`,  and `go build`...

From point 3 in the [wiki](https://github.com/golang/go/wiki/Modules#how-to-define-a-module):
 > go build will automatically add missing or unconverted dependencies as needed to satisfy imports for this particular build invocation

Might worth running `go mod tidy` if you are not already doing so to capture everything.

examples/vslm/main.go
dougm: Realizing another issue with this example, CI fails because it runs all the examples against vcsim: https://github.com/vmware/govmomi/blob/66eb1fa1fa350800914d1a699fb66bea14099a28/govc/test/vcsim.bats#L20

But, vcsim doesn't support these new APIs yet.  For now we could exclude vslm from vcsim.bats or just not include the example for now.  I plan to add the vcsim support once your PR is merged.

PR #1582, opened 2019-09-01
Content library: support library ID in Finder

### LINKED ISSUE #1581: I could create content library with same name but then can't upload image to it
# Problem
vSphere allows content library with same name backed by same datastore, but then I tried to import OVA to datastore will run into issue that it complained 
`govc: "my-content-library" matches 2 items`.

Just wonder if vSphere supports libraries with same name, how to we prevent problem like this from happening

# Step to reproduce
1. create two libraries with same name
```
$ govc library.create my-content-library
833d5efb-a1d5-4771-8136-4d1ee64d2cec
$ govc library.create my-content-library
a7c87ca3-dc1e-4da0-8730-99a09c9771eb
```

```
$ govc library.info
Name:           my-content-library
  ID:           833d5efb-a1d5-4771-8136-4d1ee64d2cec
  Path:         /my-content-library
  Description:
  Version:      2
  StorageBackings:
    DatastoreID:  nfs-1
    Type:         DATASTORE
Name:           my-content-library
  ID:           a7c87ca3-dc1e-4da0-8730-99a09c9771eb
  Path:         /my-content-library
  Description:
  Version:      2
  StorageBackings:
    DatastoreID:  nfs-1
    Type:         DATASTORE
```
2. upload image
```
govc library.import my-content-library https://cloud-images.ubuntu.com/releases/xenial/release/ubuntu-16.04-server-cloudimg-amd64.ova
govc: "my-content-library" matches 2 items
```

dougm:
vCenter supports libraries with the same name, so we need to support
use of library ID instead of name for methods that require a single library.

Fixes #1581

### DIFF
--- a/govc/library/rm.go
+++ b/govc/library/rm.go
@@ -51,6 +51,7 @@ func (cmd *rm) Description() string {
 
 Examples:
   govc library.rm /library_name
+  govc library.rm library_id # Use library id if multiple libraries have the same name
   govc library.rm /library_name/item_name`
 }
 
@@ -64,7 +65,7 @@ func (cmd *rm) Run(ctx context.Context, f *flag.FlagSet) error {
 		}
 
 		if len(res) != 1 {
-			return fmt.Errorf("%q matches %d items", f.Arg(0), len(res))
+			return ErrMultiMatch{Type: "library", Key: "name", Val: f.Arg(0), Count: len(res)}
 		}
 
 		switch t := res[0].GetResult().(type) {
--- a/vapi/library/finder/finder.go
+++ b/vapi/library/finder/finder.go
@@ -206,6 +206,12 @@ func (f *Finder) findLibraries(
 			}
 			result = append(result, findResult{result: *lib})
 		}
+		if len(result) == 0 {
+			lib, err := f.M.GetLibraryByID(ctx, token)
+			if err == nil {
+				result = append(result, findResult{result: *lib})
+			}
+		}
 		return result, nil
 	}
 
--- a/govc/library/deploy.go
+++ b/govc/library/deploy.go
@@ -104,7 +104,7 @@ func (cmd *deploy) Run(ctx context.Context, f *flag.FlagSet) error {
 			return err
 		}
 		if len(res) != 1 {
-			return fmt.Errorf("%q matches %d items", path, len(res))
+			return ErrMultiMatch{Type: "library", Key: "name", Val: f.Arg(0), Count: len(res)}
 		}
 		item, ok := res[0].GetResult().(library.Item)
 		if !ok {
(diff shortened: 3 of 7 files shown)


govc/library/rm.go
```diff
@@ -28,6 +28,8 @@ import (
 	"github.com/vmware/govmomi/vapi/rest"
 )
 
+var fmtMultiMatch = "%q matches %d items, library id must be specified"
```
akutz: Hi @dougm,

This should be a constant.

akutz: Hi @dougm,

Also, maybe consider the following - ([go playground link](https://play.golang.org/p/qt3Uc-No6c2))

```golang
package main

import (
	"fmt"
	"log"
)

// ErrMultiMatch is an error returned when a query returns more than one result.
type ErrMultiMatch struct {
	// Type is the type of object being queried.
	Type string

	// TokenKey is the key used to perform the query.
	TokenKey string

	// TokenVal is the value used to perform the query.
	TokenVal string

	// Count is the number of objects returned.
	Count int
}

// Error returns the error string.
func (e ErrMultiMatch) Error() string {
	return e.String()
}

// String returns the same value as Error().
func (e ErrMultiMatch) String() string {
	return fmt.Sprintf("%q=%q matches %d items, %q id must be specified", e.TokenKey, e.TokenVal, e.Count, e.Type)
}

func main() {
	err := ErrMultiMatch{
		Type:     "library",
		TokenKey: "name",
		TokenVal: "MyContentLibrary",
		Count:    3,
	}
	log.Fatal(err)
}
```

Prints the following:

```shell
2009/11/10 23:00:00 "name"="MyContentLibrary" matches 3 items, "library" id must be specified
```

vapi/library/finder/finder.go
```diff
@@ -206,6 +206,12 @@ func (f *Finder) findLibraries(
 			}
 			result = append(result, findResult{result: *lib})
 		}
+		if len(result) == 0 {
+			lib, err := f.M.GetLibraryByID(ctx, token)
```
akutz: Hi @dougm,

This can be collapsed to reduce scope. Since the scope of `lib` and `err` are already bound by `if len(result) == 0`, then I understand if you wish to keep things as is to prevent a longer line on 210. Up to you.

govc/library/rm.go
markpeek: This could be const instead of var.

dougm: Thanks @akutz, I like your error type and have integrated into this PR.

vapi/library/finder/finder.go
dougm: Hi @akutz, you mean `if lib, err := f.M.GetLibraryByID(ctx, token); err == nil {` ?

I personally tend to avoid that as it can be harder to read at times.  Just always having `if err != nil {` following ends up being more consistent.  The for loop body above is that way too.

dougm:
Thanks for the reviews @akutz , @markpeek , @lubronzhan

PR #1586, opened 2019-09-04
vapi: return info with current session query

dougm:
Rename Client.Get to Client.Session and return the session info.
A nil Session is returned when not authenticated.  This allows callers
to disambiguate between http.StatusUnauthorized and other errors,
as the soap session.Manager.UserSession method does.

### DIFF
--- a/vapi/rest/client.go
+++ b/vapi/rest/client.go
@@ -25,6 +25,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"time"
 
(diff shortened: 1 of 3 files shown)


dougm:
@MarcusTan wanted to make this change before we tag the next release, can you review?

The soap session.Manager.UserSession method for reference: https://github.com/vmware/govmomi/blob/5b4c2838b4006095fc766b22cdc4de4759d721c4/session/manager.go#L177-L179

PR #1623, opened 2019-09-23
vcsim: add record/playback functionality

dougm:
Remove New* funcs as Model.Load will construct via reflection
Move any remaining New* logic to init() methods call regardless of
using Model.Load or Model.Create

### DIFF
--- a/govc/object/save.go
+++ b/govc/object/save.go
@@ -295,7 +295,7 @@ func (cmd *save) Run(ctx context.Context, f *flag.FlagSet) error {
 	if cmd.verbose {
 		s = ""
 	}
-	fmt.Printf("Saved %d total objects%s:\n", cmd.n, s)
+	fmt.Printf("Saved %d total objects to %q%s:\n", cmd.n, cmd.dir, s)
 	for i := range summary {
 		fmt.Println(summary[i])
 	}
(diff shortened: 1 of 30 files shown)


govc/object/save.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+		summary = append(summary, fmt.Sprintf("%s: %d", k, v))
+	}
+	sort.Strings(summary)
+
+	s := ", including"
+	if cmd.verbose {
+		s = ""
+	}
+	fmt.Printf("Saved %d total objects%s:\n", cmd.n, s)
```
frapposelli: Could be useful to print out the dir name.
```suggestion
	fmt.Printf("Saved %d total objects to %q%s:\n", cmd.n, cmd.dir, s)
```

vcsim/README.md
```diff
@@ -67,7 +67,33 @@ Example using ESX mode:
...
 /ha-datacenter/network/VM Network
+```
 
+## Record & Playback
+
+Example to record the inventory of a live vCenter or ESX and playback with vcsim:
+
+```console
+% govc object.save -u user:pass@my-vcenter -d my-vcenter
+Saved 164 total objects, including:
```
frapposelli: This goes with the other suggestion above
```suggestion
Saved 164 total objects to "my-vcenter", including:
```

PR #1628, opened 2019-09-24
Support HoK tokens with Interactive Users

### LINKED ISSUE #1622: Bug in method sts.Signer.Sign() prevents issuing HoK token for interactive user
STS supports issuing HoK token for interactive users.  The caller is supposed to generate ephemeral pub/priv key pair, and request STS to issue a token by providing username/password, as well as the private key and a self-signed cert.

However, that will not work in govmomi, because of this logic in sts.Signer.Sign() method:

	case requestToken:
		if hasKey {
			// We need c14n for all requests, as its digest is included in the signature and must match on the server side.
			// We need the body in original form when using an ActAs or RenewTarget token, where the token and its signature are embedded in the body.
			req := x.RequestSecurityToken()
			c14n = req.C14N()
			body = req.String()
			id := newID()

			info.SecurityTokenReference = &internal.SecurityTokenReference{
				Reference: &internal.SecurityReference{
					URI:       "#" + id,
					ValueType: "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509v3",
				},
			}

			header.BinarySecurityToken = &internal.BinarySecurityToken{
				EncodingType: "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-soap-message-security-1.0#Base64Binary",
				ValueType:    "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509v3",
				ID:           id,
				Value:        base64.StdEncoding.EncodeToString(s.Certificate.Certificate[0]),
			}
		} else {
			header.UsernameToken = &internal.UsernameToken{
				Username: s.user.Username(),
			}
			header.UsernameToken.Password, _ = s.user.Password()
		}

The "else" clause is wrong.  In case of HoK token request for an interactive user, both private key *and* username/password are present in the request.  Removing "else" clause and replacing it with the following fixes the problem:

	case requestToken:
		if hasKey {
			// We need c14n for all requests, as its digest is included in the signature and must match on the server side.
			// We need the body in original form when using an ActAs or RenewTarget token, where the token and its signature are embedded in the body.
			req := x.RequestSecurityToken()
			c14n = req.C14N()
			body = req.String()
			id := newID()

			info.SecurityTokenReference = &internal.SecurityTokenReference{
				Reference: &internal.SecurityReference{
					URI:       "#" + id,
					ValueType: "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509v3",
				},
			}

			header.BinarySecurityToken = &internal.BinarySecurityToken{
				EncodingType: "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-soap-message-security-1.0#Base64Binary",
				ValueType:    "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509v3",
				ID:           id,
				Value:        base64.StdEncoding.EncodeToString(s.Certificate.Certificate[0]),
			}
		}
		// When requesting HoK token for interactive user, request will have both priv. key and username/password.
		if s.user != nil {
			header.UsernameToken = &internal.UsernameToken{
				Username: s.user.Username(),
			}
			header.UsernameToken.Password, _ = s.user.Password()
		}

dougm:
Prior to the change, only Solution Users could use HoK tokens.

Fixes #1622

### DIFF
--- a/sts/signer.go
+++ b/sts/signer.go
@@ -141,7 +141,9 @@ func (s *Signer) Sign(env soap.Envelope) ([]byte, error) {
 				ID:           id,
 				Value:        base64.StdEncoding.EncodeToString(s.Certificate.Certificate[0]),
 			}
-		} else {
+		}
+		// When requesting HoK token for interactive user, request will have both priv. key and username/password.
+		if s.user != nil {
 			header.UsernameToken = &internal.UsernameToken{
 				Username: s.user.Username(),
 			}


dougm:
Thanks @llivshin

PR #1636, opened 2019-09-26
The tree command

akutz:
This patch adds a "tree" command to govc that behaves like the Linux tree command. "govc tree PATH" prints a tree-like structure of the inventory.

This first-pass is incredibly slow and needs to be optimized. Here's some example output while listing the networks:

```shell
$ ./govc tree /SDDC-Datacenter/network
/SDDC-Datacenter/network
├── VMC Networks
│   ├── SI_LogicalSwitch_40584124-5a85-479b-a5ef-07806a2d1b73
│   ├── SI_LogicalSwitch_4e59f062-7af8-4cbc-ac1a-3a9a40511d16
│   ├── SI_LogicalSwitch_beb34a4f-50c5-4f39-b953-a9861a919747
│   ├── SI_LogicalSwitch_de819edf-141f-498c-bf9a-b706d63352b7
│   ├── VMOTION
│   ├── VSAN
│   ├── cgw-dhcp-ls
│   ├── cross-vpc-ls
│   ├── direct-connect-ls
│   ├── mgmt-app-network
│   ├── mgmt-ls
│   ├── mgw-dhcp-ls
│   ├── nsx-dlink0-vls
│   ├── o-vmk0-ls
│   ├── public-ls
│   ├── trunk-ls
│   └── vmk4-ls
├── sddc-cgw-network-1
├── sddc-cgw-network-2
├── sddc-cgw-network-3
├── sddc-cgw-network-4
├── sddc-cgw-network-5
├── sddc-cgw-network-6
├── sddc-cgw-network-lvs-1
└── sddc-cgw-network-sk8-1
```

Many thanks to @dougm!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ replace github.com/davecgh/go-xdr => github.com/rasky/go-xdr v0.0.0-201702171721
 replace github.com/kr/pretty v0.1.0 => github.com/dougm/pretty v0.0.0-20171025230240-2ee9d7453c02
 
 require (
+	github.com/a8m/tree v0.0.0-20210115125333-10a5fd5b637d
 	github.com/davecgh/go-xdr v0.0.0-20161123171359-e6a2ba005892
 	github.com/google/uuid v0.0.0-20170306145142-6a5e28554805
 	github.com/kr/pretty v0.1.0
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,7 @@
+github.com/a8m/tree v0.0.0-20181222104329-6a0b80129de4 h1:mK1/QgFPU4osbhjJ26B1w738kjQHaGJcon8uCLMS8fk=
+github.com/a8m/tree v0.0.0-20181222104329-6a0b80129de4/go.mod h1:FSdwKX97koS5efgm8WevNf7XS3PqtyFkKDDXrz778cg=
+github.com/a8m/tree v0.0.0-20210115125333-10a5fd5b637d h1:4E8RufAN3UQ/weB6AnQ4y5miZCO0Yco8ZdGId41WuQs=
+github.com/a8m/tree v0.0.0-20210115125333-10a5fd5b637d/go.mod h1:FSdwKX97koS5efgm8WevNf7XS3PqtyFkKDDXrz778cg=
 github.com/davecgh/go-xdr v0.0.0-20161123171359-e6a2ba005892 h1:qg9VbHo1TlL0KDM0vYvBG9EY0X0Yku5WYIPoFWt8f6o=
 github.com/davecgh/go-xdr v0.0.0-20161123171359-e6a2ba005892/go.mod h1:CTDl0pzVzE5DEzZhPfvhY/9sPFMQIxaJ9VAMs9AagrE=
 github.com/dougm/pretty v0.0.0-20171025230240-2ee9d7453c02 h1:tR3jsKPiO/mb6ntzk/dJlHZtm37CPfVp1C9KIo534+4=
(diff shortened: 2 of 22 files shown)


dougm:
Thanks @akutz , just trying this out, looks awesome!  I will follow up with a review and may experiment with using ContainerView instead of Finder, as I suspect the latter is to blame for when it seems slow.

govc/tree/command.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/list"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+type tree struct {
+	*flags.DatacenterFlag
```
dougm: Should use `ClientFlag` instead, since tree can work against multiple datacenters, don't want to require that `-dc` is set in that case.

govc/tree/command.go
```diff
@@ -0,0 +1,183 @@
...
+
+	ToRef bool
```
dougm: This flag isn't used currently, probably not worth implementing for tree.

govc/tree/command.go
```diff
@@ -0,0 +1,183 @@
...
+	DeRef bool
```
dougm: Could also remove this and if folks want to use moref as input:
```console
%  govc ls -L Datacenter:datacenter-2 | xargs govc tree
```

govc/tree/command.go
```diff
@@ -0,0 +1,183 @@
...
+	treeOpts := &gotree.Options{
+		Fs:      vfs,
+		OutFile: os.Stdout,
+	}
+
+	var nd, nf int
+	for i := range args {
+		inf := gotree.New(args[i])
+		d, f := inf.Visit(treeOpts)
+		nd, nf = nd+d, nf+f
```
dougm: Doesn't look like these are used?

govc/tree/command.go
```diff
@@ -0,0 +1,183 @@
...
+				listElement = &result[0]
+			}
+		}
+	}
+	if listElement == nil {
+		return nil, os.ErrNotExist
+	}
+
+	var mode os.FileMode
+	switch listElement.Object.Reference().Type {
```
dougm: Could save this suggestion for a follow up, but how about by default tree only recurses `Folder` and `Datacenter` by default.  For the others types, treat them as "symlinks" and follow tree(1)'s behavior:
> By default, when a symbolic link is encountered, the path that the symbolic link refers to is printed after the name of the link in the format:
>
>     name -> real-path
>
> If the `-l' option is given and the symbolic link refers to an actual directory, then tree will follow the path of the symbolic link as if it were a real directory.

I wish we'd done that for the `govc ls` command too.

dougm:
@akutz , sorry it took me so long to get to this.. I've made the changes I suggested in this PR.  Switching from Finder to ContainerView reduces the number of round trips quite a bit.  I also submitted a PR to the tree package so we can customize the colors (and added a `-C` flag to enable).  If you have a chance to take a look:

https://github.com/dougm/govmomi/commit/6f6d26393f5b92e23ffedb84c86cf998c9d0760d

I'd like to squash the result into a single commit and keep you as the author.  I can do that as a separate PR or you can push to yours if you're up for it.  Open to any changes/suggestions too of course.  Thanks!

akutz:
Hi @dougm,

That all makes sense to me; thank you for reaching out and checking!

dougm:
@akutz squashed and GitHub let me push to your branch, will go ahead and merge in the next day or so. Thanks again!

Issue #1621, opened 2019-09-23
Question: how to send a magic sysrq through vm.keystrokes


freedge:
Hi. Can you document how one could send a [magic sysrq](https://en.wikipedia.org/wiki/Magic_SysRq_key) combination to a VM running a Linux system (combination is done by holding an Alt key, Print Screen, and some key  - eg: b for reboot).

It would be a very convenient usage of govc vm.keystrokes in case of a crash.

I could not manage to send such a combination (vSphere 6.5 and ESX host 6.5). In the event this would be possible, it would be good to make it part of the documentation and usage screen.

dougm:
Have you tried `0x46` ?
I haven't, just skimmed through https://gist.github.com/MightyPork/6da26e382a7ad91b5496ee55fdc73db2#file-usb_hid_keys-h-L110

freedge:
yes I tried.
The issue is that we have access to [modifiers](https://www.vmware.com/support/orchestrator/doc/vro-vsphere60-api/html/VcUsbScanCodeSpecModifierType.html) to hold left or right Alt key together with the key we're pushing, but we have no way to hold the print screen. It does not suffice to just hold Alt key and press the print screen key, you need to hold Alt key, hold the print screen key, then press a key.

Ideally, we should have a new modifier for the print screen key.

freedge:
btw the documentation mentions:

```
govc vm.keystrokes -vm $vm -c 0x4c -la true -lc true
```

as a mean to send ctrl+alt+sup, but this really should be
```
govc vm.keystrokes -vm $vm -c=0x4c -la=true -lc=true
```

In the first case, govc does not parse the flag correctly and left control is set to false.

freedge:
Finally, I coded this

https://github.com/freedge/govmomi/commit/ab1da5704829db6baf8ad77d0a16c134066aa1cd

in order to be able to send multiple keys through the -c flag. However when I run showkey -k on the console, we see that the Alt key is held, but released between each pressure on a new key, so it still does not help.

My OS should be able to send a Magic SysRq if one keeps the Alt button held, then pressing SysRq then a key one after the other.

So in the end, it's still a limitation of the vmware API

freedge:
so technically, I have my answer: it's just not possible.

We should open a service request to support that I guess

PR #1638, opened 2019-09-27
Fix description of vm.keystrokes

freedge:
as mentionned in https://github.com/vmware/govmomi/issues/1621

```-la true -rc true```  will silently parse la as true and rc as false.

Not sure what is wrong with the flag parsing, but fixing at least the example in the description.


dougm:
Thanks @freedge 

Note that you can also use `-la` without any argument and it'll behave the same as `-la=true`.  But the arg is require to disable a boolean flag `-la=false`

We use Go's package: https://godoc.org/flag

PR #1640, opened 2019-09-30
add GetInventoryPath to NetworkReference interface

luthermonson:
When using NetworkList from Finder a list of interfaces is returned meaning you'd need to make a similar looking switch to type caste back to the proper type which is a bit annoying if you're just trying to get InventoryPath. Adding a helper method to NetworkReference to grab and return it.

### DIFF
--- a/object/distributed_virtual_portgroup.go
+++ b/object/distributed_virtual_portgroup.go
@@ -36,6 +36,10 @@ func NewDistributedVirtualPortgroup(c *vim25.Client, ref types.ManagedObjectRefe
 	}
 }
 
+func (p DistributedVirtualPortgroup) GetInventoryPath() string {
+	return p.InventoryPath
+}
+
 // EthernetCardBackingInfo returns the VirtualDeviceBackingInfo for this DistributedVirtualPortgroup
 func (p DistributedVirtualPortgroup) EthernetCardBackingInfo(ctx context.Context) (types.BaseVirtualDeviceBackingInfo, error) {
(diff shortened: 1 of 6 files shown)


luthermonson:
if this is merged you can replace something like this
```
...
	networks, err := finder.NetworkList(ctx, "*")
	if err != nil {
		return nil, err
	}

	var data []string
	for _, net := range networks {
		data = append(data, getNetworkInventoryPath(net))
	}
... 

func getNetworkInventoryPath(nr object.NetworkReference) string {
	var r string
	switch nr.Reference().Type {
	case "Network":
		r = nr.(*object.Network).InventoryPath
	case "OpaqueNetwork":
		r = nr.(*object.OpaqueNetwork).InventoryPath
	case "DistributedVirtualPortgroup":
		r = nr.(*object.DistributedVirtualPortgroup).InventoryPath
	case "DistributedVirtualSwitch":
		r = nr.(*object.DistributedVirtualSwitch).InventoryPath
	case "VmwareDistributedVirtualSwitch":
		r = nr.(*object.VmwareDistributedVirtualSwitch).InventoryPath
	}

	return r
}

```

to something a bit cleaner like this

```
	networks, err := finder.NetworkList(ctx, "*")
	if err != nil {
		return nil, err
	}

	var data []string
	for _, net := range networks {
		data = append(data,net.GetInventoryPath())
	}

```

dougm:
Thanks again @luthermonson

PR #1644, opened 2019-10-02
vslm: Fix to throw errors on tasks that are completed with error state

lintongj:
In the current code base, govmomi/vslm does not throw any errors even if API tasks are completed with error state, which would be confusing to govmomi client. Make the fix to throw errors as expected.

### DIFF
--- a/vslm/global_object_manager.go
+++ b/vslm/global_object_manager.go
@@ -18,6 +18,7 @@ package vslm
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/vmware/govmomi/object"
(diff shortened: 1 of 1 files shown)


lintongj:
Thanks @dougm. Do I still need to sign the CLA? It is still in the status of "Some checks haven't completed yet" due to the pending CLA, even though the cla-not-required label has been added.

dougm:
@lintongj you don't need to sign the CLA since you're @vmware.com , not sure why the status is that way.  I'll go ahead and merge, thanks again.

lintongj:
Thanks @dougm.

PR #1648, opened 2019-10-03
Debug Update to Add a Password Scrubber and stdout style LogType

luthermonson:
Most important part of this PR is the Scrub method added to the debug output to remove your password from the logs. To do this I just wrote a wrapper CloseWrite for FileType and to debug easier I wrote a LogType so decided to include it in the PR.

LogType uses go's log package and fakes a write/close to pass to the TeeReader and instead run a `log.Print()` on the contents which are normally written to file. This will show debug in your console while the command is running. @dougm would love your feedback as output is a bit messy but I didn't want to rock the boat too hard and kept some of file output like filename etc.

I additionally added a -debug to default to the stdout LogType method if no GOVC_DEBUG_PATH is not set so you don't magically write some log files without the govc user knowing (which happened to me)

### DIFF
--- a/govc/flags/debug.go
+++ b/govc/flags/debug.go
@@ -68,7 +68,11 @@ func (flag *DebugFlag) Process(ctx context.Context) error {
 	return flag.ProcessOnce(func() error {
 		// Base path for storing debug logs.
 		r := os.Getenv("GOVC_DEBUG_PATH")
-		if r == "" {
+		switch r {
+		case "-":
+			debug.SetProvider(&debug.LogProvider{})
+			return nil
+		case "":
 			r = home
 		}
 		r = filepath.Join(r, "debug")
--- a/vim25/debug/debug.go
+++ b/vim25/debug/debug.go
@@ -18,8 +18,7 @@ package debug
 
 import (
 	"io"
-	"os"
-	"path"
+	"regexp"
 )
 
 // Provider specified the interface types must implement to be used as a
@@ -31,6 +30,7 @@ type Provider interface {
 }
 
 var currentProvider Provider = nil
+var scrubPassword = regexp.MustCompile(`<password>(.*)</password>`)
 
 func SetProvider(p Provider) {
 	if currentProvider != nil {
@@ -54,28 +54,9 @@ func Flush() {
 	currentProvider.Flush()
 }
 
-// FileProvider implements a debugging provider that creates a real file for
-// every call to NewFile. It maintains a list of all files that it creates,
-// such that it can close them when its Flush function is called.
-type FileProvider struct {
-	Path string
+func Scrub(in []byte) []byte {
+	out := string(in)
+	out = scrubPassword.ReplaceAllString(out, `<password>********</password>`)
 
-	files []*os.File
-}
-
-func (fp *FileProvider) NewFile(p string) io.WriteCloser {
-	f, err := os.Create(path.Join(fp.Path, p))
-	if err != nil {
-		panic(err)
-	}
-
-	fp.files = append(fp.files, f)
-
-	return f
-}
-
-func (fp *FileProvider) Flush() {
-	for _, f := range fp.files {
-		f.Close()
-	}
+	return []byte(out)
 }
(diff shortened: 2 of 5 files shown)


luthermonson:
ah crap i commited my build binary.. fixing

luthermonson:
removed binary and force pushed, sorry about that.

govc/flags/debug.go
```diff
@@ -69,7 +69,8 @@ func (flag *DebugFlag) Process(ctx context.Context) error {
 		// Base path for storing debug logs.
 		r := os.Getenv("GOVC_DEBUG_PATH")
 		if r == "" {
-			r = home
```
dougm: Don't want to do this as it makes `govc ... -debug` output to stderr by default.  We could do something like:
```go
switch r {
      case "-":
           debug.SetProvider(&debug.LogProvider{})
           return nil
      case "":
             r = home
}
```
Various other commands use `-` to specify stdout (stderr in this case) instead of a file.

vim25/debug/debug.go
```diff
@@ -54,28 +54,12 @@ func Flush() {
...
-}
-
-func (fp *FileProvider) NewFile(p string) io.WriteCloser {
-	f, err := os.Create(path.Join(fp.Path, p))
-	if err != nil {
-		panic(err)
+func Scrub(in []byte) []byte {
+	out := string(in)
+	if strings.Contains(out, "<password>") {
+		var re = regexp.MustCompile(`<password>(.*)</password>`)
```
dougm: This can be moved to a package-level var so its only compiled once.

vim25/debug/debug.go
```diff
@@ -54,28 +54,12 @@ func Flush() {
...
-	files []*os.File
-}
-
-func (fp *FileProvider) NewFile(p string) io.WriteCloser {
-	f, err := os.Create(path.Join(fp.Path, p))
-	if err != nil {
-		panic(err)
+func Scrub(in []byte) []byte {
+	out := string(in)
+	if strings.Contains(out, "<password>") {
```
dougm: Can probably drop this and just do the regex.

vim25/debug/file.go
```diff
@@ -0,0 +1,52 @@
+package debug
```
dougm: Need license header here.

vim25/debug/log.go
```diff
@@ -0,0 +1,33 @@
+package debug
```
dougm: Need license header here.

vim25/debug/debug.go
luthermonson: yup great point

vim25/debug/file.go
luthermonson: np

vim25/debug/log.go
luthermonson: np

vim25/debug/debug.go
luthermonson: i didnt want to run all text through the regex  and only the SessionManager call with it.. i dont mind either way so ill change.

govc/flags/debug.go
luthermonson: ya i thought this one might be dinged. ill switch to -

luthermonson:
@dougm updated per feedback and rebased to one commit and force pushed

luthermonson:
@dougm added readme update to deocument `-debug` since its looking like youre going to merge this.

dougm:
Thanks @luthermonson , esp. for adding the long overdue docs.  Keep in mind that debug also works with json (tags, content library, etc).  Personally I always have:
```console
% env | grep GOVC_DEBUG
GOVC_DEBUG_PATH_RUN=last
GOVC_DEBUG=true
```
And the I run `$GOPATH/src/github.com/vmware/govmomi/scripts/debug-xmlformat.sh` to view request/response logs.

I think it'd be awesome to have `GOVC_DEBUG_PATH=-` pretty-print the output as `debug-xmlformat.sh` does, but using Go's json + xml pretty printers.  But that could be addressed in a separate PR.

PR #1649, opened 2019-10-04
Allow sending multiple characters through -c and name the keys

freedge:
This pull request does 2 things:

- it allows users to give multiple, comma separated, values to vm.keystrokes -c
- it also provides aliases for the KEYs, so when you don't type 0x28 or 0x2b, but KEY_ENTER and KEY_TAB, which is much nicer.

If you agree with the idea, I can update the USAGE.md as well. I am still testing this btw.


govc/vm/keystrokes.go
```diff
@@ -156,7 +521,7 @@ func (cmd *keystrokes) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.VirtualMachineFlag.Register(ctx, f)
 
 	f.StringVar(&cmd.UsbHidString, "s", "", "Raw String to Send")
-	f.StringVar(&cmd.UsbHidCode, "c", "", "USB HID Code (hex)")
+	f.StringVar(&cmd.UsbHidCodes, "c", "", "USB HID Codes (hex), comma separated")
```
king-jam: Can we change this to "...or hex alias" or similar?

freedge: Thanks. And thank you for the work you did on govc in the first place!

govc/USAGE.md
```diff
@@ -4514,6 +4515,190 @@ Options:
   -s=                    Raw String to Send
   -vm=                   Virtual machine [GOVC_VM]
 ```
+<details><summary>List of aliases for USB HID codes</summary>
```
dougm: USAGE.md is generated (via `make doc`), you just need to put this list with `keystrokes.Description()`, otherwise the next `make doc` update will overwrite it.

king-jam:
Apologies, I didn't realize that was the before my sign-off.

govc/USAGE.md
freedge: that would make the output of govc vm.keystrokes -h
very verbose though.

PR #1654, opened 2019-10-07
[wip] Implementation of host.esxtop

freedge:
This is a work in progress. pr only created to get feedback.

This pr demonstrates the usage of the API described in https://www.virtuallyghetto.com/2013/01/retrieving-esxtop-performance-data.html

to retrieve statistics from esxtop. It outputs the raw data as replied by the service... although with some scripting nothing is unreadable:

```
> ./govc host.esxtop  -debug -host.dns myhost.net  | grep '^|SchedGroup|' | cut -d '|' -f 4-7,24-28 | grep -E 'GroupName,STR|frigo'
GroupName,STR|IsValid,B|IsVM,B|VMName,STR|NumOfLocalitySwap,U32|NumOfLoadSwap,U32|RemoteMemoryInKB,U32|LocalMemoryInKB,U32|LocalityPct,U32
vm.13648133|1|1|frigo001|0|0|0|12578816|100
uw.13648133|1|0|frigo001|0|0|0|0|0
vmm.13648133|1|0|frigo001|0|0|0|0|0
```

Here we see that the LocalityPct is 100% for my VM, which is wrong in my case... Need to get the right value before progressing with this pr.


dougm:
Awesome @freedge !  This feature has been requested a few times too.  Happy to review when you're ready.

dougm:
@freedge subject still says `[wip]`, are you still working on this or is it ready for review?  Thanks.

freedge:
not really working on this no. I actually was expecting more from esxtop... the metrics I got from esxtop I can also get using other commands from govc (such as metrics.sample) so I don't see the need for it anymore.

Declining this pr for the time being

PR #1655, opened 2019-10-08
govc: add vm.customize command

dougm:
vcsim: add CustomizationSpecManager support


dougm:
@gressd I folded your PR #1576 in here.  With the git history it was easier this way, but the commit still has your author attribution.  I refactored vm.customize quite a bit, and added vcsim support + tests.  Can you review?

PR #1662, opened 2019-10-10
adding managed obj type to table

luthermonson:
Since you can't assign every field to MO type I added the ManagedObjectType to the table output. Also added -t filter to limit the list to only certain Manage Object Types.

```
➜  govc git:(field-type) ./govc-dev fields.ls
1    AutoDeploy.MachineIdentity                                  HostSystem
203  asdffffff                                                   VirtualMachine
2    com.vmware.vcIntegrity.customField.scheduledTask.action     ScheduledTask
3    com.vmware.vcIntegrity.customField.scheduledTask.signature  ScheduledTask
4    com.vmware.vcIntegrity.customField.scheduledTask.target     ScheduledTask
➜  govc git:(field-type) ./govc-dev fields.ls -t VirtualMachine
203  asdffffff  VirtualMachine
➜  govc git:(field-type) ./govc-dev fields.ls -t ScheduledTask
2  com.vmware.vcIntegrity.customField.scheduledTask.action     ScheduledTask
3  com.vmware.vcIntegrity.customField.scheduledTask.signature  ScheduledTask
4  com.vmware.vcIntegrity.customField.scheduledTask.target     ScheduledTask
```

### DIFF
--- a/govc/fields/ls.go
+++ b/govc/fields/ls.go
@@ -30,6 +30,7 @@ import (
 
 type ls struct {
 	*flags.ClientFlag
+	kind string
 }
 
 func init() {
@@ -39,13 +40,16 @@ func init() {
 func (cmd *ls) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
 	cmd.ClientFlag.Register(ctx, f)
+
+	f.StringVar(&cmd.kind, "type", "", "Filter by a Managed Object Type")
 }
 
 func (cmd *ls) Description() string {
 	return `List custom field definitions.
 
 Examples:
-  govc fields.ls`
+  govc fields.ls
+  govc fields.ls -type VirtualMachine`
 }
 
 func (cmd *ls) Run(ctx context.Context, f *flag.FlagSet) error {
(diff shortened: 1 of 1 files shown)


govc/fields/ls.go
```diff
@@ -64,10 +68,17 @@ func (cmd *ls) Run(ctx context.Context, f *flag.FlagSet) error {
 		return err
 	}
 
-	tw := tabwriter.NewWriter(os.Stdout, 2, 0, 2, ' ', 0)
+	tw := tabwriter.NewWriter(os.Stdout, 3, 0, 2, ' ', 0)
 
 	for _, def := range field {
-		fmt.Fprintf(tw, "%d\t%s\n", def.Key, def.Name)
+		if cmd.mot == "" {
```
dougm: Looks like output is the same in both conditions, can you collapse this?  As in:
```go
if cmd.mot == "" || cmd.mot == def.ManagedObjectType {
    fmt.Printf(...)
}
```

govc/fields/ls.go
```diff
@@ -39,13 +40,16 @@ func init() {
 func (cmd *ls) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
 	cmd.ClientFlag.Register(ctx, f)
+
+	f.StringVar(&cmd.mot, "t", "", "Filter by Object Type")
```
dougm: cosmetic: in other commands we name the var `kind`, flag `-type` and description "Managed object type", for example:

https://github.com/vmware/govmomi/blob/c0f4e9754e39f3910357a6e3e56bbfb387b35a65/govc/fields/set.go#L43

https://github.com/vmware/govmomi/blob/c0f4e9754e39f3910357a6e3e56bbfb387b35a65/govc/fields/add.go#L42

Though I know elsewhere in the govc commands we have a mix of `-t` and `-type`

govc/fields/ls.go
luthermonson: ya this is cleaner, not sure what i was thinking.

govc/fields/ls.go
luthermonson: done.

govc/fields/ls.go
```diff
@@ -39,13 +40,16 @@ func init() {
 func (cmd *ls) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
 	cmd.ClientFlag.Register(ctx, f)
+
+	f.StringVar(&cmd.kind, "type", "", "Filter by a Managed Oject Type")
```
dougm: typo: "Oject"

luthermonson: sigh

dougm:
Thanks @luthermonson

PR #1666, opened 2019-10-12
vcsim: simplify container vm arguments input

dougm:
Rather than require users to supply a json encoded argument list,
let sh|bash do the work of parsing.


dougm:
Note this borrows the toolbox logic: https://github.com/vmware/govmomi/blob/e9b8cf013e40604beb64dd668a50c74075ec6d26/toolbox/process.go#L70-L75

PR #1669, opened 2019-10-15
govc: fixup usage suggestions

dougm:
3e3d351 added usage suggestions, this follow up does:

- make 'govc -h' behave as it used to

- if there are no command suggestions, print the complete list of commands


dougm:
@luthermonson can you review?

Issue #1667, opened 2019-10-14
Unable to update individual port under a DVS portgroup


jingyizPensando:
Hi VMware team,

We are able to create a DVS and add portgroup to it. We set "VlanOverrideAllowed" to be true when formalizing "types.DVPortgroupConfigSpec.Policy". 

We can see that portgroup is being created with Vlan override enabled. Then we want to config individual port under this portgroup(for example, we want to change a vlan of one port, all other ports remain the same vlan with the portgroup).

We use "FetchDVPorts" to get all the ports which meet the criteria. "FetchDVPorts" returns an array of DistributedVirtualPort.

For example:
`ports, err := dvs.FetchDVPorts(ctx, &criteria)`

This one completed successfully. We are able to dump information of each port.

Our goal is to update individual port under a DVS portgroup. We were not able to find any API to reconfigure individual port(Reconfigure(ctx context.Context, spec types.DVPortgroupConfigSpec) is only for updating portgroup).

We add our own API to update individual port under portgroup by calling ReconfigureDVPort_Task: 

```
func (p DistributedVirtualPortgroup) ReconfigureDvsPort(ctx context.Context, spec []types.DVPortConfigSpec) (*Task, error) {
    req := types.ReconfigureDVPort_Task{
        This: p.Reference(),
        Port: spec,
    }
​
    res, err := methods.ReconfigureDVPort_Task(ctx, p.Client(), &req)
    if err != nil {
        return nil, err
    }
​
    return NewTask(p.Client(), res.Returnval), nil
}

```

Below is our way to update the port[0](first port in the array) by leveraging our own API above:

```
ports, err := myDVS.ObjDvs.FetchDVPorts(ctx, &criteria)

portVlan := new(types.VmwareDistributedVirtualSwitchVlanIdSpec)
portVlan.VlanId = vlanId

portSpec := types.DVPortConfigSpec{
    ConfigVersion: ports[0].Config.ConfigVersion,
    //Description:   ports[0].Config.Description,
    Key: ports[0].Key,
    //Name:          ports[0].Config.Name,
    Operation: string("edit"),
    Scope:     ports[0].Config.Scope,
    Setting: &types.VMwareDVSPortSetting{
        Vlan: portVlan,
    },  
} 

task, err := myPG.ObjPg.ReconfigureDvsPort(ctx, []types.DVPortConfigSpec{portSpec})
```

We can see an error after calling this "ReconfigureDvsPort"
[...]

PR #1670, opened 2019-10-15
Add a method to update ports on a distributed virtual switch

jingyizPensando:
As we discussed in issue: https://github.com/vmware/govmomi/issues/1667

Open this PR to add a new method in order to provide a way for developers to update ports on a distributed virtual switch.


dougm:
Thanks @jingyizPensando , can you name the method `ReconfigureDVPort`, without the `s`.. as that matches the method name it is wrapping.

jingyizPensando:
@dougm  Done! Thanks for the review.

PR #1678, opened 2019-10-21
Avoid possible deadlock in KeepAliveHandler

dougm:
The call to stop() triggered by a handler error would cause a deadlock.
Avoid the use of deferring notifyWaitGroup.Done(), instead calling it explicitly
on both return paths.

Note that the deadlock was not seen within the test until adding the defered
calls to Logout()


dougm:
Thanks for your help @sahangama

PR #1687, opened 2019-10-30
Generate pointer type for ResourceReductionToToleratePercent

### LINKED ISSUE #1686: ResourceReductionToToleratePercent should not be omitempty
ResourceReductionToToleratePercent could be between 0 and 100.
if ResourceReductionToToleratePercent set to 0 marshaller will consider it as empty and do not include it into the payload.

dougm:
Another field optional in the wsdl, but with Go's "omitempty" semantics prevents a value of '0'.

Fixes #1686

### DIFF
--- a/vim25/types/types.go
+++ b/vim25/types/types.go
@@ -8231,7 +8231,7 @@ func init() {
 type ClusterDasAdmissionControlPolicy struct {
 	DynamicData
 
-	ResourceReductionToToleratePercent int32 `xml:"resourceReductionToToleratePercent,omitempty"`
+	ResourceReductionToToleratePercent *int32 `xml:"resourceReductionToToleratePercent"`
 }
 
 func init() {
--- a/vim25/types/types.go
+++ b/vim25/types/types.go
@@ -5931,7 +5931,7 @@ type AttachDiskRequestType struct {
 	DiskId        ID                     `xml:"diskId"`
 	Datastore     ManagedObjectReference `xml:"datastore"`
 	ControllerKey int32                  `xml:"controllerKey,omitempty"`
-	UnitNumber    *int32                 `xml:"unitNumber"`
+	UnitNumber    *int32                 `xml:"unitNumber,omitempty"`
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


aeremenk0:
@dougm: it is better to leave `ResourceReductionToToleratePercent` to be `int32` (remove pointer), if you just remove `omitempty` it will do. If you change type to `*int32` it will have implication for vshpere terraform provider.

dougm:
@aeremenk0 if we just remove `omitempty`, the caller has no way to avoid sending a value for that field.  We chose a pointer type for the other fields to avoid the case where a caller may unintentionally overwrite a value with `0`.  I'd prefer to keep things consistent, though I understand it breaks existing API compatibility.

Another option is for the Terraform provider to define their own `ClusterDasAdmissionControlPolicy` struct (copy-n-paste from types.go) and remove `omitempty` there.  The struct is small and unlikely to change in future versions.

vim25/types/types.go
```diff
@@ -8231,7 +8231,7 @@ func init() {
 type ClusterDasAdmissionControlPolicy struct {
 	DynamicData
 
-	ResourceReductionToToleratePercent int32 `xml:"resourceReductionToToleratePercent,omitempty"`
+	ResourceReductionToToleratePercent *int32 `xml:"resourceReductionToToleratePercent"`
```
aeremenk0: if you still want to let users skip that value it is better to return `omitempty` back. If it is not there the value will be mandatory.
ResourceReductionToToleratePercent *int32 `xml:"resourceReductionToToleratePercent,omitempty"`

dougm: That's not the case with the xml encoder, see: https://golang.org/src/encoding/xml/marshal.go?s=3363:3406#L69

> or, if the pointer is nil, by writing nothing

aeremenk0: that will do! 
I would keep `omitempty` just for the simplicity so users would not need to go to the marshaller docs.

dougm: Sure, pushed a new commit that does so.

PR #1701, opened 2019-11-07
Content Library: VM Template support

dougm:
- Add vapi and vcsim support for VM Template (aka vmtx) related APIs

- Add govc library.clone command

- Add vmtx support to govc library.deploy command

### DIFF
--- a/govc/cli/register.go
+++ b/govc/cli/register.go
@@ -16,11 +16,21 @@ limitations under the License.
 
 package cli
 
+import "os"
+
 var commands = map[string]Command{}
 
 var aliases = map[string]string{}
 
-func Register(name string, c Command) {
+// hideUnreleased allows commands to be compiled into the govc binary without being registered by default.
+// Unreleased commands are omitted from 'govc -h' help text and the generated govc/USAGE.md
+// Setting the env var GOVC_SHOW_UNRELEASED=true enables any commands registered as unreleased.
+var hideUnreleased = os.Getenv("GOVC_SHOW_UNRELEASED") != "true"
+
+func Register(name string, c Command, unreleased ...bool) {
+	if len(unreleased) != 0 && unreleased[0] && hideUnreleased {
+		return
+	}
 	commands[name] = c
 }
 
--- a/govc/library/checkout.go
+++ b/govc/library/checkout.go
@@ -1,0 +1,141 @@
+/*
+Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package library
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/vapi/library"
+	"github.com/vmware/govmomi/vapi/library/finder"
+	"github.com/vmware/govmomi/vapi/rest"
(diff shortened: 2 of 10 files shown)


govc/cli/register.go
```diff
@@ -16,11 +16,18 @@ limitations under the License.
...
+import "os"
+
 var commands = map[string]Command{}
 
 var aliases = map[string]string{}
 
-func Register(name string, c Command) {
+var hideUnreleased = os.Getenv("GOVC_SHOW_UNRELEASED") != "true"
+
+func Register(name string, c Command, unreleased ...bool) {
```
corrieb: Would be good to have a comment here explaining the purpose of unreleased. It's not at all obvious

govc/library/checkout.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (cmd *checkout) Usage() string {
+	return "PATH NAME"
+}
+
+func (cmd *checkout) Description() string {
+	return `Check out Content Library item PATH to vm NAME.
+
+Examples:
+  govc library.checkout -cluster my-cluster my-content/template-vm-item my-vm`
```
corrieb: Feels to me like the template parameter should maybe have a flag - multiple unflagged trailing parameters feel somewhat hard to reason about. But I'd prefer consistency with other options. Up to you.

govc/library/checkout.go
```diff
@@ -0,0 +1,141 @@
...
+				Folder: folder.Reference().Value,
+			},
+		}
+		if pool != nil {
+			spec.Placement.ResourcePool = pool.Reference().Value
+		}
+		if host != nil {
+			spec.Placement.Host = host.Reference().Value
+		}
+		if cluster != nil {
```
corrieb: Don't you require a cluster? Where would it go if Cluster isn't specified?

govc/library/clone.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (cmd *clone) Usage() string {
+	return "PATH NAME"
+}
+
+func (cmd *clone) Description() string {
+	return `Clone VM to Content Library PATH.
+
+Examples:
+  govc library.clone -vm template-vm my-content template-vm-item`
```
corrieb: Same comment as above about multiple freeform parameters that depend on ordering.

govc/library/deploy.go
```diff
@@ -168,35 +169,68 @@ func (cmd *deploy) Run(ctx context.Context, f *flag.FlagSet) error {
...
+			}
 
-		ref, err := m.DeployLibraryItem(ctx, item.ID, deploy)
-		if err != nil {
-			return err
+			deploy := vcenter.DeployTemplate{
+				Name:          name,
+				Description:   cmd.Options.Annotation,
+				DiskStorage:   storage,
+				VMHomeStorage: storage,
```
corrieb: Would it ever make sense for the DiskStorage and VMHomeStorage to be in separate locations?

vapi/simulator/simulator.go
```diff
@@ -1490,3 +1498,219 @@ func (s *handler) libraryItemDeployID(w http.ResponseWriter, r *http.Request) {
...
+
+	var spec struct {
+		vcenter.DeployTemplate `json:"spec"`
+	}
+	if !s.decode(r, w, &spec) {
+		return
+	}
+
+	if r.Method == http.MethodGet {
+		t := &vcenter.TemplateInfo{ /*TODO*/ }
```
corrieb: Could this TODO be more descriptive?

govc/library/checkout.go
dougm: There are a few library commands, and a few others in govc that take 2 args when they are both required.  This one and the others are similar to the `cp` command: `cp SOURCE DEST`

govc/library/checkout.go
dougm: One of `pool`, `host` or `cluster` is required for placement.  The error message from vCenter makes that clear if none are set.

govc/library/deploy.go
dougm: Maybe, but we don't support splitting them out in the soap based commands such as `vm.create` and `vm.clone`.  Never had a request to support that, so kept em the same here too.

vapi/simulator/simulator.go
dougm: Done

govc/cli/register.go
dougm: Done

PR #1702, opened 2019-11-07
Added SSL Thumbprint option to create

serokles:
Added SSL Thumbprint option to create library subscription.  Call was failing in both govc and postman without a valid SSL thumbprint present in the JSON

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,3 +11,5 @@ require (
 	github.com/kr/text v0.1.0 // indirect
 	github.com/vmware/vmw-guestinfo v0.0.0-20170707015358-25eff159a728
 )
+
+go 1.13
--- a/go.mod
+++ b/go.mod
@@ -11,5 +11,3 @@ require (
 	github.com/kr/text v0.1.0 // indirect
 	github.com/vmware/vmw-guestinfo v0.0.0-20170707015358-25eff159a728
 )
-
-go 1.13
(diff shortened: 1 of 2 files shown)


govc/library/create.go
```diff
@@ -49,6 +49,7 @@ func (cmd *create) Register(ctx context.Context, f *flag.FlagSet) {
 	f.StringVar(&cmd.sub.SubscriptionURL, "sub", "", "Subscribe to library URL")
 	f.StringVar(&cmd.sub.UserName, "sub-username", "", "Subscription username")
 	f.StringVar(&cmd.sub.Password, "sub-password", "", "Subscription password")
+	f.StringVar(&cmd.sub.SslThumbprint, "ssl-thumbprint", "", "SSL Tumbprint")
```
dougm: Thanks @serokles , let's name the flag just `"thumbprint"` without the `ssl-` prefix.  We use that in a few other places, such as the `host.add`, `host.reconnect` and `cluster.add` commands:
https://github.com/vmware/govmomi/blob/7b110ddb44f68a7223d9d6d9e5b1ef8ede929701/govc/flags/host_connect.go#L56

There's a typo in the description too `"Tumbprint"`

go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	github.com/kr/text v0.1.0 // indirect
 	github.com/vmware/vmw-guestinfo v0.0.0-20170707015358-25eff159a728
 )
+
+go 1.13
```
dougm: Should exclude this change, we can update to go 1.13 separately.

Issue #1700, opened 2019-11-07
Create Subscribed Content Library


serokles:
Am having an issue deploying a subscribed content library using govc:

`govc library.create -sub=https://download3.vmware.com/software/vmw-tools/lib.json -ds=General1 -dc=HomeLab vGhettoLib`

Returns:
`govc: 400 Bad Request: {"type":"com.vmware.vapi.std.errors.resource_inaccessible","value":{"messages":[{"args":["https://download3.vmware.com/software/vmw-tools/lib.json"],"default_message":"Connection to VCSP server https://download3.vmware.com/software/vmw-tools/lib.json failed.","id":"com.vmware.vdcs.cls-main.connection_to_vcsp_server_failed"}]}}
`
I can add this using the GUI with no issue and using curl I can get the json returned form the machine.
Am using govc version 0.22.0 built from using:
`go get -u github.com/vmware/govmomi/govc`
`go install github.com/vmware/govmomi/govc`

I can see issues with ls and rm have been resolved but cannot find anything on this issue

serokles:
In fact got the same error in Postman using the API if I did not specify an ssl thumbprint.  Added the following line to create.go:
`f.StringVar(&cmd.sub.SslThumbprint, "ssl-thumbprint", "", "SSL Tumbprint")`

and there is now an option to add a thumbprint.  This then allows the library to be created.  If I knew how to go about a pull request I would...

dougm:
Fixed by #1702

PR #1706, opened 2019-11-13
govc: add default library.create thumbprint

dougm:
Follow up to #1700 and #1702

Default thumbprint to soap.Client's db (-tls-known-hosts)

Otherwise, set thumbprint to that of the subscription URL host if -k/GOVC_INSECURE is true.

### DIFF
--- a/vapi/library/library.go
+++ b/vapi/library/library.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dougm:
@serokles can you review?

serokles:
Nice - worked like a charm without having to manually enter the SHA1 thumbprint.  Cannot test it with GOVC_INSECURE=true as I dont have any valid CA's on home lab.

dougm:
Great, thanks for reviewing and testing it out @serokles

PR #1711, opened 2019-11-14
Add ability to set DVS discovery protocol on create and change

thadc23:
The requirement of the additional`Operation` field on `LinkDiscoveryProtocolConfig` requires the check of the flag, rather than direct binding to the `Protocol` field in the config spec. They need to be set together if the protocol is provided.

### DIFF
--- a/govc/dvs/change.go
+++ b/govc/dvs/change.go
@@ -34,6 +34,8 @@ type change struct {
 	types.DVSCreateSpec
 
 	configSpec *types.VMwareDVSConfigSpec
+
+	dProtocol string
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


dougm:
@thadc23 vmwclabot check is pending, had you already signed the cla?

thadc23:
Re-signed and hopefully legal gets that taken care of

dougm:
@thadc23 have you heard anything back?  I can try to track this down if it got lost somehow.

thadc23:
I haven't heard from our legal, but I have signed the form twice.

dougm:
Regarding the CLA @thadc23 , given that you were Pivotal and now a VMware employee, no CLA is required.  Welcome!

PR #1716, opened 2019-11-16
govc: add doc on vm.info -r flag

### LINKED ISSUE #1703: Get a VM's portgroup name?
In PowerCLI, one can do:

get-virtualportgroup -vm my-vm

... to retrieve the current port group setting on a VM. I find myself in the middle of a massive migration project, and we've already decided to employ govc for it, but I'm not finding a way to:

1. Get a VM's current port group and save to variable
2. Migrate the VM to a new vCenter using our working script
3. Use regex to find a similarly named port group on the new vCenter and then set the migrated VM's port group to the new name so that it comes up cleanly in the new vCenter without the need for user interaction

If this functionality were added, it would make our lives far easier. Any plans to do so?

dougm:
Fixes #1703

### DIFF
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -4425,8 +4425,12 @@ Usage: govc vm.info [OPTIONS] VM...
 
 Display info for VM.
 
+The '-r' flag displays additional info for CPU, memory and storage usage,
+along with the VM's Datastores and Networks and PortGroups.
+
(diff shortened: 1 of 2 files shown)


dougm:
@nmonday can you review?

dougm:
Thanks @nmonday

PR #1723, opened 2019-11-26
Revert "gen: retain omitempty field tag with int pointer types"

dougm:
This reverts commit 5e6f5e3f91cefdf32ec1e56cf99a14e41aa03abd.


dougm:
Revert was part of #1687 - requested for doc purposes, but had the side effect of omitting value of `0` even with pointer field types.

PR #1734, opened 2019-12-06
vcsim: bind mount BIOS UUID DMI files

### LINKED ISSUE #1709: vC Sim should bind mount BIOS UUID to DMI files when simulating lifecycle with containers
# Overview

This issue tracks a feature request for vC Simulator when it is simulating a VM's lifecycle with a container.

## Use case

The request is for vC Simulator to take the VM's BIOS UUID, ex. `04F419FE-8688-4066-A16D-CBC4193D5B93` and bind mount it to two files inside of the container:

1. `/sys/class/dmi/id/product_uuid`

    This file's content should mirror the value of the BIOS UUID

    ```shell
    $ cat /sys/class/dmi/id/product_uuid
    04F419FE-8688-4066-A16D-CBC4193D5B93
    ```

2. `/sys/class/dmi/id/product_serial`

    This file's content is a transformed version of the BIOS UUID.

    ```shell
    $ cat /sys/class/dmi/id/product_serial
    VMware-fe 19 f4 04 88 86 66 40-a1 6d cb c4 19 3d 5b 93
    ```

    * The UUID is prefixed with `VMware-`
    * The 32-character hexadecimal UUID string is formatted into distinct pairs of characters, each pair representing a byte value (`2^8`).
    * The [UUID should be encoded](https://en.wikipedia.org/wiki/Universally_unique_identifier#Encoding) where the first three components are little-endian, and the final two components are big-endian.
    * The 32 characters are separated by a `-` character after eight bytes (16 characters).
      * Please note that all characters before the `-` separator are little-endian, and all characters after the `-` separator are big-endian.

## Requirements

This feature is needed for the following reasons:

* The aforementioned files are read-only and cannot be modified once the guest is running. The only way to supply the correct data to these files when the guest is a container is to bind mount the data into the container (there is a similar feature for [setting these specific values when running a VM with VirtualBox or Fusion](https://github.com/vmware/simple-k8s-test-env/blob/1d251afbabbc27f9eab01cb70d9060e017734fc4/vagrant/Vagrantfile#L289-L302)).
* The VM lifecycle simulation feature of vC Sim is useful for performing integration or end-to-end (e2e) testing with Kubernetes because of the aforementioned files' relationship with the discovery of the _node ID_.
  * The node ID is used as the basis for the provider ID for:
    * The [Kubernetes cloud provider for vSphere](https://github.com/kubernetes/cloud-provider-vsphere)
    * The [Kubernetes cluster API provider for vSphere](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere)
  * Unless specified as a flag when starting the kubelet, Kubernetes determines a node's initial ID by reading the DMI information.
    1. The kubelet attempts to gather information about the node's status, [including the node's SystemUUID](https://github.com/kubernetes/kubernetes/blob/0599ca2bcfcae7d702f95284f3c2e2c2978c7772/pkg/kubelet/nodestatus/setters.go#L263). The method used to obtain this information depends on the cAdvisor `Interface`.
    1. The cAdvisor `Interface` [is initialized with a real system filesystem implementation](https://github.com/kubernetes/kubernetes/blob/9488fbef64afd580d053347e57819817e51ac0f9/pkg/kubelet/cadvisor/cadvisor_linux.go#L87) (`SysFs`).
    1. The cAdvisor `Interface` initializes a new `Manager` [with the `SysFs` instance](https://github.com/kubernetes/kubernetes/blob/9488fbef64afd580d053347e57819817e51ac0f9/pkg/kubelet/cadvisor/cadvisor_linux.go#L104).
    1. The cAdvisor `Manager` [gathers information about the machine](https://github.com/kubernetes/kubernetes/blob/3b440dfd5550b13278e40975a7cba13a3f129475/vendor/github.com/google/cadvisor/manager/manager.go#L188) with `machine.Info(SysFs,...)`.
    1. The `machine.Info` call uses the `SysFs` instance to [get the system's UUID via `sysinfo.GetSystemUUID`](https://github.com/kubernetes/kubernetes/blob/3b440dfd5550b13278e40975a7cba13a3f129475/vendor/github.com/google/cadvisor/machine/info.go#L138).
    1. The `sysinfo.GetSystemUUID` function is simply [a wrapper](https://github.com/kubernetes/kubernetes/blob/d1e8702d36947ccc23ad820eaae7f4afbaf0b058/vendor/github.com/google/cadvisor/utils/sysinfo/sysinfo.go#L201-L203) for the `SysFs` instance's `GetSystemUUID` function.
    1. And the real `SysFs` implementation of the `GetSystemUUID` function reads the system UUID via several methods, [starting with the DMI information](https://github.com/kubernetes/kubernetes/blob/c466fd9eaf31a6a12f901369d13129f6a8853da4/vendor/github.com/google/cadvisor/utils/sysfs/sysfs.go#L237-L249).

## Notes

* The Kubernetes cloud provider for vSphere [implements this feature with Kind for doing integration testing](https://github.com/kubernetes/cloud-provider-vsphere/tree/master/test/integration).

cc @dougm @imikushin @corrieb

dougm:
For use within container simulated VMs

Fixes #1709

### DIFF
--- a/govc/test/vcsim.bats
+++ b/govc/test/vcsim.bats
@@ -262,6 +262,19 @@ load test_helper
   run govc vm.destroy $vm
   assert_success
 
+  vm=DC0_C0_RP0_VM0
+  run govc vm.change -vm $vm -e RUN.container="busybox grep VMware- /sys/class/dmi/id/product_serial"
+  assert_success
+
+  run govc vm.power -on $vm
+  assert_success
+
+  run docker inspect -f '{{.State.ExitCode}}' $vm
+  assert_success "0"
+
+  run govc vm.destroy $vm
+  assert_success
+
   vm=DC0_C0_RP0_VM1
   run govc vm.change -vm $vm -e RUN.container="busybox sh -c 'sleep \$VMX_GUESTINFO_SLEEP'" -e guestinfo.sleep=500
   assert_success
--- a/simulator/container.go
+++ b/simulator/container.go
@@ -17,13 +17,17 @@ limitations under the License.
 package simulator
 
 import (
+	"archive/tar"
 	"bytes"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"log"
 	"os/exec"
 	"strings"
+	"time"
 
+	"github.com/google/uuid"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
@@ -88,6 +92,46 @@ func (c *container) inspect(vm *VirtualMachine) error {
 	return nil
 }
 
+// createDMI writes BIOS UUID DMI files to a container volume
+func (c *container) createDMI(vm *VirtualMachine, name string) error {
+	cmd := exec.Command("docker", "run", "--rm", "-i", "-v", name+":"+"/"+name, "busybox", "tar", "-C", "/"+name, "-xf", "-")
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		return err
+	}
+
+	err = cmd.Start()
+	if err != nil {
+		return err
+	}
+
+	tw := tar.NewWriter(stdin)
+
+	dmi := []struct {
+		name string
+		val  func(uuid.UUID) string
+	}{
+		{"product_uuid", productUUID},
+		{"product_serial", productSerial},
+	}
+
+	for _, file := range dmi {
+		val := file.val(vm.uid)
+		_ = tw.WriteHeader(&tar.Header{
+			Name:    file.name,
+			Size:    int64(len(val) + 1),
+			Mode:    0444,
+			ModTime: time.Now(),
+		})
+		_, _ = fmt.Fprintln(tw, val)
+	}
+
+	_ = tw.Close()
+	_ = stdin.Close()
+
+	return cmd.Wait()
+}
+
 // start runs the container if specified by the RUN.container extraConfig property.
 func (c *container) start(vm *VirtualMachine) {
 	if c.id != "" {
@@ -132,7 +176,14 @@ func (c *container) start(vm *VirtualMachine) {
 	}
 
 	run := append([]string{"docker", "run", "-d", "--name", vm.Name}, env...)
-	run = append(run, "--env", "VMX_CONFIG_UUID="+vm.Config.Uuid)
+
+	volume := fmt.Sprintf("vcsim-%s-%s", vm.Name, vm.uid)
+	if err := c.createDMI(vm, volume); err != nil {
+		log.Printf("%s: %s", vm.Name, err)
+		return
+	}
+	run = append(run, "-v", fmt.Sprintf("%s:%s:ro", volume, "/sys/class/dmi/id"))
+
 	args = append(run, args...)
 	cmd := exec.Command(shell, "-c", strings.Join(args, " "))
 	out, err := cmd.Output()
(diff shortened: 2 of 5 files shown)


imikushin:
What if vcsim itself is working in a container and connects to docker via a bind-mounted socket? Would the "VM" container still be able to start? Also, looks like failure to start the container isn't logged (in my setup it doesn't appear to be - the problem may be my setup :)).

Bottom line: I'd still like to have an env var and skip the bind-mounting if it doesn't work when running vcsim in a container.

akutz:
Hi @imikushin,

> What if vcsim itself is working in a container and connects to docker via a bind-mounted socket? Would the "VM" container still be able to start?

You just described the way the [upstream vSphere cloud provider handles testing](https://github.com/kubernetes/cloud-provider-vsphere/tree/master/test/integration). vCSim runs in a container in the K8s cluster being tested. It presents some unique challenges.

imikushin:
@akutz how can it possibly work? I mean bind-mounting files from one container into another container

akutz:
it works if the second container is running in the first container. Docker in Docker can happen two ways:

1. Mount the hosts Docker socket into a container
2. Run Docker in a container

The cloud provider’s tests accounts for both by allowing you to bind mount files from the host into the second container. The second DinD design is just normal Docker — albeit running in a container.

imikushin:
@akutz You're not saying option 1 doesn't have the right to exist anymore, are you?

dougm:
Haven't tried nested yet.. any difference between using a bind mount and volume in that case? Works the same w/o nested:

```diff
modified   simulator/container.go
@@ -151,7 +151,7 @@ func (c *container) start(vm *VirtualMachine) {
 			continue
 		}
 		_, _ = fmt.Fprintln(f, file.val(vm.uid))
-		run = append(run, "--mount", fmt.Sprintf("type=bind,source=%s,target=/sys/class/dmi/id/%s,readonly", f.Name(), file.name))
+		run = append(run, "-v", fmt.Sprintf("%s:/sys/class/dmi/id/%s:ro", f.Name(), file.name))
 		_ = f.Close()
 	}
```

dougm:
So vcsim is running in a container, writes the uuid files to its own /tmp/vcsim-$x directory, but when mounting through the docker socket that directory doesn't exist on the host and the mount fails.

I will add the env vars and skip the mounts if we're already running in a container for now.

dougm:
On second though, auto-skipping the mounts probably isn't right either.. maybe just an env var to disable for now @imikushin ?  Will need to look closer at all this to get the automatic part right I think.

akutz:
So what the the external vSphere cloud provider's integration testing does is use env vars to indicate:

* The root of the project if using DinD with a mapped sock ([`PROJECT_ROOT`](https://github.com/kubernetes/cloud-provider-vsphere/blob/fca2949a1e66c9e4c9e17f9b78237c37fa9862d8/test/integration/Makefile#L5-L6))
* Whether the project is using DinD via a mapped sock ([`DOCKER_SOCK_FROM_HOST`](https://github.com/kubernetes/cloud-provider-vsphere/blob/fca2949a1e66c9e4c9e17f9b78237c37fa9862d8/hack/make/docker.mk#L29-L34))

That way the [Kind config can use `{{PROJECT_ROOT}}`](https://github.com/kubernetes/cloud-provider-vsphere/blob/fca2949a1e66c9e4c9e17f9b78237c37fa9862d8/test/kind/kind-config.yaml#L1-L19), and it gets replaced with the root of the filesystem that contains the files that need to be bind mounted.

imikushin:
One possible solution working in all cases can be using docker volumes. A nice thing about volumes is you don't care about their physical location anymore. A downside is a bit more work to populate a volume, but that can be as simple as running a single command.

imikushin:
Docker commands to make this work would look something like this:
```sh
# create and populate the volume
docker run --rm -v dmi-id-04f419fe-8688-4066-a16d-cbc4193d5b93:/dmi-id photon:3.0 sh -c \
 'echo "04f419fe-8688-4066-a16d-cbc4193d5b93" > /dmi-id/product_uuid; \
  echo "VMware-fe 19 f4 04 88 86 66 40-a1 6d cb c4 19 3d 5b 93" > /dmi-id/product_serial'

# use the volume
docker run --name=simulated-vm \
  -v dmi-id-04f419fe-8688-4066-a16d-cbc4193d5b93:/sys/class/dmi/id \
  simulated-vm-container-image \
  command args..

# remove the volume
docker volume rm dmi-id-04f419fe-8688-4066-a16d-cbc4193d5b93
```

imikushin:
Note that in the example above it absolutely doesn't matter where these commands are run from. They work exactly the same way if run 
- from host
- from a container against the host's docker
- from a container against the nested docker

dougm:
@imikushin this sounds good, updated with your suggestion, can you give it a try?

imikushin:
LGTM

PR #1763, opened 2019-12-20
Added the missing RetrieveSnapshotDetails API in VSLM

lintongj:
The RetrieveSnapshotDetails API is necessary for the incremental backup of FCD or CNS block volume. Somehow, the API is not added initially.


lintongj:
Thanks @dougm. I am doing it.

PR #1776, opened 2020-01-09
Clarify System.Read privilege requirement for PortGroup backing

### LINKED ISSUE #1517: vm.create fails if read privilege on distributedVirtualSwitch isn't granted
Running `./govc vm.create -debug -net 'VDI Network 10.XXX.XXX.x' -pool 3-Low -ds XXXXX -folder XXXXXX vm01`
doesn't work. The error message is:
`./govc: no System.Read privilege on: DistributedVirtualPortgroup:dvportgroup--708.config.distributedVirtualSwitch`

If read permissions are granted, it works. 
Creating a virtual machine from WebGUI works without read permissions.

Shouldn't it work without read permission with govc?
If not, why is the read permission needed?

Thanks

dougm:
The error message sounded like a permission issue with the DVPG itself, rather than the Switch.
Attempt to make the error message more clear, add comments and vcsim based test.

Closes #1517

### DIFF
--- a/object/distributed_virtual_portgroup.go
+++ b/object/distributed_virtual_portgroup.go
@@ -50,9 +50,15 @@ func (p DistributedVirtualPortgroup) EthernetCardBackingInfo(ctx context.Context
 		return nil, err
 	}
 
+	// From the docs at https://code.vmware.com/apis/196/vsphere/doc/vim.dvs.DistributedVirtualPortgroup.ConfigInfo.html:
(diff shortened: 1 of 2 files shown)


dougm:
The error message now reads:

`
failed to create EthernetCardBackingInfo for /DC0/network/DC0_DVPG0: System.Read privilege required for config.distributedVirtualSwitch
`

cc @akutz @sheriefhammam

PR #1790, opened 2020-01-13
fix simulator disk manager fault message.

sigma:
This change makes the simulator behave more like a real vCenter in the presence
of missing VMKDs. In particular, it conforms to an expectation of VCP, which
allows to use the simulator with that k8s provider.


sigma:
@dougm I'm not sure, I provided the right legal contact email in the CLA form, hopefully that will get through

dougm:
Thanks again @sigma !

PR #1799, opened 2020-01-16
vcsim: add support for NSX backed networks

dougm:


simulator/model.go
```diff
@@ -71,6 +72,13 @@ type Model struct {
 	// Portgroup specifies the number of DistributedVirtualPortgroup entities to create per Datacenter
 	Portgroup int
 
+	// PortgroupNSX specifies the number NSX backed DistributedVirtualPortgroup entities to create per Datacenter,
```
bryanv: Nit: trailing `,`?

dougm: Fixed

fstrudel:
LGTM, thanks! One small comment: is there a README somewhere that explains the names of the objects we should expect? It's always confusing to me to find what objects are created by default.

dougm:
@fstrudel just some example output: https://github.com/vmware/govmomi/tree/master/vcsim#usage

I've been meaning to add a doc that explains the generated names, will do that in a separate PR.
btw the format came from the original 'vcsim' which was a mode within vCenter (removed in 6.0): https://www.virtuallyghetto.com/2014/03/quick-tip-adding-a-custom-prefix-to-your-vcsim-environment.html

PR #1813, opened 2020-01-25
Add the option of customizing the certificate verification method

annamooseity:
Some users request a slightly less secure verification that the normal TLS method

This could include, for example, verifying everything normally except for matching the hostname.

This allows the user to specify their own verification method (which they are then liable for)

### DIFF
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -159,6 +159,13 @@ func NewClient(u *url.URL, insecure bool) *Client {
 	return &c
 }
 
+type CertVerifyFunc func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
+
+func (c *Client) SetCertVerificationMethod(verifyMethod CertVerifyFunc) {
+	c.t.TLSClientConfig.InsecureSkipVerify = true
+	c.t.TLSClientConfig.VerifyPeerCertificate = verifyMethod
+}
+
 // NewServiceClient creates a NewClient with the given URL.Path and namespace.
 func (c *Client) NewServiceClient(path string, namespace string) *Client {
 	vc := c.URL()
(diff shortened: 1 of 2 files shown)


vim25/soap/client.go
```diff
@@ -159,6 +159,13 @@ func NewClient(u *url.URL, insecure bool) *Client {
 	return &c
 }
 
+type CertVerifyFunc func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
+
+func (c *Client) SetCertVerificationMethod(verifyMethod CertVerifyFunc) {
```
dougm: Thanks @annamooseity .  It is possible to set this field currently using:
```go
soap.Client.Transport.(*http.Transport).TLSClientConfig
```

If we were to add something else that helps simplify, how about a more general method such as:
```go
func (c *Client) DefaultTransport() *http.Transport {
   return c.t
}
```
This would give users simpler access to all Transport fields. What do you think?
There are a few other places in the repo that could use the helper as well:
```
govc/about/cert.go:102:	t := c.Client.Transport.(*http.Transport)
govc/flags/client.go:311:	if t, ok := sc.Transport.(*http.Transport); ok {
govc/flags/host_connect.go:78:			t := c.Transport.(*http.Transport)
session/manager.go:127:		c.Transport.(*http.Transport).TLSClientConfig.InsecureSkipVerify = true
vapi/library/library.go:127:				t := c.Transport.(*http.Transport)
vapi/library/library_item_updatesession_file.go:119:		if c.Transport.(*http.Transport).TLSClientConfig.InsecureSkipVerify {
```

annamooseity: I really like this! When I have time, I'd like to put up a new PR (or re-open this one, whichever you prefer) with your suggested changes.

dougm: Thanks @annamooseity , whichever is easier for you is fine with me!

PR #1820, opened 2020-02-04
Expose Soap client default transport 

annamooseity:
Now there is a "prettier" way to customize things like certificate verification (at your own risk)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # changelog
 
+### 0.22.2 (2020-02-04)
+
+* Expose soap client default transport
+
 ### 0.22.1 (2020-01-13)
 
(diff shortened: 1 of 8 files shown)


annamooseity:
@dougm On it. This may be a dumb question, but I poked around the Readme and Contributing files and couldn't figure out how to run the build/CO locally. Where/how can I run that? Or is it a github only thing?

dougm:
@annamooseity , that's a good question and something we need to add to the contrib file.

You can run `make` from the top-level to run both `check` (go fmt/vet) and `test`: https://github.com/vmware/govmomi/blob/master/Makefile#L4

Do you want to try it out and add to the contrib file? 

More info on the govc tests here: https://github.com/vmware/govmomi/tree/master/govc/test

dougm:
Also note that `make check` does not apply the gofmt / goimports fixes, it is expected the developer would do that (normally automatically done by Go editors).

PR #1829, opened 2020-02-09
vcsim: add guest operations support

dougm:
Minimal InitiateFileTransfer{To,From}Guest implementation to start,
depends on vcsim VM running a container.


govc/test/vm.bats
```diff
@@ -960,3 +960,39 @@ load test_helper
...
+
+  run govc vm.power -on $GOVC_VM
+  assert_success
+
+  if ! docker inspect "$name" ; then
+    flunk "$GOVC_VM container does not exist"
+  fi
+
+  run govc guest.upload README.md /tmp/README.md
+  assert_failure # unauthenticated
```
imikushin: Where does `/tmp/README.md` come from? Can the asserted failure actually be caused by a File Not Found?

2 lines below it gets successfully uploaded, apparently (the CI says `ok 205 guest fileops`), but still a bit of a mystery 🙂

imikushin:
Looking forward to Getting This Merged! 🤘

govc/test/vm.bats
dougm: The first arg `README.md` is the local [govc/test/README.md](https://github.com/vmware/govmomi/blob/master/govc/test/README.md).  We just need the container `/tmp` directory to exist.  We do need to handle more error cases (such as if /tmp didn't exist), but for now this just validates that auth is required.

See also: https://github.com/vmware/govmomi/blob/master/govc/USAGE.md#guestupload

PR #1837, opened 2020-02-13
Avoid ServiceContent requirement in lookup.NewClient

dougm:
As of d7430825, lookup.NewClient uses ServiceContent.Setting.
However, there is existing code that creates vim25.Client without
initializing ServiceContent, resulting in a nil pointer dereference.

The use of Setting was only needed for external PSC, make it optional
to avoid the NPE.


HakanSunay:
Ship it!

PR #1842, opened 2020-02-22
WIP: use interfaces for singleton access

### LINKED ISSUE #1841: simulator extension pattern doesn't work for singletons
The approach to simulator extension as described in `simulator/example_extend_test.go` doesn't work for a number of singletons in the system.
If nothing else, because the registry exposes convenience accessors for these that look like:
```
func (r *Registry) VirtualDiskManager() *VirtualDiskManager {
	return r.Get(r.content().VirtualDiskManager.Reference()).(*VirtualDiskManager)
}
```
so putting a wrapper object is doomed to fail.

I'm curious to know if:
- there's something I'm missing that would allow customizing, let's say, responses to calls such as `CreateVirtualDiskTask`
- there's an agreeable way of plugging that gap. In order not to break the API compatibility, I was thinking of adding for example a `CreateVirtualDiskTaskFunc` field to `VirtualDiskManager`, that would preempt the default implementation. It doesn't fit that well with the existing model tough.

sigma:
DO NOT MERGE
hacking in progress

This fixes #1841

### DIFF
--- a/lookup/simulator/registration_info.go
+++ b/lookup/simulator/registration_info.go
@@ -32,7 +32,7 @@ var (
 // The complete list can be captured using: govc sso.service.ls -dump
 func registrationInfo() []types.LookupServiceRegistrationInfo {
 	vc := simulator.Map.Get(vim25.ServiceInstance).(*simulator.ServiceInstance)
-	setting := simulator.Map.OptionManager().Setting
+	setting := simulator.Map.OptionManager().MO().Setting
 	sm := simulator.Map.SessionManager()
 	opts := make(map[string]string, len(setting))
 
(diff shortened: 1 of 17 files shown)


sigma:
@dougm this is not ready obviously, but do you mind having a look and let me know if that's the kind of change you'd consider for #1841 ?
Thanks !

sigma:
@dougm I'm not sure the internal calls are the important factor here. If we assume that we want overload Singleton using a CustomSingleton type to specialize its behavior, then the problem arises every time a Registry.Singleton() accessor is called as it would panic when attempting to return the CustomSingleton object.
So are you saying that you'd rather *remove* the accessor if it's not called internally?
If so I think only ViewManager isn't called internally.

dougm:
Sorry for the confusion @sigma . With EventManager, only PostEvent is called internally, with SearchIndex only FindByDatastorePath.. so I think we could limit the interface to just those methods called internally.

With others, such as UserDirectory and CustomFieldsManager, I wonder if there are custom use cases or can we leave them as-is until the need arises?  Even with SessionManager, we already have ways to make Login methods fail - not sure there are use cases for the other SessionManager methods.

Maybe we could just limit your initial PR to the use cases you have around VirtualDiskManager and evaluate the others as needed?

sigma:
sorry for the delay. #1910 is the VirtualDiskManager-only PR

PR #1853, opened 2020-03-02
Add Content Library subscriptions support

dougm:
Quoting the H5 UI:
> Subscriptions allow a bi-directional relationship between Publisher and Subscriber libraries.
> This enables you to synchronize content both from the Publisher (Publish) and from the Subscriber (Sync).

- vapi: add library/subscriptions bindings

- vapi: add PublishLibrary binding

- vcsim: add library/subscriptions support

- vcsim: add publish support

- govc: add '-pub' flag to library.create (enable publishing)

- govc: add Publication section to library.info

- govc: add library.publish command

- govc: add library.subscriber.{create,rm,ls,info} commands

### DIFF
--- a/govc/library/publish.go
+++ b/govc/library/publish.go
@@ -1,0 +1,80 @@
+/*
+Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package library
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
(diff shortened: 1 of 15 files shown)


govc/library/publish.go
```diff
@@ -0,0 +1,65 @@
...
+func (cmd *publish) Usage() string {
+	return "PATH [ID]..."
+}
+
+func (cmd *publish) Description() string {
+	return `Publish Content Library to subscribers.
+
+Examples:
+  govc library.publish /my-library
+  govc library.publish /my-library id1 id2`
```
rhockenbury: Might be worth clarifying that these are library subscription ids, rather than library item ids.

vapi/library/library.go
```diff
@@ -146,6 +200,18 @@ func (c *Manager) SyncLibrary(ctx context.Context, library *Library) error {
 	return c.Do(ctx, url.Request(http.MethodPost), nil)
 }
 
+// PublishLibrary publishes the library to specified subscriptions.
+// If no subscriptions are specified, then publishes the library to all its subscriptions.
+func (c *Manager) PublishLibrary(ctx context.Context, library *Library, subscriptions []string) error {
```
rhockenbury: This isn't part of my use case, but wanted to point out that it's also possible (through the vcenter console) to publish a specific library item to a specific subscription.  Might be useful for others to have this.  Having the blanket publish all library items to all subscriptions is the functionality that I'm after.

dougm: I had seen that sync supports per-item (`library.sync` makes use of that), but didn't see that for publish. I'll take another look before merging this.  Thanks for the review!

dougm: Found the per-item publish in the UI+API and updated the branch with support for that. Can see how that'd be useful.

PR #1877, opened 2020-03-22
Compute Policy support

dougm:
- Add vAPI REST client support

- Add govc compute.policy.{ls,rm.create} commands

- Add simulator support and bats tests

### DIFF
--- a/vapi/compute/policy.go
+++ b/vapi/compute/policy.go
@@ -1,0 +1,67 @@
+/*
+Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0.
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package compute
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/vmware/govmomi/vapi/internal"
+	"github.com/vmware/govmomi/vapi/rest"
+)
+
+// PolicyManager provides convenience methods to operate Compute Policy
+type PolicyManager struct {
+	*rest.Client
+}
+
+// NewPolicyManager creates a new PolicyManager with the given client
+func NewPolicyManager(client *rest.Client) *PolicyManager {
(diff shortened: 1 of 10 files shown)


vapi/compute/policy.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// PolicySpec represents the structure to create a compute policy.
+type PolicySpec struct {
+	Name        string `json:"name,omitempty"`
+	Description string `json:"description,omitempty"`
+	Capability  string `json:"capability,omitempty"`
+	HostTag     string `json:"host_tag,omitempty"`
+	VMTag       string `json:"vm_tag,omitempty"`
+}
```
shylasrinivas: Do we need 2 structs? Due to the use of omitempty, string Policy @ line 45 can be added to PolicySpec?

dougm: It could be yes, but 'PolicySpec' has more inputs (HostTag, VMTag, etc.) that will never be included as output with 'List'.  So while it'd be functional using a single type, it can be confusing that way.  While they have fields in common, the REST API documents them as different types.

vapi/compute/policy.go
```diff
@@ -0,0 +1,133 @@
...
+// Get returns information about a specific compute policy
+func (c *PolicyManager) Get(ctx context.Context, id string) ([]Policy, error) {
+	r := c.Resource(path.Join(internal.PolicyPath, id))
+	var res []Policy
+	return res, c.Do(ctx, r.Request(http.MethodGet), &res)
+}
+
+// ListCapabilities returns information about the compute policy capabilities available in this vCenter server.
+func (c *PolicyManager) ListCapabilities(ctx context.Context) ([]CapabilitySummary, error) {
+	// Name+Description fields are of type LocalizableMessage in 6.9, but string in 7.0
```
dougm: @shylasrinivas with this change, ListCapabilities works with both 6.9 and 7.0

vapi/compute/policy.go
dougm: Actually, 'Get' does return HostTag, VMTag, etc., its just 'List' that doesn't return those. I'll take another look, maybe it would be simpler to collapse into a single type as we had it originally.

vapi/compute/policy.go
shylasrinivas: @dougm thanks for making this change. We just had a team discussion and decided to support only 7.0 APIs in terraform. With that in mind, it is ok if these 2 fields are strings.

vapi/compute/policy.go
dougm: @shylasrinivas ok made this change and the others we discussed, can you review again?

shylasrinivas: lgtm, thanks!

PR #1880, opened 2020-03-23
vcsim: unique MAC address for VM NICs

### LINKED ISSUE #1879: VCSIM: all VM's NIC get the same MAC address.
I have cloned VM while adding two NICs. After successful clone operation the cloned VM had two NICs. However MAC addresses for both NICs were the same.
The problem appears to be in generateMAC() method (https://github.com/vmware/govmomi/blob/master/simulator/virtual_machine.go#L919)

dougm:
The generateMAC function was using the VM UUID for all NICs, so multiple NICs had the same MAC.
Add the device UnitNumber to make each MAC unique.

Fixes #1879


dougm:
cc @akrits

PR #1884, opened 2020-03-26
vapi: prefer header authn to cookie authn

dougm:
Cookie based authentication will be deprecated soon, switch to using
header based authentication by default.

Note: for vCenter versions that doesn't support vmware-use-header-authn,
rest.Client will continue to use cookie authn transparently via Go's
http.Client cookie.Jar

### DIFF
--- a/vapi/simulator/simulator.go
+++ b/vapi/simulator/simulator.go
@@ -363,6 +363,7 @@ func Decode(r *http.Request, w http.ResponseWriter, val interface{}) bool {
 
 func (s *handler) session(w http.ResponseWriter, r *http.Request) {
 	id := r.Header.Get(internal.SessionCookieName)
+	useHeaderAuthn := strings.ToLower(r.Header.Get(internal.UseHeaderAuthn))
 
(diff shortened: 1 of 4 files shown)


vapi/simulator/simulator.go
```diff
@@ -382,11 +383,13 @@ func (s *handler) session(w http.ResponseWriter, r *http.Request) {
 		id = uuid.New().String()
 		now := time.Now()
 		s.Session[id] = &rest.Session{User: user, Created: now, LastAccessed: now}
-		http.SetCookie(w, &http.Cookie{
-			Name:  internal.SessionCookieName,
-			Value: id,
-			Path:  rest.Path,
-		})
+		if useHeaderAuthn != "true" {
```
shylasrinivas: trying to understand when useHeaderAuthn would be false..It is set to true in line 147.

dougm: vapi/simulator is part of 'vcsim', which is not limited to govmomi clients.

PR #1895, opened 2020-04-01
Add SetTaskState SetTaskDescription UpdateProgress to object package

fdawg4l:
SetTaskState lets you set the state of the task.
SetTaskDescription sets the description of the task in the UI
UpdateProgress changes the progress bar of the task in the UI.

### DIFF
--- a/object/task.go
+++ b/object/task.go
@@ -65,8 +65,8 @@ func (t *Task) Cancel(ctx context.Context) error {
 	return err
 }
 
-// SetTaskState sets task state and optionally sets results or fault, as appropriate for state.
-func (t *Task) SetTaskState(ctx context.Context, state types.TaskInfoState, result types.AnyType, fault *types.LocalizedMethodFault) error {
+// SetState sets task state and optionally sets results or fault, as appropriate for state.
+func (t *Task) SetState(ctx context.Context, state types.TaskInfoState, result types.AnyType, fault *types.LocalizedMethodFault) error {
 	req := types.SetTaskState{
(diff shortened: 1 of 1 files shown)


object/task.go
```diff
@@ -64,3 +64,37 @@ func (t *Task) Cancel(ctx context.Context) error {
 
 	return err
 }
+
+// SetTaskState sets task state and optionally sets results or fault, as appropriate for state.
+func (t *Task) SetTaskState(ctx context.Context, state types.TaskInfoState, result types.AnyType, fault *types.LocalizedMethodFault) error {
```
dougm, line 69: Since this is a Task method, let's remove 'Task' from the name: `SetState` instead of `SetTaskState`

object/task.go
```diff
@@ -64,3 +64,37 @@ func (t *Task) Cancel(ctx context.Context) error {
...
+		State:  state,
+		Result: result,
+		Fault:  fault,
+	}
+	_, err := methods.SetTaskState(ctx, t.Common.Client(), &req)
+	return err
+}
+
+// SetTaskDescription updates task description to describe the current phase of the task.
+func (t *Task) SetTaskDescription(ctx context.Context, description types.LocalizableMessage) error {
```
dougm, line 81: Since this is a Task method, let's remove 'Task' from the name: `SetDescription` instead of `SetTaskDescription`

object/task.go
fdawg4l: You're so right.  Good call!

PR #1898, opened 2020-04-03
Refactor govc session persistence into session/cache package

### LINKED ISSUE #1810: Use existing REST session
Is there a way to save and re-use an existing session with the REST client? The vic library, and I'd like to use that same functionality with govmomi, but I can't find how its done (or if it can be).

### LINKED ISSUE #1873: Rest client - session persistence and re-use
This GH issue is for tracking session persistence and reuse changes needed for the rest client. Currently this logic is being built into clients that use rest client from govmomi. Having this support in govmomi will help clients that intend to make use of the rest client.

dougm:
- Add govc support for persistent REST sessions

- Add REST support to session.login command (including token auth)

Fixes #1873
Fixes #1810

### DIFF
--- a/session/cache/session.go
+++ b/session/cache/session.go
@@ -1,0 +1,353 @@
+/*
+Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/url"
+	"os"
+	"os/user"
+	"path/filepath"
+
+	"github.com/vmware/govmomi/session"
+	"github.com/vmware/govmomi/vapi/rest"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+// Client interface to support client session caching
+type Client interface {
+	json.Marshaler
+	json.Unmarshaler
+
+	Valid() bool
+	Path() string
+}
+
+// Session provides methods to cache authenticated vim25.Client and rest.Client sessions.
+// Use of session cache avoids the expense of creating and deleting vSphere sessions.
+// It also helps avoid the problem of "leaking sessions", as Session.Login will only
+// create a new authenticated session if the cached session does not exist or is invalid.
+// By default, username/password authentication is used to create new sessions.
+// The Session.Login{SOAP,REST} fields can be set to use other methods,
+// such as SAML token authentication (see govc session.login for example).
+type Session struct {
+	URL         *url.URL // URL of a vCenter or ESXi instance
+	Dir         string   // Dir defaults to "$HOME/.govmomi/sessions"
+	Insecure    bool     // Insecure param for soap.NewClient (tls.Config.InsecureSkipVerify)
+	Passthrough bool     // Passthrough disables caching when set to true
+
+	LoginSOAP func(context.Context, *vim25.Client) error // LoginSOAP defaults to session.Manager.Login()
+	LoginREST func(context.Context, *rest.Client) error  // LoginREST defaults to rest.Client.Login()
+}
+
+var (
+	home = os.Getenv("GOVMOMI_HOME")
+)
+
+func init() {
+	if home == "" {
+		dir, err := os.UserHomeDir()
+		if err != nil {
+			dir = os.Getenv("HOME")
+		}
+		home = filepath.Join(dir, ".govmomi")
+	}
+}
+
+// Endpoint returns a copy of the Session.URL with Password, Query and Fragment removed.
+func (s *Session) Endpoint() *url.URL {
+	if s.URL == nil {
+		return nil
+	}
+	p := &url.URL{
+		Scheme: s.URL.Scheme,
+		Host:   s.URL.Host,
+		Path:   s.URL.Path,
+	}
+	if u := s.URL.User; u != nil {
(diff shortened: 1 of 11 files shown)


dougm:
Note to reviewers: while the diff is large, the key part for review is the new cache.Session type and methods.  Much of the other changes are adapting govc's ClientFlag to use the new type.

cc @bill-rich @shylasrinivas

session/cache/session.go
```diff
@@ -0,0 +1,347 @@
...
+// such as SAML token authentication (see govc session.login for example).
+type Session struct {
+	URL         *url.URL // URL of a vCenter or ESXi instance
+	Dir         string   // Dir defaults to "$HOME/.govmomi/sessions"
+	Insecure    bool     // Insecure param for soap.NewClient (tls.Config.InsecureSkipVerify)
+	Passthrough bool     // Passthrough disables caching when set to true
+
+	LoginSOAP func(context.Context, *vim25.Client) error // LoginSOAP defaults to session.Manager.Login()
+	LoginREST func(context.Context, *rest.Client) error  // LoginREST defaults to rest.Client.Login()
+}
```
shylasrinivas, line 61: does this change they way rest client is consumed? Eg: https://github.com/vmware/govmomi/blob/08e92fb6463edf9e944643ec2f68ed2585c28a0a/vapi/tags/example_test.go#L33-L34

shylasrinivas: can you provide a sample?

dougm: It doesn't change `rest.NewClient`, that will continue to work as it always had.  The use of `cache.Session` is optional, added an example: https://github.com/vmware/govmomi/pull/1898/files#diff-83ca018f25a8e1b36bc2ce4fac1eba5eR32

session/cache/session.go
```diff
@@ -0,0 +1,353 @@
...
+	p := s.Endpoint()
+	p.Path = path
+
+	// Key session file off of full URI and insecure setting.
+	// Hash key to get a predictable, canonical format.
+	key := fmt.Sprintf("%s#insecure=%t", p.String(), s.Insecure)
+	return fmt.Sprintf("%064x", sha256.Sum256([]byte(key)))
+}
+
+func (s *Session) dir() string {
```
matthewmfrost, line 104: Considering [we are trying to match Terraform's persisted REST session location](https://github.com/vmware/govmomi/issues/1810#issuecomment-578351354). Should the persisted REST session not be stored in:

> ${HOME}/.govmomi/rest_sessions

As apposed to where the SOAP sessions are stored:

> ${HOME}/.govmomi/sessions

The [Terraform vsphere_provider seems to store them as such](https://www.terraform.io/docs/providers/vsphere/index.html#session-persistence-options). @dougm?

dougm: Thanks for reviewing @matthewmfrost .  We could use separate directories, but is it useful to have them separate?  Note that the path (`/soap` or `/rest`) is included in the hashed file name, so the file names will be unique within the same directory.

matthewmfrost: I completely agree that is doesn't gain that much having them separate directories. I am not sure why the [terraform-provider-vsphere](https://github.com/terraform-providers/terraform-provider-vsphere) chose to do it that way, other than a bit more clarity for session management debugging. 

But it would be **nice** to be consistent given that the two (govc and terraform) are often used together. The GOVC/Terraform session interoperability is really neat on the SOAP side!

dougm: Agreed, having them consistent is a must for sharing the session cache.
It's probably less trouble to change the govmomi side since this is a new package.  I'll go ahead and do that later today.

matthewmf: Great, looks good to me :)

session/cache/session.go
```diff
@@ -0,0 +1,353 @@
...
+
+// key is a digest of the URL scheme + username + host + Client.Path()
+func (s *Session) key(path string) string {
+	p := s.Endpoint()
+	p.Path = path
+
+	// Key session file off of full URI and insecure setting.
+	// Hash key to get a predictable, canonical format.
+	key := fmt.Sprintf("%s#insecure=%t", p.String(), s.Insecure)
+	return fmt.Sprintf("%064x", sha256.Sum256([]byte(key)))
```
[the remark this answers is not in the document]
dougm: #1237 switched to sha-256.. it isn't actually a security issue as we want the unique file name in a canonical format.  But, there are Go linters that complain about use of 'sha1'.

Regardless of that, the idea is Terraform would switch to using `cache.Session` as govc has, then both are using the same file name digest and load/save code moving forward. So as long as both are using the same directories, the session cache can be shared.

vapi/rest/client.go
```diff
@@ -63,6 +63,36 @@ func NewClient(c *vim25.Client) *Client {
 	return &Client{sc, ""}
 }
 
+type marshaledClient struct {
+	SoapClient *soap.Client
+	SessionID  string
+}
+
+func (c *Client) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
dougm: Right, they'll be the same if Terraform moves to using `session.Cache` instead of its own.
Also note that govmomi has switched to using header auth for REST, as cookie auth for REST will be deprecated in soon in vSphere.  See #1884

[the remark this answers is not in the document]
matthewmf: Interesting, I wasn't aware of the deprecation, thanks for letting me know.

Well at least the two comments raised are on the vsphere-provider's development radar and on record. If it does lag behind the latest release of GOVC, then I'll try and make sure its added to the [providers docs.](https://github.com/terraform-providers/terraform-provider-vsphere/blob/ad0ce41fc7ab8eb7926e881a19a5d4056aaa8fca/website/docs/index.html.markdown#govcterraform-session-interoperability)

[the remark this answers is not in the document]
dougm: I'd be ok switching back to sha1 so govc and TF can continue to share soap sessions.  Then after TF switches to using cache.Session (assuming it does), we can go back to 256 to make the linters happy.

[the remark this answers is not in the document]
matthewmf: Sounds like a reasonable compromise and a steady integration path 👍

matthewmf:
Good news! The vsphere-provider (>=1.17.3) [now uses govmomi's rest session persistence code](https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/CHANGELOG.md#1173-april-22-2020), meaning GOVC/vsphere-provider session interoperability works like a dream.

Thanks @bill-rich @dougm

dougm:
Great news @matthewmf @bill-rich , thanks for the update!

PR #1903, opened 2020-04-03
govc: avoid truncation in object.collect

### LINKED ISSUE #1893: govc object.collect -json -type m / name runtime.bootTime returns only 100 VMS
Hello,
I have several vcenter with 10000 VMs each. I would like to collect the uptime of all VMs quickly (e.g. by not dumping all information about every VM). The following command only returns 100 VMs:

govc object.collect -json -type m / name runtime.bootTime

Cheers,
      Thomas

dougm:
vCenter caps WaitForUpdates to 100 max, when doing so sets the Truncated field to true.
Check this field in object.collect to avoid truncating results.

Fixes #1893

### DIFF
--- a/govc/object/collect.go
+++ b/govc/object/collect.go
@@ -434,8 +434,8 @@ func (cmd *collect) Run(ctx context.Context, f *flag.FlagSet) error {
 	}
 
 	return cmd.WithCancel(ctx, func(wctx context.Context) error {
+		matches := 0
 		return property.WaitForUpdates(wctx, p, filter, func(updates []types.ObjectUpdate) bool {
-			matches := 0
 			for _, update := range updates {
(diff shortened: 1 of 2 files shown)


dougm:
@sithglan can you review?

lukasluecke:
@dougm Is there any chance we could get a release that includes this fix soon? The broken behaviours causes a lot of problems for me right now 🙈

dougm:
"soon" yes, but in the meantime can you build from source? https://github.com/vmware/govmomi/tree/master/govc#source

lukasluecke:
@dougm Yeah I should be able to, but obviously would prefer a "stable" release 😅

Thank you for the fix btw. 🙂

jgomsi:
Thank you so much for the fix! <3

PR #1914, opened 2020-04-07
Add non-null HostLicensableResourceInfo to HostSystem

bendavini:
HostSystem has a required field 'LicensableResource': https://code.vmware.com/apis/704/vsphere/vim.HostSystem.html#field_detail
This causes some clients that validate manditory fields with an error that might look like: 'Missing value for non-optional field resource'
The MOB view of a vanilla *real* ESXi host might look something like:
---------------------------------------------------------
| Name     | Type          | Value                      |
---------------------------------------------------------
| resource | KeyAnyValue[] | resource["numCpuPackages"] |
---------------------------------------------------------

This commit fills in this value with a sensible value that matches the NumCpuPackages specified in host_hardware_info,
and prevents clients from failing when encountering this resource.

### DIFF
--- a/simulator/esx/host_system.go
+++ b/simulator/esx/host_system.go
@@ -1740,8 +1740,8 @@ var HostSystem = mo.HostSystem{
 		CurrentEVCModeKey:  "",
 		Gateway:            (*types.HostListSummaryGatewaySummary)(nil),
 	},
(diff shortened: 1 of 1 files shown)


dougm:
@bendavini actually the goimports change doesn't look related to your change, I'll look into whats going on there.

PR #1916, opened 2020-04-07
govc: ignore ManagedObjectNotFound errors in 'find' command

### LINKED ISSUE #1908: govc: ServerFaultCode: The object 'vim.VirtualMachine:vm-2284' has already been deleted or has not been completely created
Hello,
I have a vCenter with several thousand VMs. When I run the following command:

> (ba) [~] govc find . -type m
govc: ServerFaultCode: The object 'vim.VirtualMachine:vm-2284' has already been deleted or has not been completely created

How to idenitfy the offending VM? Can govc just ignore the error and give me a list VMs it has?

Cheers,
        Thomas

dougm:
It is possible for objects to be deleted after ContainerView.Find has returned.
In which case the inventory path lookup can result in ManagedObjectNotFound.
Ignore objects in this case, instead of returning an error.

Fixes #1908

### DIFF
--- a/govc/object/find.go
+++ b/govc/object/find.go
@@ -31,6 +31,7 @@ import (
 	"github.com/vmware/govmomi/property"
 	"github.com/vmware/govmomi/view"
 	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/soap"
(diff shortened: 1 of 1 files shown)


dougm:
cc @sithglan

PR #1800, opened 2020-01-16
Update vim25/xml to Go v1.13.6

dougm:
Hasn't been updated since the original fork of 1.3.1.
Performance improvements noted in #1794

PR #1921, opened 2020-04-09
Default to separate session cache directories

dougm:
Follow up to #1898 - use the same defaults as Terraform for session file cache.

### DIFF
--- a/session/cache/session.go
+++ b/session/cache/session.go
@@ -18,7 +18,7 @@ package cache
 
 import (
 	"context"
-	"crypto/sha256"
+	"crypto/sha1"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -99,7 +99,7 @@ func (s *Session) key(path string) string {
 	// Key session file off of full URI and insecure setting.
 	// Hash key to get a predictable, canonical format.
 	key := fmt.Sprintf("%s#insecure=%t", p.String(), s.Insecure)
-	return fmt.Sprintf("%064x", sha256.Sum256([]byte(key)))
+	return fmt.Sprintf("%040x", sha1.Sum([]byte(key)))
 }
 
 func (s *Session) file(p string) string {
--- a/session/cache/session.go
+++ b/session/cache/session.go
@@ -52,7 +52,8 @@ type Client interface {
 // such as SAML token authentication (see govc session.login for example).
 type Session struct {
 	URL         *url.URL // URL of a vCenter or ESXi instance
-	Dir         string   // Dir defaults to "$HOME/.govmomi/sessions"
+	DirSOAP     string   // DirSOAP cache directory. Defaults to "$HOME/.govmomi/sessions"
+	DirREST     string   // DirREST cache directory. Defaults to "$HOME/.govmomi/rest_sessions"
 	Insecure    bool     // Insecure param for soap.NewClient (tls.Config.InsecureSkipVerify)
 	Passthrough bool     // Passthrough disables caching when set to true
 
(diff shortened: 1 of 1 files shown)


dougm:
cc @matthewmfrost @bill-rich

dougm:
@matthewmf also including the revert to sha1, so the soap session cache will remain compatible. Then we'll get rest cache compat when TF moves to using the govmomi package.  Sound good?

matthewmf:
Sounds and looks good to me, thanks for that! 

I have just started to look ahead at some related work on how to create a REST session using tokens (specifically holder-of-key tokens), as we have equivalent support for SOAP session creation. 

At present I am just struggling to get a bearer token working (see example attached), based on the [very limited examples](https://github.com/vmware/vsphere-automation-sdk-rest/blob/master/samples/postman/vSphere-Automation-REST-Resources.postman_collection.json#L103) provided by VMware. 

I don't suppose you have any insider knowledge or know where I can find more comprehensive documentation @dougm ?
[token-test.sh.zip](https://github.com/vmware/govmomi/files/4470774/token-test.sh.zip)

dougm:
@matthewmf thanks for reviewing.

govmomi's REST token auth impl is here: https://github.com/vmware/govmomi/blob/master/sts/signer.go#L231-L232

Note that you can use the token based `govc session.login` examples here: https://github.com/vmware/govmomi/blob/master/govc/USAGE.md#sessionlogin

With the new session.login `-r` flag: https://github.com/vmware/govmomi/blob/dbebdbea96a7fc9cb3bc7d7e25e57184e462e422/govc/session/login.go#L67

Used in the tests for example: https://github.com/vmware/govmomi/blob/dbebdbea96a7fc9cb3bc7d7e25e57184e462e422/govc/test/session.bats#L150-L157

dougm:
@matthewmf btw, it looks like TF's current rest_session format is just the session id itself.  So you should be able to convert a `govc session.login -r` to the TF format like so:
```sh
#!/bin/sh

govc session.login -r ...

user=$(govc env -x GOVC_USERNAME | sed -e s/@/%40/)
host=$(govc env -x GOVC_URL_HOST)
flag=$(govc env -x GOVC_INSECURE)

key=$(sha1sum <(echo -n "https://$user@$host/rest#insecure=$flag") | awk '{print $1}')

id=$(jq -r .SessionID ~/.govmomi/rest_sessions/$key)

echo -n "$id" > ~/.govmomi/rest_sessions/$key
```

dougm:
One other note, `govc session.login -l` will output the SOAP session ID without persisting the session file.  We could do similar for REST login if that simplifies this use case.

matthewmf:
Thanks for the tips, the code in _signer.go_ is what I was looking for...which I somehow missed.

With holder-of-key tokens you have to provide a cert and key of a solution user, which I assume would look something like this for REST:
`govc session.login -l -cert user.crt -key user.key -r -token "$(printf $hokToken administrator@vsphere.local)"`

....but that doesn't work:
`XML syntax error on line 1: unexpected EOF`

I suspect I have an incorrect command? 

.....

I was going to raise this separately and once I had time to figure out what's changed, but master seems to now have an issue with generating a holder-of-key token (still works in v0.22.1):
`hokToken=$(govc session.login -cert user.crt -key user.key -issue -token "$bearerToken")`

With the following error:
`govc: ns0:FailedCheck: Signature is invalid.
-:2: parser error : Start tag expected, '<' not found`

dougm:
You wouldn't use printf, that's just for the test token against vcsim: https://github.com/vmware/govmomi/blob/4966fff513584f0d4b454036a8435b6eba468936/govc/test/session.bats#L136-L137

You'd use the same commands to login soap w/ HoK, just need to add the `-r` flag.

Bearer and HoK are still working, but HoK+ActAs is indeed broken, due to something in #1800 - we have sts/client_test.go that reproduces, but that test doesn't run in CI (requires setting env to point at a real vCenter).  I'll look into it.

matthewmf:
Ah I see, sorry about that.

I have now got something working using master and 0.22.1. Thanks again for the pointers!

Issue #1186, opened 2018-07-24
Refactor generated code in the vim25 package


dougm:
# Background

The vim25/{types,methods,mo} packages are generated from vim.wsdl.  They provide complete coverage of the SOAP API, targeting the Go XML encoder/decoder as the primary consumer.
The various other packages on top of vim25 are intended to simplify the SOAP API and in more idiomatic Go way targeting developers as the primary consumer.
But these higher level govmomi types don't provide complete API coverage, so in some cases you'll see the vim25 package being used directly.

## [Managed Object Types](http://pubs.vmware.com/vsphere-6-5/topic/com.vmware.wssdk.apiref.doc/mo-types-landing.html)

vSphere inventory objects such as `VirtualMachine` and `HostSystem`, along with system services such as `PerformanceManager` and `SessionManager` are all referred to as managed objects.
These objects contain both properties and methods.  The properties can be accessed via the `PropertyCollector`.  The [vim25/mo](https://github.com/vmware/govmomi/tree/master/vim25/mo) package contains these types.

There are 133 mo types in the vim25 6.7 API.

## [Data Object Types](http://pubs.vmware.com/vsphere-6-5/topic/com.vmware.wssdk.apiref.doc/do-types-landing.html)

Data object types are used to compose managed object types and live in [vim25/types](https://github.com/vmware/govmomi/blob/master/vim25/types/types.go)

There are 2045 data types in the vim25 6.7 API.

## [Enumerated Types](http://pubs.vmware.com/vsphere-6-5/topic/com.vmware.wssdk.apiref.doc/enum-types-landing.html)

Defined in [vim25/types](https://github.com/vmware/govmomi/blob/master/vim25/types/enum.go).
All enum types are represented a string types, of which there are 352 in vim25 and a total of 1637 values defined as constants.

## [Fault Types](http://pubs.vmware.com/vsphere-6-5/topic/com.vmware.wssdk.apiref.doc/fault-types-landing.html)

Structured error types that live in the same file as the [Data types](https://github.com/vmware/govmomi/blob/master/vim25/types/types.go)

There are currently 691 fault types in the vim25 6.7 API.

## [Methods](http://pubs.vmware.com/vsphere-6-5/topic/com.vmware.wssdk.apiref.doc/methods-landing.html)

Defined in [vim25/methods](https://github.com/vmware/govmomi/blob/master/vim25/methods/methods.go).

[...]

PR #1922, opened 2020-04-09
Update to vSphere 7 APIs

dougm:


dougm:
Had wanted to address #1186 prior to this update, but that is unlikely to happen.
I any case, I plan to release 0.23 before merging this.

cc @shylasrinivas

Issue #1928, opened 2020-04-12
simulator: IP allocation for import.ova


matthewcosgrove:
I'm not seeing an IP allocated for an ova deployed to vcsim through `govc import.ova`. It's a PropertyMapping field so appreciate this is non-standard with wider design implications

```json
  "PropertyMapping": [
    {
      "Key": "ip0",
      "Value": "10.0.2.2",

```

Just wondering what it would take to get something like this working?

I can see that it's possible programmatically

```go
    spec := types.VirtualMachineConfigSpec{
        ExtraConfig: []types.BaseOptionValue{
            &types.OptionValue{Key: "SET.guest.ipAddress", Value: "10.0.0.42"},
        },
    }
```
according to the example in https://godoc.org/github.com/vmware/govmomi/simulator

dougm:
@matthewcosgrove possible with #1929 , using `SET.guest.ipAddress`.  If the key needs to be `ip0` for your use case, we can look at adding something for that too.

matthewcosgrove:
Thanks @dougm, yes the ova I'm using (Pivotal OpManager) uses `ip0` for the key as shown in the PropertyMapping excerpt above. Would be fantastic if you could add support for that in

PR #1929, opened 2020-04-13
vcsim: apply ExtraConfig after devices

dougm:
In order to set a VM guest net.config ipAddress, a VM ethernet card device must be present.
The original code worked fine for existing VMs, but we need to apply ExtraConfig after devices
for SET to work properly when creating new VMs.

Issue #1928

### DIFF
--- a/govc/test/vcsim.bats
+++ b/govc/test/vcsim.bats
@@ -142,6 +142,25 @@ EOF
 
   run govc vm.ip "$TTYLINUX_NAME"
   assert_success "10.0.0.42"
+
+  run govc vm.destroy "$TTYLINUX_NAME"
+  assert_success
+
+  govc import.ovf -options - "$GOVC_IMAGES/$TTYLINUX_NAME.ovf" <<EOF
+{
+  "PropertyMapping": [
+    {
+      "Key": "ip0",
+      "Value": "10.0.0.43"
+    }
+  ],
+  "PowerOn": true,
+  "WaitForIP": true
+}
+EOF
+
+  run govc vm.ip "$TTYLINUX_NAME"
+  assert_success "10.0.0.43"
 }
 
 @test "vcsim vm.create" {
--- a/govc/test/vcsim.bats
+++ b/govc/test/vcsim.bats
@@ -126,6 +126,22 @@ load test_helper
 
   run govc object.collect -s $vm config.uuid
   assert_success "$uuid"
+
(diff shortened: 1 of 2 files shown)


matthewcosgrove:
I have tested against an OpsManager ova by manually changing the key to `SET.guest.ipAddress` on the generated spec before importing and it LGTM, works great, thanks!

As mentioned in #1928 would be really useful to also have the key `ip0` work out the box

dougm:
Would like to understand the `ip0` use case better, is it used in a production ova you're shipping to other folks? Or just for testing?  Either way, how is it used when deployed to a real vCenter?

matthewcosgrove:
The `ip0` key is baked into the Pivotal/Tanzu OpsManager product ova which we automate rolling out and upgrading via IaC pipelines. 

Manual steps can be seen here: https://docs.pivotal.io/platform/ops-manager/2-9/vsphere/deploy.html
`ip0` corresponds to the section talking about "IP Address: The IP address of the Ops Manager network interface"
As long as I can remember the Pivotal OpsManager ova has used the `ip0` key for it's `PropertyMapping` of the IP address so it's fair to say it's standard configuration for that particular product.

Our automation scripts are very similar to those shown [here](https://github.com/pivotal-cf/pcf-pipelines/blob/master/install-pcf/vsphere/tasks/import-opsman/task.sh#L20) that were developed at Pivotal (link goes direct to the specific line with `ip0` being configured) 

The key enabler with vcsim is that we can write proper IaC "unit" tests that can be run locally and test the logic in our scripts which manage the ova deployment. For example, logic that ensures the pipeline fails fast if the IP is already in use.

dougm:
@matthewcosgrove ok, pushed a commit that maps `ip0` to `SET.guest.ipAddress`.  We could also add mappings for the others, seems like a reasonable convention to adopt.

matthewcosgrove:
I have retested both scenarios against my unit tests. 

**Set Up**
For future reference, as my project uses go modules, to test the fork I simply updated my `go.mod` as follows

```
go 1.14

require github.com/vmware/govmomi master

// only required until https://github.com/vmware/govmomi/pull/1929 merged into master
replace github.com/vmware/govmomi => github.com/dougm/govmomi issue-1928
```
(as expected the branches are resolved to exact commit sha identifiers in `go.mod` on build)

**Test results**

- Workaround with `SET.guest.ipAddress`: **PASS**
- No workaround using the provided `ip0`: **PASS**

This is awesome work, thanks @dougm

dougm:
@matthewcosgrove good stuff, merged!

PR #1930, opened 2020-04-13
vcsim: fixes for PowerCLI

dougm:
Commit 08adb5d6 (#1740) removed the obsolute sso:sts type in favor of cs.identity,
but PowerCLI still uses sso:sts.

Commit 43d69860 (#1595) changed endpoint registration to happen before vcsim's TLS cert is known,
leaving ServiceEndpoints.SslTrust empty, but PowerCLI requires this field be non-empty.

### DIFF
--- a/govc/test/sso.bats
+++ b/govc/test/sso.bats
@@ -27,6 +27,13 @@ load test_helper
   run govc sso.service.ls -t cs.identity -U
   assert_success "$sts"
 
+  run govc sso.service.ls -t sso:sts -U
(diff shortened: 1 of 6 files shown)


mattmoor:
I'm going to pull this fix in and test it with our PowerCLI sample now.  thanks!

dougm:
Thanks for testing it @mattmorr , I hadn't tried `Get-VIEvent` before.  We just needed to implement the `ResetCollector` method, pushed a new commit with that.

PR #1932, opened 2020-04-15
Fix delegated Holder-of-Key token signature

dougm:
The vim25/xml sync in #1800 broke delegated HoK signatures,
due to an empty AttributeValue.Type in the RequestSecurityToken body.

Make use of the 'typeattr' field flag to ensure the Type field is
properly marshaled in this case.

### DIFF
--- a/sts/internal/types.go
+++ b/sts/internal/types.go
@@ -473,7 +473,7 @@ func (a *AttributeStatement) C14N() string {
 
 type AttributeValue struct {
 	XMLName xml.Name
-	Type    string `xml:"type,attr"`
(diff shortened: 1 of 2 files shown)


dougm:
@matthewmf this fixes the issue we discussed in #1921

matthewmf:
Good stuff! I gave it a test and it works as expected. Overall, it looks like we are in a good place GOVC/vsphere-provider session interpolation wise.

Thanks @dougm!

dougm:
Excellent, thanks @matthewmf !

PR #1933, opened 2020-04-15
Avoid possible nil pointer dereference in guest TransferURL

dougm:
Avoid using host.Config if nil, SDK doc says "This might not be available for a disconnected host."

### DIFF
--- a/guest/file_manager.go
+++ b/guest/file_manager.go
@@ -18,6 +18,7 @@ package guest
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/url"
 	"sync"
@@ -150,7 +151,7 @@ func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error
 	c := property.DefaultCollector(m.c)
 
 	var vm mo.VirtualMachine
-	err = c.RetrieveOne(ctx, m.vm, []string{"runtime.host"}, &vm)
+	err = c.RetrieveOne(ctx, m.vm, []string{"name", "runtime.host"}, &vm)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


guest/file_manager.go
```diff
@@ -167,6 +167,11 @@ func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error
 		return nil, err
 	}
 
+	if host.Config == nil {
+		// SDK doc says "This might not be available for a disconnected host."
+		return turl, nil
```
mikekim, line 172: Shouldn't this return an error? How useful is `turl` in this case?

dougm: If the host is actually running in this case, turl can still work.  host.Config is used for 2 things:

1) The hostname in `turl` is valid in most cases, but we try here to "prefer an ESX management IP" as there have been a few cases where the hostname cannot be resolved (DNS) depending on where the client code is running.

2) soap.Client.SetThumbprint - the client may already trust the host's cert, have already saved its thumbprint via a previous call or have `InsecureSkipVerify=true`

mikekim: Thank you for explaining this.

dougm: Added the comments above and more to the code itself, do you mind reviewing the comments?

guest/file_manager.go
```diff
@@ -167,6 +167,24 @@ func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error
...
+		// SDK doc says "This might not be available for a disconnected host."
+		// We can't be sure here, since regardless of vCenter's HostSystem object state,
+		// the host may indeed be running and likely is running otherwise vm.Runtime.Host
+		// would nil and we return earlier (above) in that case.
+		// Worst case if the host is not running: Go's http.Client returns an error clearly indicating that.
+		// Assuming the host is running, turl is most likely valid, so return it as-is.
+		// host.Config is used below for 2 things:
+		// 1) There have been a few cases where the hostname cannot be resolved due to DNS config,
+		//    so we prefer an IP.
+		//    Worst case: http.Client returns a DNS lookup error
```
mikekim, line 180: I actually think this may be a problem since likely it'd send me searching for the problem in the DNS configuration and asking why suddenly the host can't be resolved if it worked before...
As opposed to an error saying "host config is not set, check if the host is connected".

dougm: The DNS issue is rare, and not an issue when vpxd and the client (e.g. wcpsvc) have the same DNS config.  But I don't mind changing it to return an error, as this is the first time seeing `host.Config` being nil in this code path that I'm aware of.

dougm: updated to return an error, let me know if you want any changes to the format:

**guest TransferURL failed for vm "DC0_H0_VM0" (VirtualMachine:vm-57): host "DC0_H0" (HostSystem:host-21) config==nil, connectionState==connected**

Also added a test, separate commit as vcsim only added GuestOperationsManager support recently.

mikekim: I just think it may be misleading because this function may give different non-error results for the same host now that may or may not work. When we're dealing with logs from customers there's no way for us to inspect the environment and it may be misleading to see DNS error so I'd prefer to be explicit.

guest/file_manager.go
```diff
@@ -159,14 +160,25 @@ func (m FileManager) TransferURL(ctx context.Context, u string) (*url.URL, error
...
+	}
 
 	var host mo.HostSystem
 	err = c.RetrieveOne(ctx, *vm.Runtime.Host, props, &host)
 	if err != nil {
 		return nil, err
 	}
 
+	if host.Config == nil {
+		return nil, fmt.Errorf("guest TransferURL failed for vm %q (%s): host %q (%s) config==nil, connectionState==%s",
```
mikekim, line 177: Appreciate logging `connectionState` here, thanks!

PR #1963, opened 2020-05-04
fix a few data races when creating VMs concurrently

sigma:
This PR fixes 3 data races that manifest themselves when creating VMs concurrently.

### DIFF
--- a/simulator/folder.go
+++ b/simulator/folder.go
@@ -336,20 +336,22 @@ func (c *createVM) Run(task *Task) (types.AnyType, types.BaseMethodFault) {
 	vm.ResourcePool = &c.req.Pool
 
 	if c.req.Host == nil {
-		var hosts []types.ManagedObjectReference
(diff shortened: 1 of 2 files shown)


sigma:
@dougm I've been wondering about that too.
I *think* that might be because the race test uses a vmomi *client* to trigger the concurrent operations. So that the data race detector watches the client and the server independently, and in the middle the fact that answering those requests is pretty quick means that there's actually no concurrency (normally) on the server side.
Just a working theory though

dougm:
Yes, but it does start a simulator within in the same process / go runtime, and has caught other races in the simulator package. Just realizing, maybe because the `Datacenter.VmFolder` is locked while creating the VM. Since the test uses the same folder, it avoids concurrent read/write of the fields referenced within.  Guessing if we changed the test to use different folders we'd have seen the race prior to your change.

PR #1989, opened 2020-05-22
govc: ipath search flag does not require a Datacenter

dougm:
The SearchIndex methods all require a Datacenter param,
except for FindByInventoryPath.


floragj:
Looks like it works to me! Thanks for addressing this.

dougm:
Thanks for reviewing and testing @floragj !

Issue #1991, opened 2020-05-25
Simulator CreateSnapshotTask does not return snapshot ref


jmpfar:
Hey,
I've taken a snapshot in the simulator and then tried to get the result from the task. I've verified that vSphere returns the snapshot mo-ref in the task result (see [here](https://pubs.vmware.com/vi3/sdk/ReferenceGuide/vim.VirtualMachine.html#createSnapshot)).
Unfortunately the simulator always returns nil.

https://github.com/vmware/govmomi/blob/61b30e20be49709237d0acf2d41a1e72fcd493ce/simulator/virtual_machine.go#L1897

See snapshot task code above, the line returns nil, compare with clone vm: https://github.com/vmware/govmomi/blob/61b30e20be49709237d0acf2d41a1e72fcd493ce/simulator/virtual_machine.go#L1689

dougm:
Looks like changing line 1897 to return `snapshot.Self, nil` would fix that. Can you give it try?

PR #1992, opened 2020-05-26
vcsim: CreateSnapshotTask now returns moref in result

jmpfar:
simulator CreateSnapshotTask should return the created snapshot mo-ref on success,
similar to vSphere behavior
issue #1991


jmpfar:
thanks for the help! :)

PR #1994, opened 2020-05-26
Add appliance log forwarding config handler and govc verb

fdawg4l:
Appliance logging config can be retrieved via the VAPI REST
endpoint.  The govc output is as follows.

```
$ govc appliance.logging.forwaring
Hostname:  10.186.54.118
Port:      5480
Protocol:  TLS
Hostname:  2.2.2.2
Port:      9999
Protocol:  TLS
```

### DIFF
--- a/govc/appliance/logging/forwarding.go
+++ b/govc/appliance/logging/forwarding.go
@@ -28,17 +28,13 @@ import (
 	vlogging "github.com/vmware/govmomi/vapi/appliance/logging"
 )
 
-func main() {
-	fmt.Println("vim-go")
-}
-
 type forwarding struct {
 	*flags.ClientFlag
 	*flags.OutputFlag
 }
 
 func init() {
-	cli.Register("appliance.logging.forwarding", &forwarding{})
+	cli.Register("vcsa.log.forwarding.info", &forwarding{})
 }
 
 func (cmd *forwarding) Register(ctx context.Context, f *flag.FlagSet) {
@@ -59,15 +55,11 @@ func (cmd *forwarding) Process(ctx context.Context) error {
 	return nil
 }
 
-func (cmd *forwarding) Usage() string {
-	return ""
-}
-
 func (cmd *forwarding) Description() string {
 	return `Retrieve the VC Appliance log forwarding configuration
 
 Examples:
-  govc appliance.logging.forwarding`
+  govc vcsa.log.forwarding.info`
 }
 
 func (cmd *forwarding) Run(ctx context.Context, f *flag.FlagSet) error {
@@ -76,9 +68,9 @@ func (cmd *forwarding) Run(ctx context.Context, f *flag.FlagSet) error {
 		return err
 	}
 
-	fwd := vlogging.NewForwarding(c)
+	fwd := vlogging.NewManager(c)
 
-	res, err := fwd.Config(ctx)
+	res, err := fwd.Forwarding(ctx)
 	if err != nil {
 		return nil
 	}
--- a/govc/appliance/logging/forwarding.go
+++ b/govc/appliance/logging/forwarding.go
@@ -1,0 +1,101 @@
+/*
+Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logging
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"text/tabwriter"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	vlogging "github.com/vmware/govmomi/vapi/appliance/logging"
+)
+
+func main() {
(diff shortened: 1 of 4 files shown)


govc/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,101 @@
...
+	"io"
+	"text/tabwriter"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	vlogging "github.com/vmware/govmomi/vapi/appliance/logging"
+)
+
+func main() {
+	fmt.Println("vim-go")
```
dougm, line 32: s/vim/emacs/

govc/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type forwarding struct {
+	*flags.ClientFlag
+	*flags.OutputFlag
+}
+
+func init() {
+	cli.Register("appliance.logging.forwarding", &forwarding{})
```
dougm, line 41: Should we want to add support for `set` or `test`, how about naming this `appliance.logging.forwarding.info` ?
Also wondering if we should shorten `appliance` to `vcsa` or even `vcenter`. And `logging` to `log` ?
`vcsa.log.forwarding.info` ?

govc/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,101 @@
...
+		return err
+	}
+	if err := cmd.OutputFlag.Process(ctx); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (cmd *forwarding) Usage() string {
+	return ""
```
dougm, line 63: You can just omit this func for commands that don't have a usage arg.

vapi/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,56 @@
...
+package logging
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/vmware/govmomi/vapi/rest"
+)
+
+const ApplianceLoggingForwardingPath = "/appliance/logging/forwarding"
```
dougm, line 26: Other than `rest.Path`, we haven't been exporting path constants.

vapi/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Forwarding provides convenience methods to configure appliance logging forwarding.
+type Forwarding struct {
+	*rest.Client
+}
+
+// NewForwarding creates a new Forwarding with the given client
+func NewForwarding(client *rest.Client) *Forwarding {
```
dougm, line 34: While there's only a few govmomi/vapi packages, we've been doing a package level `Manager` type + `func NewManager`. I think you could do that here too?

vapi/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,56 @@
...
+	Port     int    `json:"port,omitempty"`
+	Protocol string `json:"protocol,omitempty"`
+}
+
+func (f *Forwarding) getForwardingResource() *rest.Resource {
+	return f.Resource(ApplianceLoggingForwardingPath)
+}
+
+// Config returns all logging forwarding config.
+func (f *Forwarding) Config(ctx context.Context) ([]Config, error) {
```
dougm, line 52: And with the comment above, this'd become:
```go
func (m *Manager) Forwarding(ctx context.Context) ([]Forwarding, error) {
```
And if needed down the road:
```go
func (m *Manager) SetForwarding(...)
```

govc/appliance/logging/forwarding.go
fdawg4l: Oh man.  :facepalm: Will remove.

govc/appliance/logging/forwarding.go
fdawg4l: I like `vcsa.log.forwarding.info`.  

Will try adding a setter.  Standby.

vapi/appliance/logging/forwarding.go
fdawg4l: Will make this package private.

govc/appliance/logging/forwarding.go
dougm: No need to add a setter until it is needed, I was just looking at what else the API can do.

vapi/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,56 @@
...
+
+// NewManager creates a new Manager with the given client
+func NewManager(client *rest.Client) *Manager {
+	return &Manager{
+		Client: client,
+	}
+}
+
+// Config represents configuration for log message forwarding.
+type Config struct {
```
dougm, line 41: Can we rename this `Forwarding` ?

govc/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,93 @@
...
+	vlogging "github.com/vmware/govmomi/vapi/appliance/logging"
+)
+
+type forwarding struct {
+	*flags.ClientFlag
+	*flags.OutputFlag
+}
+
+func init() {
+	cli.Register("vcsa.log.forwarding.info", &forwarding{})
```
dougm, line 37: Do you mind changing the file path to match? `vcsa/log/forwarding/info.go` instead of `appliance/logging/forwarding.go` ?

govc/appliance/logging/forwarding.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+	"io"
+	"text/tabwriter"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	vlogging "github.com/vmware/govmomi/vapi/appliance/logging"
+)
+
+type forwarding struct {
```
dougm, line 31: Can you name this type `info`? (and the package "forwarding")

PR #1997, opened 2020-05-28
govc: support raw object references in import.ova NetworkMapping

### LINKED ISSUE #1996: `import.ova` does not work with NSXT 3.0 and  [distribute switch name] / [port group name]
Our current network is setup with switches with identical named port groups in each switch.
The `ipath` returned from the `govc ls` return the same `ipath` for different resources.
For example,

```
$ govc ls -l /some-ds-123/network

/some-ds-123/network/pas-edc2-eng-x-01-ha-proxy (DistributedVirtualPortgroup)
/some-ds-123/network/pas-edc2-eng-x-01-ha-proxy (DistributedVirtualPortgroup)
...
```

When specifying the `NetworkMapping` in the `options.json`, of one of these port groups an error message is returned.

```
$ govc import.ova
govc: Host did not have any virtual network defined.
exit status 1
```

We tried, then, to use a more unique identifier as document in this [terraform comment](https://github.com/terraform-providers/terraform-provider-vsphere/issues/349#issuecomment-412061458).

This also does not work.

After discussion, it appears that there current no way to uniquely identify a port group within NSXT with `import.ova`.
The MO reference (a unique identifier for resources in vSphere) could be use to identify the port group.

Looking up an MO can be done by:

```
# print MO ref(s) of port group with the same name
% govc find -i / -name $pg_name
DistributedVirtualPortgroup:dvportgroup-1117
DistributedVirtualPortgroup:dvportgroup-1116
# find MO ref of the switch
ref=$(govc find -i / -type VmwareDistributedVirtualSwitch -name $switch_name)
# print MO ref of the port group on the specific switch
% govc find -i / -type DistributedVirtualPortgroup -name $pg_name -config.distributedVirtualSwitch $ref
DistributedVirtualPortgroup:dvportgroup-1117
```

Related issue:
* https://github.com/terraform-providers/terraform-provider-vsphere/issues/349

dougm:
vCenter doesn't allow duplicate port group names within the network folder when creating via the UI
or the AddDVPortgroupTask() API. However, NSXt 3.0 can do so via other (unknown) means. With this
condition, govc would fail to map the given network name to its managed object reference.
To workaround this, a raw MO ref can now be used in the NetworkMapping, instead of the Network name
or inventory path.

Related history:
223168f0 * Add NetworkMapping section to importx options.
e3c3cd0a * Populate network mapping from ovf envelope (#546)

Fixes #1996

### DIFF
--- a/govc/importx/ovf.go
+++ b/govc/importx/ovf.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"path"
 
+	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
 	"github.com/vmware/govmomi/nfc"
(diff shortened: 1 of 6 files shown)


dougm:
@jtarchie can you review?

jtarchie:
I don't know enough to review the code. We are going test it against a live env to see if it resolves our issue.

datianshi:
So user has to find the raw MO first? Is that possible this can be done by reference "distributed_switch/port_group_name" Only port group name is unique, while the distributed switch is unique

dougm:
@datianshi yes, see examples in #1996

In general, the inventory paths used by govmomi/govc are generic to the vCenter inventory hierarchy (https://blogs.vmware.com/vsphere/2012/05/introduction-to-the-vsphere-api-part-3-inventory-hierarchy.html)

The paths are composed using parent/child relationships, regardless of type.  The catch here is that switches and port groups are _siblings_ in the vCenter inventory. It is possible we could make "distributed_switch/port_group_name" work in the long term, but the fallback to moref in this PR is more straightforward for the short term.

dougm:
@datianshi @jtarchie looking at this closer, I think we can make "distributed_switch/port_group_name"  work without too much trouble.  The native vSphere `SearchIndex` doesn't support inventory paths with this structure, but I think govmomi's Finder can adapt to this cleanly. Trying it out now and will update soon.

dougm:
Added a new commit to support "distributed_switch/port_group_name". Using raw moref is still an option, but not required when switch name is included.

PR #2007, opened 2020-06-07
Use correct var type to fix tpm attestation report

arijit8972:
Adding this to get entire pcr digest and event digest.

### DIFF
--- a/vim25/types/types.go
+++ b/vim25/types/types.go
@@ -21930,7 +21930,7 @@ type HostDigestInfo struct {
 	DynamicData
 
 	DigestMethod string `xml:"digestMethod"`
-	DigestValue  []byte `xml:"digestValue"`
+	DigestValue  []int `xml:"digestValue"`
 	ObjectName   string `xml:"objectName,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


arijit8972:
Yes, encoding/xml has the same issue. I tested it out in go playground. For some reason, xml unmarshaller ignores the errors and typecasts it anyways, not sure if that is by design. I am not sure if changing it for all is a good option, because it might break the structs where it actually needs a []byte type. 
Fixing the xml decoder might be a better option. Also, just letting you know, this is also an issue with certificates. I faced the same issue while fetching the TLS certs for the ESXi hosts. The fix is similar but not part of this pull request. But if the files are auto generated, this fix won't work right ?

dougm:
Since it can be reproduced with Go's encoding/xml, I think we should start by putting together a small test case to reproduce the issue. Something that doesn't depend on any of the govmomi code. Then we can use that test case to help find a solution, either with a custom type or a change to encoding/xml itself.

arijit8972:
Hey Doug,

Just checking. Is there any update on this bug ?

dougm:
Hey @arijit8972 , I don't have any updates. Can you share the go playground test case you used to reproduce?

Issue #1054, opened 2018-03-02
vSAN management API support


dougm:
The vSAN management API has its own URL endpoint and vmodl extension.  To access via govmomi, we'll need to generate the managed objects, types and methods similar to the [pbm package](https://github.com/vmware/govmomi/tree/master/pbm) that was added in PR #711

The structure should be similar to that of the pbm package:
```
vsan/client.go
vsan/types/
vsan/methods/
vsan/mo/
```
Unlike the PBM vmodl, the vSAN management vmodl includes the entire vim25 vmodl, so we can't use the existing [code generator](https://github.com/vmware/govmomi/blob/master/gen) as we did with PBM.

pnhowe:
When is vSAN support expected to be added?  I have a monitoring project that could really use this.

dougm:
@shuklanirdesh82 is currently looking into this.

pnhowe:
here is a basic python snippet of the direction I'm trying to go.

`from pyVim.connect import SmartConnect, Disconnect
from pyVmomi import vim, SoapStubAdapter
import vsanmgmtObjects
import ssl
import sys

VSAN_API_VC_SERVICE_ENDPOINT = '/vsanHealth'
VSAN_API_ESXI_SERVICE_ENDPOINT = '/vsan'
VERSION = 'vim.version.version10'

hostname = 'XXXXXX'
user = 'XXXXXX'
password = 'XXXXXX'

context = ssl._create_unverified_context()
si = SmartConnect(host=hostname, user=user, pwd=password, port=443, sslContext=context)

content = si.RetrieveContent()

aboutInfo = content.about

if aboutInfo.apiType == 'VirtualCenter':
  majorApiVersion = aboutInfo.apiVersion.split('.')[0]
  if int(majorApiVersion) < 6:
     print('The Virtual Center with version {0} (lower than 6.0) is not supported.'.format(aboutInfo.apiVersion))
     sys.exit(1)

  vsanStub = SoapStubAdapter(host=hostname, path=VSAN_API_VC_SERVICE_ENDPOINT, version=VERSION, sslContext=context)
  vsanStub.cookie = si._stub.cookie
  
  vpm = vim.cluster.VsanPerformanceManager( 'vsan-performance-manager', vsanStub )
  vhs = vim.cluster.VsanVcClusterHealthSystem( 'vsan-cluster-health-system', vsanStub )

else:
  majorApiVersion = aboutInfo.apiVersion.split('.')[0]
[...]

PR #2019, opened 2020-06-15
Add vSAN 7.0 bindings

gh05tn0va:
Issue #1054
This patch exports all vSAN public APIs to govmomi/vsan70, as we need to keep compatibility with exsisting vsan packages.
Bindings are generated by wsdl emitter and duplicate types of govmomi/vim25/types are excluded.

### DIFF
--- a/vsan70/methods/methods.go
+++ b/vsan70/methods/methods.go
@@ -1,0 +1,3984 @@
+/*
+Copyright (c) 2014-2020 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package methods
+
+import (
+	"context"
+
+	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vsan70/types"
+)
+
+type CnsAttachVolumeBody struct {
+	Req    *types.CnsAttachVolume         `xml:"urn:vsan CnsAttachVolume,omitempty"`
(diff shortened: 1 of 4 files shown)


divyenpatel:
@gh05tn0va 

I looked changes made in this PR. I have following questions.

1. Are we going to add internal APIs here in vsan70?
2. Are we going to add the vSAN client for these binding? I see we do not have a client here in this PR.
3. We are using vSAN binding located at https://github.com/vmware/govmomi/tree/master/vsan, if we need to add new internal API, Can we add new API there?
4. I see we have CNS bindings is located at https://github.com/vmware/govmomi/tree/master/cns. With this change, CNS bindings will be available here in vsan70 as well. What will be guidelines for maintaining them? at two different locations? 

cc: @SandeepPissay  @baludontu

gh05tn0va:
> @gh05tn0va
> 
> I looked changes made in this PR. I have following questions.
> 
> 1. Are we going to add internal APIs here in vsan70?
> 2. Are we going to add the vSAN client for these binding? I see we do not have a client here in this PR.
> 3. We are using vSAN binding located at https://github.com/vmware/govmomi/tree/master/vsan, if we need to add new internal API, Can we add new API there?
> 4. I see we have CNS bindings is located at https://github.com/vmware/govmomi/tree/master/cns. With this change, CNS bindings will be available here in vsan70 as well. What will be guidelines for maintaining them? at two different locations?
> 
> cc: @SandeepPissay @BaluDontu

1. No. We are not adding internal APIs here. We can use internal projects to track them.
2. This is a good proposal. I'll add this for sure.
3. Since this PR could break existing codes, my proposal is to keep both vsan and vsan70 bindings. Users can switch to vsan70 if possible. And we can deprecate old packages later.
4. Same. I only intend to maintain vsan70. What do you think?

dougm:
What would break? I'd very much prefer a single `vsan` package, rather than have duplication in a separate `vsan70` package.

divyenpatel:
@gh05tn0va 

> No. We are not adding internal APIs here. We can use internal projects to track them.


vSphere CSI driver is already using some internal vSphere APIs for example extendVolume. We will need binding for internal APIs as well to consume auto-generated binding from a single location. After this PR if we delete bindings located at https://github.com/vmware/govmomi/tree/master/cns it will break the driver.


> Same. I only intend to maintain vsan70. What do you think?

Same go bindings in multiple folders can confuse users. If bindings for CNS APIs is included as part of vSAN, we should remote CNS directory. CNS APIs we will also need bindings for internal APIs before we can remove CNS directory.

Issue #2018, opened 2020-06-15
VSAN Performance Data API


ysobale:
Enhance existing vsan api to fetch vsan performance data

PR #2021, opened 2020-06-15
Vsan Performance Data Collection API

ysobale:
Added VSAN Performance data collection API
https://github.com/vmware/govmomi/issues/2018


vsan/types/types.go
```diff
@@ -1199,7 +1199,7 @@ func init() {
 type VsanPerfQueryPerfRequestType struct {
 	This       types.ManagedObjectReference  `xml:"_this"`
 	QuerySpecs []VsanPerfQuerySpec           `xml:"querySpecs"`
-	Cluster    *types.ManagedObjectReference `xml:"cluster,omitempty"`
+	Cluster    types.ManagedObjectReference `xml:"cluster,omitempty"`
```
dougm, line 1202: The `Cluster` param is optional and this struct is generated from the API wsdl. Rather than change this, you can pass a pointer as the param (e.g. `&cluster` above)

ysobale: I had incorporated the feedback in and kept it as a original. Can you please check again and approve pull request.

ysobale:
Thanks @dougm

PR #2052, opened 2020-07-08
Honoring the instance uuid provided in spec by caller

parveenchahal:
This change is for honoring the InstanceUuid provided by caller in case of VM cloning.
Actual vSphere supports this scenario.

### DIFF
--- a/simulator/virtual_machine.go
+++ b/simulator/virtual_machine.go
@@ -1621,8 +1621,9 @@ func (vm *VirtualMachine) CloneVMTask(ctx *Context, req *types.CloneVM_Task) soa
 			return nil, &types.InvalidArgument{InvalidProperty: "spec.location.pool"}
 		}
 		config := types.VirtualMachineConfigSpec{
-			Name:    req.Name,
-			GuestId: vm.Config.GuestId,
+			Name:         req.Name,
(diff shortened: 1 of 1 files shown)


simulator/virtual_machine.go
```diff
@@ -1621,8 +1621,9 @@ func (vm *VirtualMachine) CloneVMTask(ctx *Context, req *types.CloneVM_Task) soa
 			return nil, &types.InvalidArgument{InvalidProperty: "spec.location.pool"}
 		}
 		config := types.VirtualMachineConfigSpec{
-			Name:    req.Name,
-			GuestId: vm.Config.GuestId,
+			Name:         req.Name,
+			GuestId:      vm.Config.GuestId,
+			InstanceUuid: req.Spec.Config.InstanceUuid,
```
dougm, line 1626: Thanks @parveenchahal 
Just need to move the setting of this field inside the `if req.Spec.Config != nil {` block below to avoid possible panic.
CI failed due to this: https://travis-ci.org/github/vmware/govmomi/jobs/706282464

parveenchahal: Thanks you, @dougm . I've updated the PR with the change.

PR #2057, opened 2020-07-12
Setting hardware properties in clone VM spec from template VM

parveenchahal:


dougm:
Thanks @parveenchahal

PR #2067, opened 2020-07-26
Add ExtendDisk and InflateDisk wrappers to vlsm/object_manager

### LINKED ISSUE #2066: vlsm ObjectManager is missing ExtendDisk and InflateDisk API
Currently, there is no way to Extend or Inflate disk using the vlsm ObjectManager. 

Let's use this issue to track adding simple wrappers for [HostExtendDisk_Task](https://github.com/vmware/govmomi/blob/master/vim25/methods/methods.go#L6554) and [HostInflateDisk_Task](https://github.com/vmware/govmomi/blob/master/vim25/methods/methods.go#L6634).

yashdesai93:
Fixes #2066

### DIFF
--- a/vslm/object_manager.go
+++ b/vslm/object_manager.go
@@ -258,6 +258,37 @@ func (m ObjectManager) RegisterDisk(ctx context.Context, path, name string) (*ty
 	return &res.Returnval, nil
 }
 
+func (m ObjectManager) ExtendDisk(ctx context.Context, ds mo.Reference, id string, newCapacityInMB int64) (*object.Task, error) {
+	req := types.HostExtendDisk_Task{
+		This:            m.Reference(),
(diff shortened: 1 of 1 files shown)


vslm/object_manager.go
```diff
@@ -258,6 +258,37 @@ func (m ObjectManager) RegisterDisk(ctx context.Context, path, name string) (*ty
 	return &res.Returnval, nil
 }
 
+func (m ObjectManager) ExtendDisk(ctx context.Context, id string, newCapacityInMB int64, ds mo.Reference) (*object.Task, error) {
```
dougm, line 261: The other methods here with similar params all have `ds` as the 2nd param and `id` as the third param. Can we do the same for these new methods?

PR #2068, opened 2020-07-26
Add AttachDisk and DetachDisk wrappers for the virtualMachine object.

### LINKED ISSUE #1846: Add API to detach and attach disks using FCD id for virtual machine object
Currently, there are methods to perform attach-disk[1] and detach-disk[2] tasks using the FCD id on a VM, but it would be useful to have API for the VirtualMachine object itself of the form:
```
func (v VirtualMachine) DetachDisk(ctx context.Context, id string) (error) 
func (v VirtualMachine) AttachDisk(ctx context.Context, id string, datastore ...) (error) 
```
For my usage this is how I implemented DetachDisk:
```
// DetachDisk detaches the disk from the vm.
func DetachDisk(ctx context.Context, vm *object.VirtualMachine, client *vim25.Client, id string) error {
	req := types.DetachDisk_Task{
		This:   vm.Reference(),
		DiskId: types.ID{Id: id},
	}

	res, err := methods.DetachDisk_Task(ctx, client, &req)
	if err != nil {
		return err
	}

	task := object.NewTask(client, res.Returnval)
	return task.Wait(ctx)
}
```

[1] https://github.com/vmware/govmomi/blob/master/vim25/methods/methods.go#L874
[2] https://github.com/vmware/govmomi/blob/master/vim25/methods/methods.go#L4334

yashdesai93:
Simple change to add wrappers around AttachDisk_Task and DetachDisk_Task,
to allow calling them from the VirtualMachine object.
Fixes #1846

### DIFF
--- a/object/virtual_machine.go
+++ b/object/virtual_machine.go
@@ -469,6 +469,41 @@ func (v VirtualMachine) RemoveDevice(ctx context.Context, keepFiles bool, device
 	return v.configureDevice(ctx, types.VirtualDeviceConfigSpecOperationRemove, fop, device...)
 }
 
+// AttachDisk attaches the given disk to the VirtualMachine
+func (v VirtualMachine) AttachDisk(ctx context.Context, id string, datastore *Datastore, controllerKey int32, unitNumber int32) error {
+	req := types.AttachDisk_Task{
+		This:          v.Reference(),
+		DiskId:        types.ID{Id: id},
+		Datastore:     datastore.Reference(),
+		ControllerKey: controllerKey,
+		UnitNumber:    &unitNumber,
+	}
+
+	res, err := methods.AttachDisk_Task(ctx, v.c, &req)
+	if err != nil {
+		return err
+	}
+
+	task := NewTask(v.c, res.Returnval)
+	return task.Wait(ctx)
+}
+
+// DetachDisk detaches the given disk from the VirtualMachine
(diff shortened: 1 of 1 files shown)


object/virtual_machine.go
```diff
@@ -469,6 +469,41 @@ func (v VirtualMachine) RemoveDevice(ctx context.Context, keepFiles bool, device
 	return v.configureDevice(ctx, types.VirtualDeviceConfigSpecOperationRemove, fop, device...)
 }
 
+// AttachDisk attaches the given disk to the VirtualMachine
+func (v VirtualMachine) AttachDisk(ctx context.Context, id string, datastore *Datastore, controllerKey int32, unitNumber int32) error {
```
dougm, line 473: I think the `unitNumber` param should also be `* int32`, otherwise it isn't possible to omit this optional field in the request. What do you think?

yashdesai93: I copied this from [vslm](https://github.com/vmware/govmomi/blob/f2eef8fc745fec4f654bbc85ea1c75bb88c24779/vslm/global_object_manager.go#L660) 
```
func (this *GlobalObjectManager) AttachDisk(ct context.Context, id vim.ID, vm mo.Reference, controllerKey int32,
	unitNumber int32) (*Task, error) {
```
I am ok either way though.

dougm: ok, let's keep it this way for consistency.

PR #2069, opened 2020-07-31
vcsim: avoid ViewManager.ModifyListView race

dougm:


dougm:
I'm not sure either. But `Registry.WithLock` is used to protect `This` by default when methods are invoked, such as the `ListView` instance when the `ModifyListView` method is invoked. The PropertyCollector also uses `WithLock` where appropriate.  Registry has a `[types.ManagedObjectReference]sync.Locker` and the race this test produced was the read of `mo.Reference.Reference()`. I suppose the race detector may not know what happens behind that interface impl, unlike it does when we use the `ListView.Self` field directly.  How we've avoided this problem with other objects, I don't know.
https://github.com/vmware/govmomi/blob/0614f98752bbc46dc2c46516d96898922a5898ff/simulator/registry.go#L524

PR #2076, opened 2020-08-04
Convert int to string using rune()

### LINKED ISSUE #2075: [BUG] Go 1.15: conversion from int to string yields a string of one rune
Go 1.15 rc 1 on Fedora Rawhide:

```
github.com/vmware/govmomi/vim25/xml
# github.com/vmware/govmomi/vim25/xml
./xml.go:1059:14: conversion from uint64 to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
./xml.go:1081:15: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
FAIL	github.com/vmware/govmomi/vim25/xml [build failed]
```

eclipseo:
See https://github.com/golang/go/issues/32479

Fix #2075.

Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>

### DIFF
--- a/vim25/xml/xml.go
+++ b/vim25/xml/xml.go
@@ -1056,7 +1056,7 @@ Input:
 					d.buf.WriteByte(';')
 					n, err := strconv.ParseUint(s, base, 64)
 					if err == nil && n <= unicode.MaxRune {
-						text = string(n)
+						text = string(rune(n))
 						haveText = true
 					}
 				}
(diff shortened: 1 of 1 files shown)


jtarchie:
Why was the XML packaged forked? This seems like a large effort.

dougm:
@jtarchie mostly to support type mapping when decoding interface fields. History is here: https://github.com/vmware/govmomi/commits/master/vim25/xml

It might be something that could be added to Go's stdlib, but I don't recall if anything was in the way of that other than time.

dougm:
Thanks again @eclipseo , #2100 includes this with the update to 1.15

Issue #1563, opened 2019-08-15
ability to add PCI device to a vm


warroyo:
it doesn't look like govc supports adding PCI devices to a vm today. Is this something that is already being considered? 

Happy to work on a PR for this if it is not already being looked at

dougm:
@warroyo happy to accept PRs that add to the govc device.* family of commands.  Not aware of any work underway in that area currently.

saedalavinia:
Hey all! 
We also have the need for this use case. 
@warroyo  have you worked on this sissue at all ?

ading007:
@warroyo we need to integrate it for cluster api, when will it be ready?

leslie-qiwa:
@dougm I'm exploring the effort for that. Can you please share the potential APIs to do the operation? Also includes listing available PCI devices?

dougm:
@leslie-qiwa great, have a look at the existing device commands, in particular the add commands:

```console
% ls -1 govc/device/*/add.go
govc/device/cdrom/add.go
govc/device/floppy/add.go
govc/device/scsi/add.go
govc/device/serial/add.go
govc/device/usb/add.go
```

Docs for the complete list of device types are here (follow the "Extended by" links):
https://vdc-download.vmware.com/vmwb-repository/dcr-public/b50dcbbf-051d-4204-a3e7-e1b618c1e384/538cf2ec-b34f-4bae-a332-3820ef9e7773/vim.vm.device.VirtualDevice.html

The `device.ls` and `device.info` commands should be able to list PCI devices.

leslie-qiwa:
wonderful! 

I'm going to work on some side subcommands firstly:

- govc host.hardware.ls: list all hardwares. --passthough option will only list passthrough capable devices
- govc host.hardware.passthrough enable <hardware ID>: enable passthrough on one device
- govc host.hardware.passthrough disable <hardware ID>: disable passthrough on one device

After that, I'll add support to add / delete one device for one VM.

dougm:
Added in #2092

PR #2086, opened 2020-08-18
add host.hardware.pci.ls subcommand to list all pci devices

leslie-qiwa:
this allows user to easily know PCI devices and passthrough capability

Signed-off-by: leslie-qiwa <leslie.qiwa@gmail.com>

Sample Output:

```
$ ./govc host.hardware.pci.ls  -k
Name:           localhost.localdomain
  Address       Description                      Parent        Passthrough
  0000:00:00.0  Intel(R)                                       Not Capable
  0000:00:04.0  Intel Corporation                              Not Capable
  0000:00:04.1  Intel Corporation                              Not Capable
  0000:00:04.2  Intel Corporation                              Not Capable
  0000:00:04.3  Intel Corporation                              Not Capable
  0000:00:04.4  Intel Corporation                              Not Capable
  0000:00:04.5  Intel Corporation                              Not Capable
  0000:00:04.6  Intel Corporation                              Not Capable
  0000:00:04.7  Intel Corporation                              Not Capable
  0000:00:05.0  Intel Corporation                              Not Capable
  0000:00:05.2  Intel(R)                                       Not Capable
  0000:00:05.4  Intel(R)                                       Not Capable
  0000:00:08.0  Intel Corporation                              Not Capable
  0000:00:08.1  Intel Corporation                              Not Capable
  0000:00:08.2  Intel Corporation                              Not Capable
  0000:00:11.0  Intel(R)                                       Not Capable
  0000:00:11.1  Intel(R)                                       Not Capable
  0000:00:11.5  Intel Corporation                              Not Capable
  0000:00:14.0  Intel Corporation                              Not Capable
  0000:00:14.2  Intel(R)                                       Not Capable
  0000:00:16.0  Intel Corporation                              Not Capable
  0000:00:16.1  Intel Corporation                              Not Capable
  0000:00:16.4  Intel Corporation                              Not Capable
  0000:00:1c.0  Intel Corporation                              Not Capable
  0000:00:1c.4  Intel Corporation                              Not Capable
  0000:00:1f.0  Intel Corporation                              Not Capable
  0000:00:1f.2  Intel Corporation                              Not Capable
  0000:00:1f.5  Intel Corporation                              Not Capable
  0000:02:00.0  Emulex Corporation               0000:00:1c.4  Not Capable
  0000:03:00.0  Matrox Electronics Systems Ltd.  0000:02:00.0  Disabled
  0000:17:05.0  Intel(R)                                       Not Capable
```

### DIFF
--- a/govc/host/hardware/pci/ls.go
+++ b/govc/host/hardware/pci/ls.go
@@ -1,0 +1,142 @@
+/*
+Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pci
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"text/tabwriter"
+
+	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+)
+
+type ls struct {
+	*flags.ClientFlag
+	*flags.HostSystemFlag
+	*flags.OutputFlag
+}
+
+func init() {
+	cli.Register("host.hardware.pci.ls", &ls{})
+}
+
+func (cmd *ls) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
+	cmd.ClientFlag.Register(ctx, f)
+
+	cmd.HostSystemFlag, ctx = flags.NewHostSystemFlag(ctx)
+	cmd.HostSystemFlag.Register(ctx, f)
+
+	cmd.OutputFlag, ctx = flags.NewOutputFlag(ctx)
+	cmd.OutputFlag.Register(ctx, f)
+}
+
+func (cmd *ls) Process(ctx context.Context) error {
+	if err := cmd.HostSystemFlag.Process(ctx); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (cmd *ls) Description() string {
+	return `
+Examples:
+  govc host.hardware.pci.ls`
+}
+
+func (cmd *ls) Run(ctx context.Context, f *flag.FlagSet) error {
+	c, err := cmd.Client()
+	if err != nil {
+		return err
+	}
+
+	// We could do without the -host flag, leaving it for compat
+	host, err := cmd.HostSystemIfSpecified()
+	if err != nil {
+		return err
+	} else if host == nil {
+		host, err = cmd.HostSystem()
+		if err != nil {
+			return err
+		}
+	}
+
+	res := []mo.HostSystem{}
+	refs := []types.ManagedObjectReference{host.Reference()}
+	pc := property.DefaultCollector(c)
+	props := []string{"summary", "hardware.pciDevice", "config.pciPassthruInfo"}
+	err = pc.Retrieve(ctx, refs, props, &res)
+	if err != nil {
+		return err
+	}
+
+	printInfo := passthruInfo{}
+	for _, obj := range res {
+		infos := map[string]*types.HostPciPassthruInfo{}
+		for _, o := range obj.Config.PciPassthruInfo {
+			info := o.GetHostPciPassthruInfo()
+			infos[info.Id] = info
+		}
+		for _, o := range obj.Hardware.PciDevice {
+			status := "Not Capable"
+			info, ok := infos[o.Id]
+			if ok {
+				if info.PassthruActive {
+					status = "Active"
+				} else if info.PassthruEnabled {
+					status = "Enabled"
+				} else if info.PassthruCapable {
+					status = "Disabled"
+				}
+			}
+			printInfo.Passthru = append(printInfo.Passthru, passthru{
+				Id: o.Id, VendorName: o.VendorName, ParentBridge: o.ParentBridge, Status: status,
+			})
+		}
+	}
+
+	return cmd.WriteResult(&printInfo)
+}
+
+type passthru struct {
+	Id           string
+	VendorName   string
+	ParentBridge string
+	Status       string
+}
+
+type passthruInfo struct {
+	Passthru []passthru
+}
+
+func (info *passthruInfo) Write(w io.Writer) error {
+	tw := tabwriter.NewWriter(os.Stdout, 2, 0, 2, ' ', 0)
+	fmt.Fprintf(tw, "Address\tDescription\tParent\tPassthrough\n")
+	for _, o := range info.Passthru {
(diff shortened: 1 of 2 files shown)


leslie-qiwa:
part of https://github.com/vmware/govmomi/issues/1563

leslie-qiwa:
@dougm Can you please share insight on this PR? If it is good, I can work on other parts.

leslie-qiwa:
Thanks for the reply. To achieve #1563, people need know the address of PCI device. In addition, people need enable this PCI device and activate it. Of course, people can do it through ESX UI, but I thought it would be nice to do it through govc. It is the rational behind.
I more prefer to a single host, and can remove multiple hosts option because seems like like govc commands support multiple hosts. So I thought VMWare wants to have more hosts support. I can change this part.
What you described is correct, it can be achieved with `govc host.esxcli hardware pci list`, but seems like esxcli doesn't have `toggle PCI passthrough` command. Or maybe it has already, please let me know.
With assumption of lack of `toggle PCI passthrough` esxcli commands, I'm thinking to develop list/toggle commands natively in govc directly. Let me know your thought.

dougm:
When I was asking can you describe how the command would be used, I meant more like an example:
```console
% id=$(govc host.hardware.pci.ls ...)
% govc device.something $id
```
I'm just trying to understand the input/output required, to see how the command(s) could be used to automate this process.

The doc link above mentions `hardware pci pcipassthru set`, have you looked at that for toggling?

As for single vs multiple host support, I think multiple has made some of the existing govc code more complex than needed. If single host will satisfy the use case, let's start with that.

leslie-qiwa:
> When I was asking can you describe how the command would be used, I meant more like an example:
> 
> ```
> % id=$(govc host.hardware.pci.ls ...)
> % govc device.something $id
> ```
From automation point of view, what I hope is it would be something like below:
```
% id=$(govc host.hardware.pci.ls ... | grep <vendor name> | grep Disabled | awk -F ' ' '{print $1}')
% govc device.something $id
% govc enter maintenance
% ipmitool reboot
% govc exit maintenance
% govc vm.pci.add vm1 $id[0]
% govc vm.pci.add vm2 $id[1]
```
> 
> I'm just trying to understand the input/output required, to see how the command(s) could be used to automate this process.
> 
> The doc link above mentions `hardware pci pcipassthru set`, have you looked at that for toggling?

Yes. It is provided in ESX 7.0, but unfortunately we are still using 6.7 and some are 6.5, which only provides list command. Not set command.
```
[root@localhost:~] esxcli hardware pci --help
Usage: esxcli hardware pci {cmd} [cmd options]

Available Commands:
  list                  List all of the PCI devices on this host.
```
> 
> As for single vs multiple host support, I think multiple has made some of the existing govc code more complex than needed. If single host will satisfy the use case, let's start with that.

will do.

leslie-qiwa:
also `esxcli hardware  pci list` only list the device's passthru capability, but not display current enable/disable state. Below is one example output. This is why we need the command to present the state.
```
0000:d8:00.0
   Address: 0000:d8:00.0
   Segment: 0x0000
   Bus: 0xd8
   Slot: 0x00
   Function: 0x0
   VMkernel Name: vmhba2
   Vendor Name: Intel(R)
   Device Name: <class> Non-Volatile memory controller
   Configured Owner: Unknown
   Current Owner: VMkernel
   Vendor ID: 0x8086
   Device ID: 0x0a54
   SubVendor ID: 0x8086
   SubDevice ID: 0x4714
   Device Class: 0x0108
   Device Class Name: Non-Volatile memory controller
   Programming Interface: 0x02
   Revision ID: 0x00
   Interrupt Line: 0x0b
   IRQ: 255
   Interrupt Vector: 0x00
   PCI Pin: 0x00
   Spawned Bus: 0x00
   Flags: 0x3201
   Module ID: 26
   Module Name: nvme
   Chassis: 0
   Physical Slot: 5
   Slot Description: SlotID:2
   Passthru Capable: true
   Parent Device: PCI 0:215:0:0
   Dependent Device: PCI 0:216:0:0
   Reset Method: Function reset
   FPT Sharable: true
```

leslie-qiwa:
Thanks for the hint. Pushed another change and allow output to be json or xml with the desired fields

govc/host/hardware/pci/ls.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/vmware/govmomi/vim25/types"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+)
+
+type ls struct {
+	*flags.ClientFlag
+	*flags.HostSystemFlag
+	*flags.OutputFlag
```
dougm, line 38: You can remove `ClientFlag` here since `HostSystemFlag` already includes it. Likewise for `OutputFlag`.

govc/host/hardware/pci/ls.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func init() {
+	cli.Register("host.hardware.pci.ls", &ls{})
+}
+
+func (cmd *ls) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
```
dougm, line 46: Same here, `ClientFlag` and `OutputFlag` can be removed, only need `HostSystem` flag.

govc/host/hardware/pci/ls.go
```diff
@@ -0,0 +1,142 @@
...
+	cmd.ClientFlag.Register(ctx, f)
+
+	cmd.HostSystemFlag, ctx = flags.NewHostSystemFlag(ctx)
+	cmd.HostSystemFlag.Register(ctx, f)
+
+	cmd.OutputFlag, ctx = flags.NewOutputFlag(ctx)
+	cmd.OutputFlag.Register(ctx, f)
+}
+
+func (cmd *ls) Process(ctx context.Context) error {
```
dougm, line 56: You can actually omit this method after removing the fields above, since the embedded `HostSystemFlag` will satisfy the interface.

govc/host/hardware/pci/ls.go
```diff
@@ -0,0 +1,142 @@
...
+	if err := cmd.HostSystemFlag.Process(ctx); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (cmd *ls) Description() string {
+	return `
```
dougm, line 64: Can you add a command description here as the first line as we do for other commands?

govc/host/hardware/pci/ls.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (cmd *ls) Run(ctx context.Context, f *flag.FlagSet) error {
+	c, err := cmd.Client()
+	if err != nil {
+		return err
+	}
+
+	// We could do without the -host flag, leaving it for compat
+	host, err := cmd.HostSystemIfSpecified()
```
dougm, line 76: Just use `cmd.HostSystem()` now that the command requires the `-host` flag:
```go
host, err := cmd.HostSystem()
if err != nil {
    return err
}
```

govc/host/hardware/pci/ls.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	res := []mo.HostSystem{}
+	refs := []types.ManagedObjectReference{host.Reference()}
+	pc := property.DefaultCollector(c)
+	props := []string{"summary", "hardware.pciDevice", "config.pciPassthruInfo"}
+	err = pc.Retrieve(ctx, refs, props, &res)
```
dougm, line 90: Using `RetrieveOne` would simplify a bit here, and for loop would no longer be needed below.

govc/host/hardware/pci/ls.go
```diff
@@ -0,0 +1,142 @@
+/*
+Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
```
dougm, line 2: This can just be year `2020`

dougm:
@leslie-qiwa closing with #2092 merged, feel free to re-open if this PR is still needed, thanks.

PR #2095, opened 2020-08-27
govc: add vm.console -wss flag

### LINKED ISSUE #2083: Acquire webmks ticket
It would be great to add this as an option to the govc `vm.console` command.

dougm:
Fixes #2083

### DIFF
--- a/govc/vm/console.go
+++ b/govc/vm/console.go
@@ -36,6 +36,7 @@ type console struct {
 	*flags.VirtualMachineFlag
 
 	h5      bool
+	wss     bool
 	capture string
 }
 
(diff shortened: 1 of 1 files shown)


dougm:
@lukasluecke can you review?

lukasluecke:
Looks good at first glance, I'll try it out later 🙂 Thanks!

govc/vm/console.go
```diff
@@ -163,6 +165,13 @@ func (cmd *console) Run(ctx context.Context, f *flag.FlagSet) error {
 		}.Encode()
 
 		link = u.String()
+	} else if cmd.wss {
+		ticket, err := vm.AcquireTicket(ctx, string(types.VirtualMachineTicketTypeWebmks))
```
lukasluecke, line 169: Should this switch maybe be moved to where the "normal" (clone) ticket would otherwise be acquired? So we don't always acquire both if we only need the Webmks.

dougm: Yes, good catch, moved the wss block to avoid AcquireCloneTicket

Issue #1396, opened 2019-03-05
Storage Policy Based Management (SPBM) support in govc CLI


cormachogan:
This is a new feature request. 

I would like to see support added to govc for storage policy based management, SPBM for short. (It appears as VM Storage Policies in the vSphere Client). This vSphere feature is used predominantly by vSAN but also Virtual Volumes (VVols) and VAIO (IO Filters). It also has a tag concept that can be used by any vSphere storage, including VMFS and NFS. The idea of SBPM is to add storage capabilities to a policy which can then be attached to a VM's storage. For example, if a VM was deployed on a vSAN, you would be able to assign a policy that specified availability characteristics.

I would like govc CLI to be able to do the following:

1. Query the list of available policies on a vCenter
2. Display the attributes/rule-set of a particular policy
3. Query which policy is assigned to a VM / VMDK
4. Assign a new policy to a VM / VMDK or range of VMs / VMDK
5. Change the attributes of an existing policy
6. Create a new policy and set attributes accordingly
7. Display which VMs/VMDKs are using a policy
8. Delete a policy

Attached are some screenshot from the vSphere Client which displays some of the requests highlighted above. These are from a vSAN environment, which is the most popular consumer of SPBM.

![1 2](https://user-images.githubusercontent.com/18258500/53803625-26526600-3f3d-11e9-9fe2-ee1170b25203.png)

![3 4](https://user-images.githubusercontent.com/18258500/53803707-6fa2b580-3f3d-11e9-9b3c-115b5e6e0b24.png)

cormachogan:
When changing the attributed of a policy that is already in use by a VM / VMDK, the user should state whether they want to change the policy immediately, or change it later. There are operational reasons for this.
![6a](https://user-images.githubusercontent.com/18258500/53804063-69f99f80-3f3e-11e9-83ea-06d7d28f5458.png)
![6b](https://user-images.githubusercontent.com/18258500/53804075-6ebe5380-3f3e-11e9-8129-343234cd25a6.png)

cormachogan:
Some details on how some of this functionality is implemented in PowerCLI cmdlets.

**Get-SpbmCompatibleStorage** - This cmdlet performs placement compatibility checking based on a storage requirement policy - https://www.vmware.com/support/developer/PowerCLI/PowerCLI651/html/Get-SpbmCompatibleStorage.html

[...]

PR #2096, opened 2020-08-28
govc: add storage.policy commands

### LINKED ISSUE #2085: How can I find an UUID of the storage policy?
I need to find an UUID of storage policy with govc.

dougm:
Issue #1396
Fixes #2085


dacodas:
Thanks for the hard work :)

PR #2100, opened 2020-08-29
vim25/xml: sync with Go 1.15 encoding/xml

dougm:
Applied via:

% cd ~/go/src/github.com/golang/go
% git diff go1.13.6..go1.15 src/encoding/xml > encoding-xml.patch
% cd ~/go/src/github.com/vmware/govmomi/vim25/xml
% patch -p4 < ~/go/src/github.com/golang/go/encoding-xml.patch

Diff after sync: https://gist.github.com/dougm/a941f77324a6172712430c3ff9ff8c63


dougm:
Fixes #2075
Closes #2076

cc @eclipseo @jtarchie

PR #2103, opened 2020-09-03
govc: add CNS volume ls and rm commands

### LINKED ISSUE #2051: Add govc cns.* commands to manage CNS volumes
While working with CAPV I ran into this issue: https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/251. If you delete your cluster before deleting your volumes you get a ton of orphaned CNS volumes.

There is no way to delete these volumes from the vCenter console. @chethanv28 suggested using VSAN Managed Object Browser (MOB) to delete the volumes: https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/251#issuecomment-654395041. But didn't want to enable a debugging tool on my cluster, and I would have had to delete each volume manually.

Please consider adding all the [CNS client methods](https://github.com/vmware/govmomi/blob/master/cns/client.go) as `govc cns.*` command line tools. It would be nice to script together simple things like `govc cns.ls | govc cns.rm` to delete batches of volumes at once. It might be nice to query by metadata from the command line too.

I ended up writing a simple go program using govmomi to do delete all CNS volumes:
https://gist.github.com/bburky/7047be4866881aa22030ba4a74b89538

dougm:
Fixes #2051

### DIFF
--- a/cns/simulator/simulator.go
+++ b/cns/simulator/simulator.go
@@ -32,6 +32,14 @@ import (
 	vim25types "github.com/vmware/govmomi/vim25/types"
 )
 
+func init() {
+	simulator.RegisterEndpoint(func(s *simulator.Service, r *simulator.Registry) {
+		if r.IsVPX() {
+			s.RegisterSDK(New())
+		}
+	})
+}
+
 func New() *simulator.Registry {
 	r := simulator.NewRegistry()
 	r.Namespace = cns.Namespace
--- a/govc/main.go
+++ b/govc/main.go
@@ -96,6 +96,7 @@ import (
 	_ "github.com/vmware/govmomi/govc/vm/option"
 	_ "github.com/vmware/govmomi/govc/vm/rdm"
 	_ "github.com/vmware/govmomi/govc/vm/snapshot"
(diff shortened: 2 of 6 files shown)


dougm:
cc @bburky

bburky:
Thanks, adding `volume.ls` and `volume.rm` commands should cover the scenarios I was initially concerned with. Because the PVC name is used for the volume names it is possible to match up the volumes with a running cluster, and only delete the undesired ones.

It might be nice to expose the other metadata fields or the remaining CNS client methods, but I have no need for them at this time personally.

dougm:
Thanks @bburky . All the metadata can be viewed using `volume.ls -json` at least. We can look at including more details in another format if needed.  I had started to look at doing `volume.create`, but likewise don't have the need currently.

divyenpatel:
Thanks you @dougm for adding support for this in govc.

bburky:
@dougm Thanks again for implementing this. I just wrote a shell script that uses this to delete orphaned volumes. It seems to work well. https://gist.github.com/bburky/fd729f129347b9141255fbc2185c52ea

dougm:
Nice @bburky !

PR #2106, opened 2020-09-06
sts: fix SignRequest bodyhash for non-empty request body

dougm:
The only request that is signed by this method always has an empty body:
https://vmware.github.io/vsphere-automation-sdk-rest/vsphere/operations/com/vmware/cis/session.create-operation.html
The 'bodyhash' header param of an empty string is still required in the signature however.
But should the request body be non-empty, we need to use sha256.Sum256() as @dgryski pointed out.


dougm:
cc @dgryski

PR #2142, opened 2020-10-05
govc (+ vapi): namespace.cluster.enable for enabling namespaces on a cluster

benjvi:
Hi! 

This covers some functionality I wanted to use for programmatically enabling Kubernetes in Vsphere. I saw some TODO comments around this so hopefully this PR is welcome. 

This did turn out to be more involved than I initially expected, this covers:

- Structs corresponding to the enable spec from the API added to the VAPI package, along with EnableCluster and DisableCluster functions
- Corresponding cluster.namespace.enable and cluster.namespace.disable commands
- Also added ListCompatibleDistributedSwitches and ListCompatibleEdgeClusters to VAPI, which are needed to get the IDs for these objects in the format expected by the API
- Simple E2E test via govc/bats with the vcsim returning success on the relevant endpoint
- Ran the corresponding commands against my own Vsphere installation, and the supervisor cluster came up and was working

There is more that could be done (e.g. DefaultKubernetesServiceContentLibrary, validating some cases that the API doesn't), but hopefully this is enough for a first pass

### DIFF
--- a/govc/namespace/cluster/disable.go
+++ b/govc/namespace/cluster/disable.go
@@ -1,0 +1,84 @@
+/*
+Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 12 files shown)


benjvi:
It seems I need to reopen this PR to appease the CLA bot

PR #2143, opened 2020-10-05
govc (+ vapi): namespace.cluster.enable for enabling namespaces on a cluster

benjvi:
Hi!

This covers some functionality I wanted to use for programmatically enabling Kubernetes in Vsphere. I saw some TODO comments around this so hopefully this PR is welcome.

This did turn out to be more involved than I initially expected, this covers:
- Structs corresponding to the enable spec from the API added to the VAPI package, along with EnableCluster and DisableCluster functions
- Corresponding cluster.namespace.enable and cluster.namespace.disable commands
- Also added ListCompatibleDistributedSwitches and ListCompatibleEdgeClusters to VAPI, which are needed to get the IDs for these objects in the format expected by the API
- Simple E2E test via govc/bats with the vcsim returning success on the relevant endpoint
- Ran the corresponding commands against my own Vsphere installation, and the supervisor cluster came up and was working

There is more that could be done (e.g. DefaultKubernetesServiceContentLibrary, validating some cases that the API doesn't), but hopefully this is enough for a first pass

### DIFF
--- a/vapi/namespace/namespace.go
+++ b/vapi/namespace/namespace.go
@@ -82,7 +82,7 @@ type AddressRange struct {
 	SubnetMask      string `json:"subnet_mask,omitempty"`
 	StartingAddress string `json:"starting_address"`
 	Gateway         string `json:"gateway"`
-	AddressCount    int    `json:"address_count"`
+	AddressCount    int    `json:"address_count,omitempty"`
 }
 
 type MasterManagementNetwork struct {
@@ -101,10 +101,9 @@ type DefaultImageRegistry struct {
 func (c *Manager) EnableCluster(ctx context.Context, id string, spec *EnableClusterSpec) error {
 	var response interface{}
 	url := c.Resource(path.Join(internal.NamespaceClusterPath, id)).WithParam("action", "enable")
-	bytes, _ := json.MarshalIndent(spec, "", "\t")
-	fmt.Println(string(bytes))
+	// bytes, _ := json.MarshalIndent(spec, "", "\t")
+	// fmt.Println(string(bytes))
 	err := c.Do(ctx, url.Request(http.MethodPost, spec), response)
-	fmt.Printf("%#v", response)
 	return err
 }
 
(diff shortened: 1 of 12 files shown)


vapi/namespace/namespace.go
```diff
@@ -40,6 +41,82 @@ func NewManager(client *rest.Client) *Manager {
...
+type DefaultImageRegistry struct {
+	Hostname string `json:"hostname"`
+	Port     int    `json:"port,omitempty"`
+}
+
+// EnableCluster enables vSphere Namespaces on the specified cluster, using the given spec.
+func (c *Manager) EnableCluster(ctx context.Context, id string, spec *EnableClusterSpec) error {
+	var response interface{}
+	url := c.Resource(path.Join(internal.NamespaceClusterPath, id)).WithParam("action", "enable")
+	// bytes, _ := json.MarshalIndent(spec, "", "\t")
```
dougm, line 104: You can also use govc `-debug` to trace: https://github.com/vmware/govmomi/blob/master/govc/README.md#debug-flag
There's also a script to help read it: https://github.com/vmware/govmomi/blob/master/scripts/debug-xmlformat.sh

vapi/namespace/namespace.go
```diff
@@ -94,3 +171,35 @@ func (c *Manager) SupportBundleRequest(ctx context.Context, bundle *SupportBundl
...
+type DistributedSwitchCompatibilitySummary struct {
+	Compatible        bool   `json:"compatible"`
+	DistributedSwitch string `json:"distributed_switch"`
+}
+
+func (c *Manager) ListCompatibleDistributedSwitches(ctx context.Context, clusterId string) (result []DistributedSwitchCompatibilitySummary, err error) {
+	listUrl := c.Resource(internal.NamespaceDistributedSwitchCompatibility).
+		WithParam("cluster", clusterId).
+		WithParam("compatible", "true")
+	if err != nil {
```
dougm, line 184: Don't see how err could `!= nil` here?

vapi/namespace/namespace.go
```diff
@@ -94,3 +171,35 @@ func (c *Manager) SupportBundleRequest(ctx context.Context, bundle *SupportBundl
...
+	EdgeCluster string `json:"edge_cluster"`
+	DisplayName string `json:"display_name"`
+}
+
+func (c *Manager) ListCompatibleEdgeClusters(ctx context.Context, clusterId string, switchId string) (result []EdgeClusterCompatibilitySummary, err error) {
+	listUrl := c.Resource(internal.NamespaceEdgeClusterCompatibility).
+		WithParam("cluster", clusterId).
+		WithParam("distributed_switch", switchId).
+		WithParam("compatible", "true")
+	if err != nil {
```
dougm, line 201: Don't see how err could `!= nil` here?

dougm (approved):
Thank you @benjvi !  Looks good to me.

vapi/namespace/namespace.go
benjvi: I hadn't seen that, thanks! I'll clean these comments up

vapi/namespace/namespace.go
benjvi: This was an oversight, I did a couple of iterations on the changes to WithParam. Now it swallows the error returned by ParseQuery, so these can be removed

dougm:
Thanks again @benjvi !

PR #2144, opened 2020-10-06
simulator: fix container vm example

sigma:
This fixes an issue in setups where the code directory doesn't have enough
permissions for a container to access.
This would result in the test hanging until timeout, and dangling containers.

Since there is no need to serve the source code via http in the first place,
generate a dummy file instead, and use it in the test.


dougm (approved):
Thanks @sigma

PR #2147, opened 2020-10-07
UseParseInt(x, 10, 32) instead of int32(Atoi(x))

disconnect3d:
The Go's [`Atoi(x)`](https://golang.org/pkg/strconv/#Atoi) is equivalent of [`ParseInt(x, 10, 0)`](https://golang.org/pkg/strconv/#ParseInt) which may return 64-bit numbers.

This PR makes it so instead of overflowing this number, the `CustomFieldsManager.FindKey` function will return an `ErrKeyNameNotFound` if a too big number is passed in the `name` string.

PS: I haven't tested this change, please do so, but it should just work, assuming that the code did not depend on int overflow before.

### DIFF
--- a/object/custom_fields_manager.go
+++ b/object/custom_fields_manager.go
@@ -127,7 +127,7 @@ func (m CustomFieldsManager) FindKey(ctx context.Context, name string) (int32, e
 		}
 	}
 
-	k, err := strconv.Atoi(name)
+	k, err := strconv.ParseInt(name, 10, 32)
 	if err == nil {
 		// assume literal int key
 		return int32(k), nil


disconnect3d:
Hey,

I'd prefer not to give my mailing address or phone number in order to sign a CLA.
I waive my rights to the contribution made in this pull request - changing the `k, err := strconv.Atoi(name)` line to `k, err := strconv.ParseInt(name, 10, 32)`. I hope this is enough to merge this without signing a CLA...

dougm (approved):
Thanks @disconnect3d , understood.  We haven't been able to make CLA exceptions in the past, but seems like we should be able to in this case.  I will look into it.

PR #2150, opened 2020-10-09
Add option to disable secure cookies with non-TLS endpoints

dougm:
vCenter login methods include the Secure flag in Set-Cookie responses.
Go's cookiejar impl does not send secure cookies unless the URL scheme is https.
This prevents use of govmomi/govc against the non-TLS vpxd endpoint (127.0.0.1:8085)

To enable use of secure cookies in this case: export GOVMOMI_INSECURE_COOKIES=true

To make vcsim include the Secure flag in Set-Cookie responses: export VCSIM_SECURE_COOKIES=true

### DIFF
--- a/govc/README.md
+++ b/govc/README.md
@@ -182,6 +182,16 @@ Test connection using curl:
 Inventory path arguments with a leading '/' are subject
 to [Posix path conversion](http://www.mingw.org/wiki/Posix_path_conversion).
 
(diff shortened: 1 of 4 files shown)


akutz:
lgtm. small nit - should we document that this is useful when connecting directly to vpxd?

akutz:
Thanks! LGTM!

PR #2155, opened 2020-10-14
support trunk mode port group

datianshi:
Current govc implementation only creates vlan based port group
This pull requests

- Preserves the usage of the default vlan based port group creation/modification
- Add two flags -vlan-mode and -vlan-range, so users can create trunking port group with a list of ranges
- the default range is all -> 0 - 4094


dougm (approved):
Thanks @datianshi !

PR #2172, opened 2020-11-07
govc: add device.boot -firmware option

### LINKED ISSUE #2126: govc: device.boot misses an option to set EFI/BIOS boot
As a `govc` user, I'd like to have a way to set EFI or BIOS firmware setting via `device.boot` on an existing VM, so that I can switch between BIOS / plain-EFI / EFI with Secure Boot modes.

Right now trying to set Secure Boot mode on a machine configured with BIOS firmware fails with:
```
$ govc device.boot -vm "${MACHINE}" -secure
govc: Invalid virtual machine configuration.
```

After manually changing to EFI boot in the machine settings in the webUI, the command above succeeds. However, I'm missing a way to perform this BIOS<->EFI switch through `govc device.boot`.

https://github.com/vmware/govmomi/pull/1072 added this to `govc vm.create`, but as I'm deploying via `library.deploy` I cannot go through that. Also, IMHO it would be good anyway to have a way to flip the value after machine creation.

dougm:
Fixes #2126

### DIFF
--- a/govc/device/boot.go
+++ b/govc/device/boot.go
@@ -23,13 +23,15 @@ import (
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/govc/vm"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
 type boot struct {
 	*flags.VirtualMachineFlag
 
-	order string
+	firmware string
+	order    string
 	types.VirtualMachineBootOptions
 }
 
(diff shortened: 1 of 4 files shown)


lucab:
LGTM. I've manually validated that the commands below effectively switch the VM boot setting, as reflected into the webUI:
```
$ govc device.boot -vm "${MACHINE}" --firmware efi

$ govc device.boot -vm "${MACHINE}" --firmware bios  
```

lucab (approved):

dougm:
Great, thanks @lucab !

PR #2184, opened 2020-11-20
Add vSAN 7.0 U1 API bindings

gh05tn0va:
The bindings are auto generated from vSAN vMODL definations during esx build.


gh05tn0va:
note: commit 6454dbd is for 7.0 U1 APIs. The description was mistaken.

divyenpatel:
@gh05tn0va if these bindings are generated automatically we are good. 
but vSphere CSI driver is using CNS bindings and client from https://github.com/vmware/govmomi/tree/master/cns

Auto-generated binding does not have some internal methods and types, but we need them for the vSphere CSI driver.

gh05tn0va:
@divyenpatel The solution is to make these required cns APIs public.
Once this is done, these bindings can be regenerated using latest build.
For now, I think we can tolerate the incomplete cns bindings here.

gh05tn0va:
@divyenpatel what is the status of this review? Would you approve if no other requirements?

divyenpatel (approved):

dougm (approved):
Thanks @gh05tn0va

PR #2205, opened 2020-12-11
Add function to get vsan host config

gh05tn0va:
VsanHostConfigInfoEx is a property of HostVsanSystem in "vsan" namespace, which is an extension of vim25types.VsanHostConfigInfo.
Adding function to retrieve this property from given vsanSystem.


dougm (approved):
Thanks @gh05tn0va , lgtm

PR #2207, opened 2020-12-13
govc: add cluster.usage command

dougm:


dougm:
Example output:
```console
% govc cluster.usage -S /folder-WCP_DC/WCP_DC/host/compute-cluster
CPU usage:         28.3%
CPU capacity:      75.4GHz
CPU used:          21.4GHz
CPU free:          54.0GHz
                   
Memory usage:      67.0%
Memory capacity:   96.0GB
Memory used:       64.3GB
Memory free:       31.6GB
                   
Storage usage:     29.7%
Storage capacity:  858.3GB
Storage used:      254.8GB
Storage free:      603.4GB
```

govc/cluster/usage.go
```diff
@@ -0,0 +1,185 @@
...
+		return err
+	}
+
+	obj, err := finder.ClusterComputeResource(ctx, f.Arg(0))
+	if err != nil {
+		return err
+	}
+
+	var res Usage
+	props := []string{"summary", "datastore", "host"}
```
sidharthsurana, line 75: 1. Nit: Seems like we are not using the "summary" property at all, thus could be removed.
2. I see that you are explicitly iterating over the hosts/datastores to calculate the cpu/mem/storage usage. Question: Couldn't/Shoudn't we use the GetResourceUsage method on the ClusterComputeResource object for this data?  see https://code.vmware.com/apis/358/vsphere/doc/vim.ClusterComputeResource.html#getResourceUsage (Note: Earlier I tried to correlate the data returned from this method to what we see in the UI and those did not align completely as I expected. So it is possible that what you are doing is more accurate. But I wanted to refer to this object here so that we can document at least why we did not use this for future code readers reference)

dougm: nit fixed.  I hadn't actually seen the `GetResourceUsage` method, I can give it a try.
The cluster "summary" property was there as I was trying to use https://vdc-download.vmware.com/vmwb-repository/dcr-public/b50dcbbf-051d-4204-a3e7-e1b618c1e384/538cf2ec-b34f-4bae-a332-3820ef9e7773/vim.ComputeResource.Summary.html , but that data didn't align with the UI either.

So I opted for using the host data, which is just an aggregate of what you'd see with the govc `host.info` command.

dougm: @sidharthsurana I'd like to get govc 0.24 released before the break, but probably won't have time to look at switching to `GetResourceUsage`. But could look at using that after the break to optimize the cluster.usage command.  What do you think?

sidharthsurana: sure, that would work.

PR #2226, opened 2021-01-03
vcsim: fix EventHistoryCollector fixes

### LINKED ISSUE #2134: Incorrect behavior of event history collector in vcsim simulator
Compared to a regular vCenter installation (tested with vCenter Server v6.7) the event history collector and some of its APIs, e.g. `ReadNextEvent` in particular, show some different and likely undesired behavior:

- The embedded `EventHistoryCollector` in `vcsim` and its ring buffer purely relies on (and serves from) `latestPage`
- To comply with `latestPage` API behavior, events are not sorted (thus `govc` leverages a sort function to print the events in order) - the internal ring buffer reads backwards via `page.Prev()`
- `ReadNextPage` in `vcsim` also relies on `latestPage` which is not the case in a real vCenter
- `PostEvent` uses `Map.Update(c, []types.PropertyChange{{Name: "latestPage", Val: c.GetLatestPage()}})` which means custom `EventHistoryCollector` created without this option/with specific `EventFilter`s won't be triggered

For example, given the following code:

```go
func newHistoryCollector(ctx context.Context, vcClient *vim25.Client, begin *time.Time) (*event.HistoryCollector, error) {
	mgr := event.NewManager(vcClient)
	root := vcClient.ServiceContent.RootFolder
	// configure the event stream filter (begin of stream)
	filter := types.EventFilterSpec{
		// EventTypeId: []string{...}, // only stream specific types, e.g. VmEvent
		Entity: &types.EventFilterSpecByEntity{
			Entity:    root,
			Recursion: types.EventFilterSpecRecursionOptionAll,
		},
		Time: &types.EventFilterSpecByTime{
			BeginTime: types.NewTime(*begin),
		},
	}
	return mgr.CreateCollectorForEvents(ctx, filter)
}
```

`vcsim` will only serve the first call to `ReadNextEvents` but no later events, such as `VmPoweredOffEvent` triggered via `govc` against the simulator.

Since more users are relying on vCenter events/programmatically using its APIs, it would be nice to see `vcsim` events API behave like a real vCenter, especially for integration testing. `latestPage` might not be the standard use case for retrieving events from vCenter, especially when a user wants control over the polling interval or start of the event stream (via filter/`ReadNext`). `latestPage` is also used in the event manager in govmomi and hardcoded via `props := []string{"latestPage"}` in `run()`.

dougm:
Fix Next/Previous methods:

The original implementation only applied to the LatestPage ring buffer.
Switch EventManager's history to use a linked list that goes beyond the
LatestPage size.

Fix SetCollectorPageSize: had caused duplicate events in LatestPage

Fixes #2134

### DIFF
--- a/simulator/event_manager.go
+++ b/simulator/event_manager.go
@@ -18,6 +18,7 @@ package simulator
 
 import (
 	"bytes"
+	"container/list"
 	"container/ring"
 	"log"
 	"reflect"
@@ -40,7 +41,7 @@ type EventManager struct {
 	mo.EventManager
 
 	root       types.ManagedObjectReference
-	page       *ring.Ring
+	history    *list.List
 	key        int32
 	collectors map[types.ManagedObjectReference]*EventHistoryCollector
 	templates  map[string]*template.Template
@@ -54,7 +55,7 @@ func (m *EventManager) init(r *Registry) {
 		m.MaxCollector = 1000
 	}
 	m.root = r.content().RootFolder
-	m.page = ring.New(maxPageSize)
+	m.history = list.New()
 	m.collectors = make(map[types.ManagedObjectReference]*EventHistoryCollector)
 	m.templates = make(map[string]*template.Template)
 }
@@ -155,9 +156,14 @@ func (m *EventManager) PostEvent(ctx *Context, req *types.PostEvent) soap.HasFau
 	event.CreatedTime = time.Now()
 	event.UserName = ctx.Session.UserName
 
-	m.page = m.page.Prev()
-	m.page.Value = req.EventToPost
 	m.formatMessage(req.EventToPost)
+
+	if m.history.Len() > maxPageSize*5 {
+		// Prune history
+		m.history.Remove(m.history.Front())
+	}
+
+	m.history.PushBack(req.EventToPost)
 
 	for _, c := range m.collectors {
 		ctx.WithLock(c, func() {
@@ -179,7 +185,7 @@ type EventHistoryCollector struct {
 
 	m    *EventManager
 	page *ring.Ring
-	pos  int
+	pos  *list.Element
 }
 
 // doEntityEventArgument calls f for each entity argument in the event.
@@ -327,31 +333,11 @@ func (c *EventHistoryCollector) eventMatches(event types.BaseEvent) bool {
 
 // filePage copies the manager's latest events into the collector's page with Filter applied.
 func (c *EventHistoryCollector) fillPage(size int) {
-	c.pos = 0
-	l := c.page.Len()
-	delta := size - l
-
-	if delta < 0 {
-		// Shrink ring size
-		c.page = c.page.Unlink(-delta)
-		return
-	}
-
 	matches := 0
-	mpage := c.m.page
 	page := c.page
 
-	if delta != 0 {
-		// Grow ring size
-		c.page = c.page.Link(ring.New(delta))
-	}
-
-	for i := 0; i < maxPageSize; i++ {
-		event, ok := mpage.Value.(types.BaseEvent)
-		mpage = mpage.Prev()
-		if !ok {
-			continue
-		}
+	for e := c.m.history.Back(); e != nil; e = e.Prev() {
+		event := e.Value.(types.BaseEvent)
 
 		if c.eventMatches(event) {
 			page.Value = event
@@ -385,6 +371,7 @@ func (c *EventHistoryCollector) SetCollectorPageSize(ctx *Context, req *types.Se
 	}
 
 	ctx.WithLock(c.m, func() {
+		c.page = ring.New(size)
 		c.fillPage(size)
 	})
 
@@ -393,7 +380,7 @@ func (c *EventHistoryCollector) SetCollectorPageSize(ctx *Context, req *types.Se
 }
 
 func (c *EventHistoryCollector) ResetCollector(ctx *Context, req *types.ResetCollector) soap.HasFault {
-	c.pos = len(c.GetLatestPage())
+	c.pos = c.m.history.Back()
 
 	return &methods.ResetCollectorBody{
 		Res: new(types.ResetCollectorResponse),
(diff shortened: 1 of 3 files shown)


simulator/event_manager.go
```diff
@@ -327,31 +333,11 @@ func (c *EventHistoryCollector) eventMatches(event types.BaseEvent) bool {
 
 // filePage copies the manager's latest events into the collector's page with Filter applied.
 func (c *EventHistoryCollector) fillPage(size int) {
```
embano1, line 335: This seems to be thread-safe since the calling `SetCollectorPageSize()` holds a lock on the manager.

simulator/event_manager.go
```diff
@@ -179,7 +185,7 @@ type EventHistoryCollector struct {
 
 	m    *EventManager
 	page *ring.Ring
-	pos  int
+	pos  *list.Element
```
embano1, line 188: Checked all reads/writes on this field and they seem to be protected by holding a lock on the manager.

simulator/event_manager.go
```diff
@@ -179,7 +185,7 @@ type EventHistoryCollector struct {
 
 	m    *EventManager
 	page *ring.Ring
```
embano1, line 187: Checked all reads/write to `page` and it's mostly protected but not in `GetLatestPage()`. Not sure if there could be a race between `GetLatestPage()` and e.g. `PostEvents()` which does acquire a lock to protect the `page`:

```go
for _, c := range m.collectors {
		ctx.WithLock(c, func() {
			if c.eventMatches(req.EventToPost) {
				c.page = c.page.Prev()
				c.page.Value = req.EventToPost
				Map.Update(c, []types.PropertyChange{{Name: "latestPage", Val: c.GetLatestPage()}})
			}
		})
	}
```

embano1 (commented):
Verified the PR against a dummy event producer/consumer using `event.Manager` and `EventHistoryCollector` and the PR seems to have fixed the issues:

- ordered (ascending by key) delivery of events
- delivery of events to manager and history collector

```bash
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 47 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 48 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 49 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 50 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 51 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 52 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 53 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 54 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 55 and type: VmPoweredOnEvent
2021/01/05 09:41:45 [STREAM] received event order ID 0 with key 56 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 0 with key 31 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 1 with key 32 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 2 with key 33 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 3 with key 34 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 4 with key 35 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 5 with key 36 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 6 with key 37 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 7 with key 38 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 8 with key 39 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [COLLECTOR] received event order ID 9 with key 40 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 57 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 58 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 59 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 60 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 61 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 62 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 63 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 64 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 65 and type: VmPoweredOnEvent
2021/01/05 09:41:46 [STREAM] received event order ID 0 with key 66 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 0 with key 41 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 1 with key 42 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 2 with key 43 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 3 with key 44 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 4 with key 45 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 5 with key 46 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 6 with key 47 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 7 with key 48 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 8 with key 49 and type: VmPoweredOnEvent
2021/01/05 09:41:47 [COLLECTOR] received event order ID 9 with key 50 and type: VmPoweredOnEvent
```

> Note: the collector retrieves the full history (snippet) while the manager always serves from latest page using an event consumer producing 10 events/s

Also verified against a VMware Event Broker installation and events are delivered correctly (ascending key order) and functions are correctly triggered, e.g. `VmPoweredOnEvent`.

simulator/event_manager.go
dougm: `c.GetLatestPage()` reads from `c.page`, but is also within the `ctx.WithLock(c,...)`.

And API access to `GetLatestPage()` is via the PropertyCollector, which locks the object here: https://github.com/vmware/govmomi/blob/fe8ef53a7a43c875e08bfb1c119b698073fe0376/simulator/property_collector.go#L441

embano1: Thx, concurrent access to GetLatestPage via API was my concern, so looks like we're good 👍

embano1 (approved):
Thx for this fix Doug, helps me to implement the Knative work now!

dougm:
Thanks for the thorough review and testing @embano1 !

PR #2233, opened 2021-01-08
vcsim: use linked list for EventHistoryCollector

dougm:
PR #2134 changed EventManager's history ring buffer to a linked list.
Doing the same for EventHistoryCollector as it simplifies the code and
avoid having to lock the EventManager in the Read{Next,Prev}Events methods.

### DIFF
--- a/simulator/event_manager.go
+++ b/simulator/event_manager.go
@@ -19,7 +19,6 @@ package simulator
 import (
 	"bytes"
 	"container/list"
-	"container/ring"
 	"log"
 	"reflect"
 	"text/template"
@@ -72,10 +71,11 @@ func (m *EventManager) createCollector(ctx *Context, req *types.CreateCollectorF
 
 	collector := &EventHistoryCollector{
 		m:    m,
-		page: ring.New(size),
+		page: list.New(),
+		size: size,
 	}
 	collector.Filter = req.Filter
-	collector.fillPage(size)
+	collector.fillPage()
 
 	return collector, nil
 }
@@ -148,6 +148,13 @@ func (m *EventManager) formatMessage(event types.BaseEvent) {
 	}
 }
 
+func pushEvent(l *list.List, event types.BaseEvent) {
+	if l.Len() > maxPageSize*5 {
+		l.Remove(l.Front()) // Prune history
+	}
+	l.PushBack(event)
+}
+
 func (m *EventManager) PostEvent(ctx *Context, req *types.PostEvent) soap.HasFault {
 	m.key++
 	event := req.EventToPost.GetEvent()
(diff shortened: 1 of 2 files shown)


simulator/event_manager.go
```diff
@@ -370,25 +364,24 @@ func (c *EventHistoryCollector) SetCollectorPageSize(ctx *Context, req *types.Se
...
+	c.page = list.New()
+	ctx.WithLock(c.m, c.fillPage)
 
 	body.Res = new(types.SetCollectorPageSizeResponse)
 	return body
 }
 
 func (c *EventHistoryCollector) ResetCollector(ctx *Context, req *types.ResetCollector) soap.HasFault {
-	c.pos = c.m.history.Back()
+	c.pos = c.page.Back()
```
embano1, line 376: Having a hard time understanding the intended behavior of this method reading the docs:

```
Moves the "scrollable view" to the item immediately preceding the "viewable latest page". 
If you use "readPrev", ReadPreviousTasks or ReadPreviousEvents, all items are retrieved from the 
newest item to the oldest item.
```

Is this line correct based on the above method description?

simulator/event_manager.go
```diff
@@ -370,25 +364,24 @@ func (c *EventHistoryCollector) SetCollectorPageSize(ctx *Context, req *types.Se
...
 
 	return &methods.ResetCollectorBody{
 		Res: new(types.ResetCollectorResponse),
 	}
 }
 
 func (c *EventHistoryCollector) RewindCollector(ctx *Context, req *types.RewindCollector) soap.HasFault {
```
embano1, line 383: Assuming these methods are called via lock (as discussed earlier), so they're thread-safe right?

embano1 (commented):
Just two questions for clarification.

Verified the behavior with VEBA (VMware Event Router) and a binary using EventManager against this updated vcsim and events are retrieved as expected.

Just noticed that we don't use any filterspec when creating a collector, filed an issue for this: https://github.com/vmware/govmomi/issues/2240

simulator/event_manager.go
dougm: Yes. When you call `ResetCollector()` pos is set to the back of the list, and new events are added to the back of the list.  After the Reset call, Read**Next**Events() will return `nil` until new events come in.
But you'll be able to read the entire history with `ReadPreviousEvents()` after calling Reset.

`RewindCollector()` is the opposite: Read**Prev**Events() will return `nil`, but can read the entire history with`ReadNextEvents()`

simulator/event_manager.go
dougm: Yes 👍

simulator/event_manager.go
embano1: Thx, I found the name `Reset` a bit confusing :)

embano1:
LGTM

embano1 (approved):

PR #2234, opened 2021-01-09
Example doc comments update using API docs content

atc0005:
I'm new to the API and this package and am learning my way around both. The main project README helpfully directs the reader to the API docs for more information (thanks for that). That said, is there any interest in adding documentation directly in this package as doc comments? I'm going to guess "no", but I thought it would be worth asking to be sure.

This commit (and subsequent PR) is an example of what I'm referring to. Here, I copied the content 1:1 from the API documentation.

Thanks in advance for your feedback.

refs https://vdc-download.vmware.com/vmwb-repository/dcr-public/a5f4000f-1ea8-48a9-9221-586adff3c557/7ff50256-2cf2-45ea-aacd-87d231ab1ac7/vim.ManagedEntity.html

### DIFF
--- a/vim25/mo/mo.go
+++ b/vim25/mo/mo.go
@@ -1231,6 +1231,12 @@ func init() {
 	t["LocalizationManager"] = reflect.TypeOf((*LocalizationManager)(nil)).Elem()
 }
 
+// ManagedEntity is an abstract base type for all managed objects present in
+// the inventory tree. The base type provides common functionality for
+// traversing the tree structure, as well as health monitoring and other basic
+// functions.
+//
+// Most Virtual Infrastructure managed object types extend this type.
 type ManagedEntity struct {
 	ExtensibleManagedObject
 


dougm:
Thanks @atc0005 . Problem is `vim25/mo/mo.go` is generated, as is `vim25/types/types.go` and `vim25/methods/methods.go`

Ideally the generated code would include the doc you see in the API ref.  However, that doc is generated from an internal model ("vmodl"), whereas the govmomi code is generated from the public wsdl file.  I had started to look at refactoring the code generator in #1186 , but probably won't get to work on it anytime soon.

In the meantime, we could add a package-level `vim25/mo/doc.go` that includes some of the API doc, what do you think?

atc0005:
@dougm reading over #1186 and linked issues it sounds like completing that work will have a nice payoff, but will take a lot of time to complete it. Hopefully you get the space/time to push forward in that direction.

> In the meantime, we could add a package-level vim25/mo/doc.go that includes some of the API doc, what do you think?

This sounds like an idea. In fact, there were a few cases where I went looking for package-level doc coverage after finding it available for another (I don't recall which offhand). If adding the coverage via a doc.go file, is there a way to tie the coverage to specific types (such as `ManagedEntity`) so its available to `gopls`, or is it mostly useful as an intro block of text?

**EDIT**

This may not be the package level doc coverage I was thinking of, but it's one such example: 

- https://pkg.go.dev/github.com/vmware/govmomi@v0.24.0/find
- https://github.com/vmware/govmomi/blob/v0.24.0/find/doc.go

dougm:
Right, find/doc.go is an example of the package level doc I was referring to.  I don't think there's a way to make it available to gopls however.

atc0005:
@dougm thanks for the feedback. Should I create an issue for this work?

dougm:
@atc0005 no need to create a separate issue if you want to create the package doc.go , thanks!

atc0005:
> @dougm: @atc0005 no need to create a separate issue if you want to create the package doc.go , thanks!

Thanks for the feedback, and my apologies that I was (unintentionally) so vague. I was referring to creating a sort of parent issue for tracking the addition of doc.go files for packages missing them, starting with one for `vim25/mo`.

dougm:
No problem, understood. I just meant Pull Requests don't require a parent issue, but feel free to create any tracking issues.

atc0005:
> No problem, understood. I just meant Pull Requests don't require a parent issue, but feel free to create any tracking issues.

Acknowledged, will do.

PR #2236, opened 2021-01-10
vcsim: move product suffix in ServiceContent.About

### LINKED ISSUE #1779: vcsim Terraform support?
I want to use vcsim to test out our Terraform VM templates for VMware. Does vcsim support Terraform VM provisioning? If so is there any reference what kind of Terraform configuration should work against vcsim? In the Terraform template I want to **clone** an existing VM or Template and not just create the VM item. Is this possible with vcsim? I have tried to do this with Terraform against vcsim but Terraform plan just says: `No changes. Infrastructure is up-to-date.`

Any tested Terraform template that actually creates a VM to vcsim with clone functionality?

I found this gist: https://gist.github.com/zpratt/0ad7bcf382b321501bd92e958bf9f8a9
But it doesn't use clone and it actually time outs with error:
`vsphere_virtual_machine.test-node: timeout waiting for an available IP address`

dougm:
Terraform expects About.Name to equal "VMware vCenter Server" to make use of
Content Library features, which fails against vcsim.
Move the "govmomi simulator" suffix to About.FullName to avoid this problem.

- Add "FullName" to 'govc about' output

Fixes #1779


dougm:
cc @Satak

PR #2237, opened 2021-01-12
Expose reflected types for managed objects to users

gh05tn0va:
Problem: User needs to get managed object types by reflection. In current code vim25/mo, the reflected types aren't exposed to user.

Solution: Expose reflected types in vim25/mo, similar to vim25/vsan. Change vsan/mo accordingly.

### DIFF
--- a/vim25/mo/registry.go
+++ b/vim25/mo/registry.go
@@ -1,12 +1,9 @@
 /*
-Copyright (c) 2014 VMware, Inc. All Rights Reserved.
-
+Copyright (c) 2014-2021 VMware, Inc. All Rights Reserved.
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-
     http://www.apache.org/licenses/LICENSE-2.0
-
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -16,6 +13,28 @@ limitations under the License.
 
 package mo
 
-import "reflect"
+import (
+	"reflect"
+	"strings"
+)
 
 var t = map[string]reflect.Type{}
+
+func Add(name string, kind reflect.Type) {
+	t[name] = kind
+}
+
+type Func func(string) (reflect.Type, bool)
+
+func TypeFunc() Func {
+	return func(name string) (reflect.Type, bool) {
+		typ, ok := t[name]
+		if !ok {
+			// The /sdk endpoint does not prefix types with the namespace,
+			// but extension endpoints, such as /pbm/sdk do.
+			name = strings.TrimPrefix(name, "vim25:")
+			typ, ok = t[name]
+		}
+		return typ, ok
+	}
+}
--- a/vim25/mo/registry.go
+++ b/vim25/mo/registry.go
@@ -1,9 +1,12 @@
 /*
 Copyright (c) 2014-2021 VMware, Inc. All Rights Reserved.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


vim25/mo/registry.go
```diff
@@ -1,12 +1,9 @@
 /*
-Copyright (c) 2014 VMware, Inc. All Rights Reserved.
-
```
dougm, line 3: Please just change the year w/o remove the empty lines.

vim25/mo/registry.go
```diff
@@ -16,6 +13,28 @@ limitations under the License.
...
 
 var t = map[string]reflect.Type{}
+
+func Add(name string, kind reflect.Type) {
+	t[name] = kind
+}
+
+type Func func(string) (reflect.Type, bool)
+
+func TypeFunc() Func {
```
dougm, line 29: Is TypeFunc being used somewhere?  The vim25/types `TypeFunc` is used for xml decoding: https://github.com/vmware/govmomi/blob/50c576d6470e7ab9803ceb466ac34f936971dfc1/vim25/soap/client.go#L129

I think you just need the `Add()` func, so the vsan mo types can be inflated via PropertyCollector helpers, right?
But in that case, you wouldn't use the `vsan:` namespace prefix to register.  Can you clarify how your change is being used?

gh05tn0va: I want to get all named fields from a given vmodl type name.
This is possible in rvc(rbvmomi). And we have similar requirements.

```
> type -h
usage: type name
Display information about a VMODL type
  name: VMODL type name
  -h, --help    Show this message

> type HostVsanSystem
Managed Object HostVsanSystem

Properties:
 config: VsanHostConfigInfo

Methods:
 AddDisks_Task(disk : HostScsiDisk[]) : VsanHostDiskMapResult
 EvacuateVsanNode_Task(maintenanceSpec : HostMaintenanceSpec, timeout : int) : void
 FetchVsanSharedSecret() : string
 InitializeDisks_Task(mapping : VsanHostDiskMapping[]) : VsanHostDiskMapResult
 QueryDisksForVsan(canonicalName : string[]) : VsanHostDiskResult
 QueryHostStatus() : VsanHostClusterStatus
 RecommissionVsanNode_Task() : void
 RemoveDiskMapping_Task(mapping : VsanHostDiskMapping[], maintenanceSpec : HostMaintenanceSpec) : VsanHostDiskMapResult
 RemoveDisk_Task(disk : HostScsiDisk[], maintenanceSpec : HostMaintenanceSpec, timeout : int) : VsanHostDiskResult
 UnmountDiskMapping_Task(mapping : VsanHostDiskMapping[]) : VsanHostDiskMapResult
 UpdateVsan_Task(config : VsanHostConfigInfo) : void
```

However, even with this commit, I still can't figure out a way to get **methods** of a given mo type.

dougm: I think we'd need to generate more code to have a mapping from mo type to its method set, and the result type of each _Task method.  Not sure the best way to do that.  Don't want to mess with the existing generator, would prefer to replace it (#1186).  Also would not want to add more generated code to the existing packages with this data.  But would be ok with having in a separate package.  What do you think?

As for this change, the `mo` "registry" is used by methods such as `mo.ObjectContentToType` and `mo.LoadObjectContent`, which convert a PropertyCollector response into Go objects.  Those methods are not namespace aware, so I think you'd want to remove the `vsan:` prefix.

As for reading from the map, the vim25/types `TypeFunc` is used for xml decoding.  Maybe the mo package should have a more direct way to read from the map.  Note that you can currently read from it like so:
```go
	c := types.ObjectContent{Obj: types.ManagedObjectReference{Type: "Datastore"}}
	obj, _ := mo.ObjectContentToType(c)
	fmt.Printf("%T\n", obj)
```

gh05tn0va: Thanks for your explanation! Your advice really helps.
Since we do not use property collector for vsan mo, this change is not necessary.
I also agree that we don't want addtional mappings to the existing package. I think I just need a generated map[string][]string.

Issue #2248, opened 2021-01-19
govc 0.24.0: device.pci.add: Incorrect DeviceID value


yet-another-aligator:
**[Edit] Issue reproduced on a vSphere ESXi 6.7.0u3 host with same behavior.**

Hi,

I'm running govc v0.24.0 with a standalone vSphere ESXi 7.0u1 host and attempted to add a PCI passthru device to a VM by using the `device.pci.add` subcommand, which completed without error.

However, the VM couldn't be powered on : `The systemId does not match the current system or the deviceId and vendorId do not match the device currently at <bus:device.function>`.

The issue could be resolved manually by removing and adding again the PCI device from vSphere web client, as described [there](https://kb.vmware.com/s/article/2001216) but that's not compatible with my wish to automate my VI provisioning.

After a short investigation, I found that an incorrect DeviceID value was provided when adding PCI device to the VM :

---

**govc call**:

```
govc device.pci.add -vm VM-XXXX "0000:65:00.0"
```

---

**The real device ID (decimal value):**

```
govc host.esxcli -json hardware pci list | jq -e -r '.Values[] | select((.DeviceClass[] | select(. == "768")) and (.VendorID[] | select(. == "4318"))) | .DeviceID[0]'
```

**Output** : `7684`

---

**VMX data after adding PCI device with govc:**

```
pciPassthru0.id = "00000:101:00.0"
pciPassthru0.deviceId = "0x7684" <-- Incorrect value
pciPassthru0.vendorId = "0x10de"
pciPassthru0.systemId = "5fe9d4dc-68b5-75f6-e9ae-d45d64bbdaca"
pciPassthru0.present = "TRUE"
```

---

**VMX data after adding PCI device with vSphere Web Client:**

```
pciPassthru0.id = "00000:101:00.0"
pciPassthru0.deviceId = "0x1e04" <-- Correct value
pciPassthru0.vendorId = "0x10de"
pciPassthru0.systemId = "5fe9d4dc-68b5-75f6-e9ae-d45d64bbdaca"
pciPassthru0.present = "TRUE"
```

---

As you can read above, the incorrect DeviceID is actually DeviceID's decimal value but enclosed in an hexadecimal notation.

In `pci.device.add` source code,

https://github.com/vmware/govmomi/blob/1ec59a7c0002598c594fb58e03907d7ff0ca84b4/govc/device/pci/add.go#L133

[...]

PR #2249, opened 2021-01-19
Fix govc device.pci.add: incorrect DeviceID value

yet-another-aligator:
As described in the issue https://github.com/vmware/govmomi/issues/2248,

`DeviceId` string attribute of `VirtualPCIPassthroughDeviceBackingInfo` object should contain the hexadecimal value of PCI device's DeviceID instead of decimal value.

Passing the decimal value does not lead to a govc execution error but prevents the targeted virtual machine boot as the real PCI device's DeviceID does not match the one stored in virtual machine VMX data.

This fix has been tested on 7.0u1 and 6.7.0u3 ESXi hosts and resolved the issue on both 2 hosts.

### DIFF
--- a/govc/device/pci/add.go
+++ b/govc/device/pci/add.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"strconv"
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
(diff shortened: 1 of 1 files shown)


dougm (approved):
lgtm, thanks @yet-another-aligator 

cc @leslie-qiwa

leslie-qiwa:
cool! @yet-another-aligator thanks for fixing it.

yet-another-aligator:
Thanks @dougm @leslie-qiwa 😄

PR #2255, opened 2021-01-26
govc: fix session curl when given a URL query

dougm:
The session.login PATH command set URL.Path directly, causing query string to be escaped.
Switch to url.Parse which properly handles both Path and the optional Query.

- Add more docs on session curl

- Add session curl POST test, including URL query.

### DIFF
--- a/govc/session/login.go
+++ b/govc/session/login.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


govc/session/login.go
```diff
@@ -105,6 +106,10 @@ The session.login command can be used to:
 - Avoid passing credentials to other govc commands
 - Send an authenticated raw HTTP request
 
+The session.login command can be used for authenticated curl-style HTTP requests when a PATH arg is given.
+PATH may also contain a query string. The '-u' flag (GOVC_URL) is used for the URL scheme, host and port.
+The request method (-X) defaults to GET. When set to POST or PATCH, a request body must be provided via stdin.
```
fstrudel, line 111: POST, PUT and PATCH?

fstrudel (approved):

govc/session/login.go
dougm: Yes, added PUT

PR #2256, opened 2021-01-26
Remove duplicate cns bindings from vsan directory

chethanv28:
This PR is removing duplicate cns bindings in vsan folder. This is need to fix panic observed in vsphere csi driver repo which consumes govmomi 
```
2021/01/26 10:41:45 http: panic serving 127.0.0.1:55874: reflect: Call using *types.CnsCreateVolume as type *types.CnsCreateVolume
goroutine 136 [running]:
net/http.(*conn).serve.func1(0xc000120a00)
        /usr/local/go/src/net/http/server.go:1801 +0x147
panic(0x293bc60, 0xc000bab250)
        /usr/local/go/src/runtime/panic.go:975 +0x47a
```
The duplicate bindings are added by the PR: https://github.com/vmware/govmomi/pull/2184/


dougm (commented):
I think the vsan/{types,methods,mo} packages are all generated ( cc @gh05tn0va ), but cns is currently hand written?  Can the cns package use the vsan generated types?  Or could the vsan generator filter out Cns @gh05tn0va ?

chethanv28:
> I think the vsan/{types,methods,mo} packages are all generated ( cc @gh05tn0va ), but cns is currently hand written? Can the cns package use the vsan generated types? Or could the vsan generator filter out Cns @gh05tn0va ?

@dougm @divyenpatel @gh05tn0va 
If its fine from your end, I have created a separate issue to track the changes needed to either filter out cns or make cns package use the generated vsan types.
Can we merge this PR provided we will follow-up using this issue: https://github.com/vmware/govmomi/issues/2258

dougm:
That'd be fine with me to merge this for the short-term and address later for the long-term.

divyenpatel (approved):

divyenpatel:
Generated CNS bindings were missing some internal APIs and some internal types. 

If we go with auto-generated bindings in the future, we need to make sure all types and APIs used by the CSI driver are generated and published, otherwise the vSphere CSI driver will break.

Issue #2116, opened 2020-09-14
govc: Unset MAC address / change addressType


lukasluecke:
#282 added the `net.address` parameter to `vm.network.change`, which allows setting a MAC address. Is there a way to do the opposite, i.e. remove the MAC address, and go back to `generated` / `assigned` for `addressType`? Maybe allowing these as magic string values and changing the addressType respectively?

dougm:
Haven't tried it, but this might work:
```diff
modified   govc/flags/network.go
@@ -143,5 +143,8 @@ func (flag *NetworkFlag) Change(device types.BaseVirtualDevice, update types.Bas
 
 	if changed.AddressType != "" {
 		current.AddressType = changed.AddressType
+		if current.AddressType == string(types.VirtualEthernetCardMacTypeGenerated) {
+			current.AddressType = ""
+		}
 	}
 }
```

semora81:
Is there an example of how to use this from the command line? i'm trying to do the exact same thing where I need to set a vm's addresstype back to "automatic" 

thanks

semora81:
#2265 worked like a charm, i've got the call setup as : 
```
govc vm.network.change -vm=testvm -net "Management" -net.address="" ethernet-0       
```
it has set the value to automatic and at next startup of the vm, it auto assigned a mac.

dougm:
Fixed by #2265

PR #2265, opened 2021-02-02
Change the address type to automatic

saamalik:
Change the address type to automatic

Addresses #2116.

### DIFF
--- a/govc/flags/network.go
+++ b/govc/flags/network.go
@@ -121,13 +121,13 @@ func (flag *NetworkFlag) Device() (types.BaseVirtualDevice, error) {
 		return nil, err
 	}
 
-	if flag.address != "" {
-		card := device.(types.BaseVirtualEthernetCard).GetVirtualEthernetCard()
-		card.AddressType = string(types.VirtualEthernetCardMacTypeManual)
-		card.MacAddress = flag.address
-	} else {
+	if flag.address == "-" {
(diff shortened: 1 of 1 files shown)


govc/flags/network.go
```diff
@@ -125,6 +125,10 @@ func (flag *NetworkFlag) Device() (types.BaseVirtualDevice, error) {
 		card := device.(types.BaseVirtualEthernetCard).GetVirtualEthernetCard()
 		card.AddressType = string(types.VirtualEthernetCardMacTypeManual)
 		card.MacAddress = flag.address
+	} else {
+		card := device.(types.BaseVirtualEthernetCard).GetVirtualEthernetCard()
+		card.AddressType = string(types.VirtualEthernetCardMacTypeGenerated)
+		card.MacAddress = flag.address
```
dougm, line 131: Thanks @saamalik .  We don't want to change `AddressType` if `flag.address` is empty.  If the intention is to have `MacAddress` empty in this case, we could use a special value for `flag.address`, such as `-`, which indicates use of default in several other govc flags.  For example:
```go
} else if flag.address == "-" {
    card := device.(types.BaseVirtualEthernetCard).GetVirtualEthernetCard()
    card.AddressType = string(types.VirtualEthernetCardMacTypeGenerated)
    card.MacAddress = ""
}

saamalik:
@dougm changed. Please review.

dougm (approved):
Thanks @saamalik

PR #2268, opened 2021-02-04
Update .goreleaser.yml for support a new cpu architecture 

houfangdong:
hello，I am going to submit mips64le architecture，The main reasons for adding support for mips64le architecture are:
1、The mips64le architecture is also the mainstream cpu architecture (amd64, arm64), which is used by many users;
2、Golang supports cross compilation, and mips64le is officially supported.
3、This project is an underlying dependency library of the kubernetes project  [k/k](https://github.com/kubernetes/kubernetes), therefore I wish to submit this patch.
Therefore, I hope that the release package also supports the mips64le architecture, which is convenient for more users.
Signed-off-by: houfangdong houfangdong@loongson.com

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,6 +11,7 @@ builds:
     - amd64
     - 386
     - arm64
+    - mips64le
   env:
     - CGO_ENABLED=0
   main: ./govc/main.go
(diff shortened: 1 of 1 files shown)


dougm (commented):
Thanks @houfangdong 
We can include this, but goreleaser is currently unused until #1375 is done.
The govc release process is currently done using: https://github.com/vmware/govmomi/blob/854ba4efe87ecd9bf34d4bc466119c1748e8b7ef/govc/build.sh#L16

houfangdong:
Thanks，@dougm 
I see. and if you have any questions about mips64le, please contact me in time.

Issue #2243, opened 2021-01-15
How to get paired name/size values for each snapshot associated with a VM?


atc0005:
Here is a PowerCLI snippet (either or): 

```powershell
Get-VM testVM | Get-Snapshot | Measure-Object -Sum SizeMB
Get-Snapshot -VM testVM | Measure-Object -Sum SizeMB
```

which helped me generate a list of snapshot details like this from a Nagios plugin that I wrote about a year ago: 

```
OK: 4 snapshots found for non-excluded VMs, but configured thresholds not exceeded [WARNING: 30 GB, CRITICAL: 50 GB]

* 'testVM' [Age: 291, SnapSize: 13.861 GB, Combined SnapSize: 28.551 GB, SnapName: '2020-03-29']
* 'testVM' [Age: 406, SnapSize: 6.251 GB, Combined SnapSize: 28.551 GB, SnapName: '2019-12-05']
* 'testVM' [Age: 457, SnapSize: 8.438 GB, Combined SnapSize: 28.551 GB, SnapName: '2019-10-15']
* 'testVM' [Age: 1045, SnapSize: 0.000 GB, Combined SnapSize: 28.551 GB, SnapName: 'Fresh install, activation and patches']
```

I'm looking to do the same thing with govmomi, but after digging the closest
I've found is #1197. I expect I'm overlooking something obvious.

Is the approach here to loop over `vm.LayoutEx.File` (where `vm` is a
`mo.VirtualMachine`), filter type to `snapshotData` and build a mapping of sizes and
keys, then loop over `vm.Snapshot.RootSnapshotList` and pair together the id
values here with key values from the other structure?

Thanks in advance for any pointers.

atc0005:
I think I found the correlation. I'll work on this over the next few days and will link to the results in case it is helpful to someone else.

atc0005:
Scratch notes, since this seem as good a place as any to place them.

---

## Hierarchy of types

Levels above and some below skipped for simplicity and due to my ignorance.

- `VirtualMachine`
  - `VirtualMachineSnapshotInfo`
    - `VirtualMachineSnapshotTree`
  - `VirtualMachineFileLayoutEx`
    - `VirtualMachineFileLayoutExFileInfo`
    - `VirtualMachineFileLayoutExSnapshotLay`

## Fields expanded

- `vm`
  - `Name` (**Used in output**)
  - `Snapshot`
    - `RootSnapshotList` (`VirtualMachineSnapshotTree`)
[...]

PR #2269, opened 2021-02-04
Add govc snapshot size command, standalone SnapshotSize function

atc0005:
Add support for snapshot size calculations

- govc: Add size (`s`) flag to `snapshot.tree` subcommand
- Add `object.SnapshotSize` and helper functions
  to support calculating snapshot size from other
  client code

Credit to @dougm for providing successive code examples,
explanation and guidance for understanding the required
steps to implement this support.

As noted on GH-2243, code provided by @dougm was (AFAIK)
based heavily on existing C# code for PowerCLI
implementation of the `Get-Snapshot` cmdlet.

- credit: @dougm
- fixes vmware/govmomi#2243

### DIFF
--- a/govc/vm/snapshot/tree.go
+++ b/govc/vm/snapshot/tree.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+Copyright (c) 2021 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -25,6 +25,8 @@ import (
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/units"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 )
@@ -38,8 +40,10 @@ type tree struct {
 	description bool
 	fullPath    bool
 	id          bool
+	size        bool
 
-	info *types.VirtualMachineSnapshotInfo
+	info   *types.VirtualMachineSnapshotInfo
+	layout *types.VirtualMachineFileLayoutEx
 }
 
 func init() {
@@ -59,6 +63,7 @@ func (cmd *tree) Register(ctx context.Context, f *flag.FlagSet) {
 	f.BoolVar(&cmd.fullPath, "f", false,
 		"Print the full path prefix for snapshot")
 	f.BoolVar(&cmd.id, "i", false, "Print the snapshot id")
+	f.BoolVar(&cmd.size, "s", false, "Print the snapshot size")
 }
 
 func (cmd *tree) Description() string {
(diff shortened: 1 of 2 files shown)


atc0005:
@dougm Setting this as a draft since I've not squashed commits and finished cleaning up the branch yet.

What are your preferences regarding code comments? I assumed that the project style was to omit them unless necessary and have pruned the majority of them back. a54302c4f266d64fda14d9970e2a710d0657ee62 represents that state, except for a few lines I wanted to get your feedback on.

7dfb85b21fd30b889b17a8cd88f1b754edc9c145 represents a version of the code with comments, but without the "note to self" verbosity that I used to begin with as I explored the code. Is this form useful?

govc/vm/snapshot/tree.go
```diff
@@ -161,8 +174,15 @@ func (cmd *tree) Run(ctx context.Context, f *flag.FlagSet) error {
 	}
 
 	cmd.info = o.Snapshot
+	cmd.layout = o.LayoutEx
+
+	// @dougm: This was in the original code snippets shared, but I assume
+	// isn't desired in a final PR?
+	//
+	// size := SnapshotSize(o.Reference(), nil, o.LayoutEx, true)
```
atc0005, line 182: This comment block is what I made reference to in my last comment.

govc/vm/snapshot/size.go
```diff
@@ -0,0 +1,95 @@
+/*
+Copyright (c) 2016 VMware, Inc. All Rights Reserved.
```
dougm, line 2: 2021

govc/vm/snapshot/size.go
```diff
@@ -0,0 +1,95 @@
...
+func removeKey(l *[]int, key int) {
+	for i, k := range *l {
+		if k == key {
+			*l = append((*l)[:i], (*l)[i+1:]...)
+			break
+		}
+	}
+}
+
+func SnapshotSize(info types.ManagedObjectReference, parent *types.ManagedObjectReference, vmlayout *types.VirtualMachineFileLayoutEx, isCurrent bool) int {
```
dougm, line 44: This could be in the `object` package so it can be used outside of govc (without pulling in govc's baggage).
Probably add it along side the snapshot related helpers here: https://github.com/vmware/govmomi/blob/854ba4efe87ecd9bf34d4bc466119c1748e8b7ef/object/virtual_machine.go#L635
Leaving it as a function is ok too, rather than VirtualMachine methods like the others.
And regarding comments, it'd be good to include doc for this func since it'll show up in the godoc.

govc/vm/snapshot/tree.go
dougm: Right, this can be removed.

dougm (commented):
Thanks @atc0005 !
Code comments are welcome, especially in this case where vSphere's own docs don't help much.

atc0005:
@dougm I've made a pass to incorporate suggested changes. I've added light doc comments to the functions themselves, but not to the function body. Are those welcome, or is it preferred to keep them specific to what godoc exposes?

dougm:
> @dougm I've made a pass to incorporate suggested changes. I've added light doc comments to the functions themselves, but not to the function body. Are those welcome, or is it preferred to keep them specific to what godoc exposes?

Thanks @atc0005 , looks great.  In general, I think it's preferred to have godoc comments convey the API functionality/contract and function body comments to document the implementation details. Looks like your current PR aligns with that. But I'm not too picky either, since personally I tend to read the source code as the source of truth, regardless of any API doc :)

atc0005:
@dougm Squashed commits and rebased on current `master` branch content. Ready for review.

dougm (approved):
Awesome, thanks @atc0005 !

atc0005:
> Awesome, thanks @atc0005 !

Thank *you* for all of the help with this! I learned a lot along the way.

PR #2272, opened 2021-02-08
Add vSAN 7.0U1 release constant

RaunakShah:
This PR adds a constant for vSAN 7.0U1 release. This version will be used by CSI, just like for older versions.

Screen Shot 2021-02-08 at 2 11 49 PM

### DIFF
--- a/cns/client.go
+++ b/cns/client.go
@@ -36,6 +36,7 @@ const (
 const (
 	ReleaseVSAN67u3 = "vSAN 6.7U3"
 	ReleaseVSAN70   = "7.0"
+	ReleaseVSAN70u1 = "vSAN 7.0U1"
 )
 
 var (


cns/client.go
```diff
@@ -36,6 +36,7 @@ const (
 const (
 	ReleaseVSAN67u3 = "vSAN 6.7U3"
 	ReleaseVSAN70   = "7.0"
+	ReleaseVSAN70u1 = "vSAN 7.0U1"
```
SandeepPissay, line 39: Since you are not using this constant in this repo, why do you need this constant? Wouldn't linter fail when you add a var and not use it?

RaunakShah: I'm planning to use this in a PR in CSI - https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/642/files#diff-f3c4e2561d16ec3070bae674afe9d3afb7c7155fcfa186d8256da34c10f048b1R593
and i thought adding it here maintains consistency. 
AFAIK linter checks dont fail for unused constants that are exported by the package.

dougm: Fine with me @RaunakShah , thanks.

dougm (approved):

PR #2274, opened 2021-02-09
vim25: fix race in TemporaryNetworkError retry func

### LINKED ISSUE #2266: Dara race at vim25/retry.go:49
Tried to use vim25.retry in our code to retry the request but I was seeing Data Race at variable n. 
https://github.com/vmware/govmomi/blob/master/vim25/retry.go#L49

As we share retry Func object across diff requests once request 1 reaches max retry limit, all other requests also reach max limit. We use only one vimClient for all the requests which can be concurrent.
I think we should add _retryCounter_  in  **retry struct** and set it to 0 for new request.

mbhadale:
To maintain backward compatibility introduced new optional
variable in retry struct and new function with parameter
retryAttempts.

- Add vim25.RetryTemporaryNetworkError to replace deprecated vim25.TemporaryNetworkError

- Export IsTemporaryNetworkError function

- Add test to retry network error

Fixes #2266

### DIFF
--- a/govc/flags/client.go
+++ b/govc/flags/client.go
@@ -371,7 +371,7 @@ func (flag *ClientFlag) Client() (*vim25.Client, error) {
 
 	// Retry twice when a temporary I/O error occurs.
 	// This means a maximum of 3 attempts.
-	c.RoundTripper = vim25.Retry(c.Client, vim25.TemporaryNetworkError(3))
+	c.RoundTripper = vim25.Retry(c.Client, vim25.RetryTemporaryNetworkError, 3)
 	flag.client = c
 
 	return flag.client, nil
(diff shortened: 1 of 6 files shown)


dougm (commented):
Thanks @mbhadale , looks good. But does this fix the race described in the bug, with the TemporaryNetworkError function?  Do you have a simple way to reproduce that race we could use in the tests?

dougm:
@mbhadale can you review the commit I pushed?  If it looks ok, I can squash and merge with your changes. Thanks.

mbhadale:
Tha

> , I can squash and merg

mbhadale:
Change looks good to me. Thanks for making the change.

PR #2287, opened 2021-02-17
Update cluster.override.change help message

Knappek:
Since vSphere API 6.5 it is possible to set cluster HA restart priority also to `lowest` and `highest` (see [API ref](https://vdc-download.vmware.com/vmwb-repository/dcr-public/da47f910-60ac-438b-8b9b-6122f4d14524/16b7274a-bf8b-4b4c-a05e-746f2aa93c8c/doc/vim.cluster.DasVmSettings.RestartPriority.html)). 

This is only missing in the CLI help message and in the usage documentation, the command itself using `highest` or `lowest` is already working.

### DIFF
--- a/govc/cluster/override/change.go
+++ b/govc/cluster/override/change.go
@@ -52,10 +52,12 @@ func (cmd *change) Register(ctx context.Context, f *flag.FlagSet) {
 
 	// HA
 	rp := []string{
-		string(types.DasVmPriorityDisabled),
-		string(types.DasVmPriorityLow),
-		string(types.DasVmPriorityMedium),
-		string(types.DasVmPriorityHigh),
+		string(types.ClusterDasVmSettingsRestartPriorityDisabled),
+		string(types.ClusterDasVmSettingsRestartPriorityLowest),
+		string(types.ClusterDasVmSettingsRestartPriorityLow),
+		string(types.ClusterDasVmSettingsRestartPriorityMedium),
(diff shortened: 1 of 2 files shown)


vim25/types/enum.go
```diff
@@ -650,9 +650,11 @@ type DasVmPriority string
 
 const (
 	DasVmPriorityDisabled = DasVmPriority("disabled")
+	DasVmPriorityLowest   = DasVmPriority("lowest")
```
dougm, line 653: The vim25/types package is generated from API's WSDL, so this change would be overwritten.

govc/cluster/override/change.go
```diff
@@ -53,9 +53,11 @@ func (cmd *change) Register(ctx context.Context, f *flag.FlagSet) {
 	// HA
 	rp := []string{
 		string(types.DasVmPriorityDisabled),
```
dougm, line 55: Based on the API ref link you posted, we should be using `types.ClusterDasVmSettingsRestartPriority*` for all the options in this list.  Values are the same, but see comment below on the generated code.

dougm (commented):
Thanks @Knappek .  Looks like we were using the wrong enum, can you also switch to that in this PR?

Knappek:
Thanks @dougm for reviewing. I see, that makes sense. I have fixed it and squashed the commits in order to have a clean commit history.

dougm (approved):
Thanks again @Knappek

PR #2291, opened 2021-02-21
Fixes in SOAP download progress

yelly:
Fixing a bug that caused a progress reader to be created but never used in soap.client.WriteFile.
In addition, treating the situation in which a reader size is invalid and returning 0 as the progress in those cases. I suppose -1 might also work here.

### DIFF
--- a/vim25/progress/reader.go
+++ b/vim25/progress/reader.go
@@ -36,6 +36,9 @@ type readerReport struct {
 }
 
 func (r readerReport) Percentage() float32 {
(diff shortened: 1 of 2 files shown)


dougm (approved):
Thanks @yelly , lgtm. We can merge when vmwclabot approves.

yelly:
Awesome, just waiting for legal here.

dougm:
Thanks @yelly !

Issue #2278, opened 2021-02-11
govc json output, no error provided


matiniamirhossein:
When running commands with -json=true there is no output on errors:

(data already provided with env)

```
root@ubuntu:~# govc env
GOVC_USERNAME=root
GOVC_PASSWORD=test
GOVC_URL=1.2.3.4
GOVC_INSECURE=true

root@ubuntu:~# govc about
govc: ServerFaultCode: Cannot complete login due to an incorrect user name or password.
root@ubuntu:~# govc about -json=true
{}
```

Is there any options on this?

embano1:
The issue is due to the private `fault` field in `soapFaultError` which is not marshalled to JSON by default:

```go
type soapFaultError struct {
	fault *Fault
}
```

I tried a quick fix (see patch below) and this is the result:

```bash
./govc about -json
{
  "Fault": {
    "XMLName": {
      "Space": "http://schemas.xmlsoap.org/soap/envelope/",
      "Local": "Fault"
    },
    "Code": "ServerFaultCode",
    "String": "Cannot complete login due to an incorrect user name or password.",
    "Detail": {
      "Fault": {
        "FaultCause": null,
        "FaultMessage": null
      }
    }
  }
}
```

Patch:

```
diff --git a/vim25/soap/error.go b/vim25/soap/error.go
index 46111556..41593191 100644
--- a/vim25/soap/error.go
+++ b/vim25/soap/error.go
@@ -17,6 +17,7 @@
 package soap
 
 import (
+	"encoding/json"
 	"fmt"
 	"reflect"
 
@@ -35,6 +36,16 @@ type soapFaultError struct {
 	fault *Fault
 }
 
+func (s soapFaultError) MarshalJSON() ([]byte, error) {
+	out := struct {
+		Fault *Fault
+	}{
+		Fault: s.fault,
+	}
+
+	return json.Marshal(out)
+}
+
 func (s soapFaultError) Error() string {
 	msg := s.fault.String
```

cc/ @dougm

dougm:
Thanks @embano1 , lgtm.  Another option would be:
```diff
modified   govc/flags/output.go
@@ -165,7 +165,7 @@ func (flag *OutputFlag) WriteResult(result OutputWriter) error {
 		if flag.formatIndent {
 			e.SetIndent("", "  ")
 		}
-		err = e.Encode(result)
+		err = e.Encode(dumpValue(result))
 	case flag.XML:
 		e := xml.NewEncoder(flag.Out)
[...]

PR #2303, opened 2021-03-03
Marshal soapFaultError as JSON

embano1:
govc optionally allows JSON formatted output. In certain cases, e.g.
when a SOAP error is thrown, that causes the JSON output to be empty due
to unexported fields in the struct.

Closes: #2278


dougm (approved):

dougm:
Thanks @embano1 !

PR #2310, opened 2021-03-05
Update vcsim Dockerfile

### LINKED ISSUE #2045: vcsim will not run in scratch container
In a `scratch` container there is no `/tmp` directory.

https://github.com/vmware/govmomi/blob/cfbe1d59b449be9b5378251bc6e52c65bc5e1cbd/Dockerfile.vcsim#L1-L4

The following `Dockerfile` fails with `stat /tmp: no such file or directory`

```
FROM golang:1.14 as build
WORKDIR /go/src/app
RUN CGO_ENABLED=0 go get -u github.com/vmware/govmomi/vcsim

FROM scratch
COPY --from=build /go/bin/vcsim /vcsim
ENTRYPOINT ["/vcsim"]
```

Copying over the `/tmp` directory (for fun), I get:

```
mkdir /tmp/govcsim-ha-datacenter-LocalDS_0-110559439: permission denied
```

jnovack:
Fixes #2045
Allows vcsim to be built within the docker container rather than built externally and copied in.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -68,7 +68,7 @@ brews:
   - name: govc
     ids:
       - govc
-    github:
+    tap:
       owner: govmomi
       name: homebrew-tap
     commit_author:
@@ -84,7 +84,7 @@ brews:
   - name: vcsim
     ids:
       - vcsim
-    github:
+    tap:
       owner: govmomi
       name: homebrew-tap
     commit_author:
--- a/Dockerfile.govc
+++ b/Dockerfile.govc
@@ -1,4 +1,43 @@
+# Create a builder container
+# golang:1.16.0-buster amd64
+FROM golang@sha256:f254180c5defa2653955e963fb0626e3d4fbbb162f7cff6490e94607d1d867ff AS build
+WORKDIR /go/src/app
+
+# Create appuser to isolate potential vulnerabilities
+# See https://stackoverflow.com/a/55757473/12429735
+ENV USER=appuser
+ENV UID=10001
+RUN adduser \
+    --disabled-password \
+    --gecos "" \
+    --home "/nonexistent" \
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid "${UID}" \
+    "${USER}"
+
+# Create a new tmp directory so no bad actors can manipulate it
+RUN mkdir /temporary-tmp-directory && chmod 777 /temporary-tmp-directory
+
+###############################################################################
+# Final stage
 FROM scratch
-LABEL maintainer="fabio@vmware.com"
-COPY govc /
-ENTRYPOINT [ "/govc" ]
+
+# Run all commands as non-root
+USER appuser:appuser
+
+# Allow container to use latest TLS certificates
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+# Copy over appuser to run as non-root
+COPY --from=build /etc/passwd /etc/passwd
+COPY --from=build /etc/group /etc/group
+
+# Copy over the /tmp directory for golang/os.TmpDir
+COPY --chown=appuser --from=build /temporary-tmp-directory /tmp
+
+# Copy application from external build
+COPY govc /govc
+
+# Set CMD to application with container defaults
+CMD ["/govc"]
--- a/Dockerfile.vcsim
+++ b/Dockerfile.vcsim
@@ -1,4 +1,46 @@
+# Create a builder container
+# golang:1.16.0-buster amd64
+FROM golang@sha256:f254180c5defa2653955e963fb0626e3d4fbbb162f7cff6490e94607d1d867ff AS build
+WORKDIR /go/src/app
+
+# Create appuser to isolate potential vulnerabilities
+# See https://stackoverflow.com/a/55757473/12429735
+ENV USER=appuser
+ENV UID=10001
+RUN adduser \
+    --disabled-password \
+    --gecos "" \
+    --home "/nonexistent" \
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid "${UID}" \
+    "${USER}"
+
+# Create a new tmp directory so no bad actors can manipulate it
+RUN mkdir /temporary-tmp-directory && chmod 777 /temporary-tmp-directory
+
+###############################################################################
+# Final stage
 FROM scratch
-LABEL maintainer="fabio@vmware.com"
-COPY vcsim /
-ENTRYPOINT [ "/vcsim" ]
+
+# Run all commands as non-root
+USER appuser:appuser
+
+# Allow container to use latest TLS certificates
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+# Copy over appuser to run as non-root
+COPY --from=build /etc/passwd /etc/passwd
+COPY --from=build /etc/group /etc/group
+
+# Copy over the /tmp directory for golang/os.TmpDir
+COPY --chown=appuser --from=build /temporary-tmp-directory /tmp
+
+# Expose application port
+EXPOSE 8989
+
+# Copy application from external build
+COPY vcsim /vcsim
+
+# Set entrypoint to application with container defaults
+ENTRYPOINT ["/vcsim", "-l", "0.0.0.0:8989"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+Dockerfile*
+.*ignore


embano1:
Thx! QQ: did you check the contents of tmp you copy from the builder, ie is it empty when you copy?

Btw: we're in the process of completely overhauling the release process, incl. Docker builds. For example, I'm considering merging the Dockerfiles to simplify the build and user experience.

Dockerfile.vcsim
```diff
@@ -1,4 +1,9 @@
...
+WORKDIR /go/src/app
+RUN CGO_ENABLED=0 go get -u github.com/vmware/govmomi/vcsim
+
 FROM scratch
-LABEL maintainer="fabio@vmware.com"
-COPY vcsim /
-ENTRYPOINT [ "/vcsim" ]
\ No newline at end of file
+ENV TMPDIR=/tmp
+COPY --from=build /tmp /tmp
```
embano1, line 7: Is this an empty dir? Btw: this might not work when running as non-root. It would be great to get non-root mode in this PR as well, if possible since several platforms these days enforce non-root execution.

embano1: just saw your comment in the issue:

COPY --chown=appuser --from=build /tmp /tmp

can you please add a vcsim user/group and copy /etc/passwd as well and assume this role in scratch? 🙏

jnovack: I have revamped the Dockerfile.  Yes, this is an empty directory.

image

jnovack:
```
docker run --rm -it \
    -v /var/run/docker.sock:/var/run/docker.sock \
    wagoodman/dive:latest IMAGE:TAG
```

Give the PR a build, then run the previous command replacing `IMAGE:TAG` with the name you built. (e.g. `vcsim:pr2310`)

jnovack:
> Btw: we're in the process of completely overhauling the release process, incl. Docker builds. For example, I'm considering merging the Dockerfiles to simplify the build and user experience.

I wouldn't.  You want the VCSIM built separately (although, with every update).  As a user who codes against vCenter, I want to `docker pull vmware/vcsim:latest` and start it up with my config `docker run ...` or `docker-compose ...` as necessary. 

The `vcsim` container **SHOULD** (see RFC 2118) be standalone and built each release for maximum user benefit.

Dockerfile.vcsim
```diff
@@ -1,4 +1,48 @@
+# Create a builder container
+FROM golang:latest as build
```
embano1, line 2: Since we're talking security vulnerabilities here (+1 on this), can we please pin the builder image then via sha-digest as a best practice?

E.g.:

```Dockerfile
# golang:1.16.0-buster amd64
FROM golang@sha256:f254180c5defa2653955e963fb0626e3d4fbbb162f7cff6490e94607d1d867ff AS build
```

Dockerfile.vcsim
```diff
@@ -1,4 +1,48 @@
...
+    --disabled-password \
+    --gecos "" \
+    --home "/nonexistent" \
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid "${UID}" \
+    "${USER}"
+
+# Update ca-certificates, for final stage.
+RUN apt-get update && apt-get install -y ca-certificates
```
embano1, line 19: If we pin the image digest there's no need to run `apt-get` which could otherwise lead to non-deterministic builds again.

embano1:
> I wouldn't. You want the VCSIM built separately (although, with every update). As a user who codes against vCenter, I want to docker pull vmware/vcsim:latest and start it up with my config docker run ... or docker-compose ... as necessary.

Thx for the feedback. This use case would still be possible. What I was thinking is that likely, as a `vcsim` user, you might also want `govc` in the same image (`vcsim` would be `ENTRYPOINT` though). So the experience for you would not change but you'd also get `govc` with the same image since the go hand in hand, especially for troubleshooting.

> The vcsim container SHOULD (see RFC 2118) be standalone and built each release for maximum user benefit.

Did you reference the correct RFC here (finding Microsoft Point-To-Point Compression (MPPC) Protocol)? Since official govmomi images are currently only built on a release tag, this naturally includes govc as part of the repo. Why would that not be considered a benefit then?

Btw: thx for the work and feedback.

jnovack:
> Thx for the feedback. 

Just dialoging, trying to mentally work through scenarios.  I appreciate the back-and-forth.

>  you might also want govc in the same image.

6 of one, half-a-dozen of the other.  Potato, Potatoe.   The OCD part of me would want two separate containers, `vmware/govc:latest` and `vmware/vcsim:latest`.  If they are smashed into a single container, running in one place is going to be awkward.

```
> docker ps
CONTAINER ID   IMAGE                COMMAND         CREATED       STATUS        NAMES
4eae5f9bf2f5   vmware/govc:latest   "/vcsim"        8 days ago    Up 8 days     govc-or-vcsim-who-knows
```

...or...

```
> docker run -it --rm --entrypoint vmware/vcsim:latest govc
```

It just seems a smidge confusing at first glance.  [Why not both?](https://i.imgur.com/C1z7Wtd.jpg)  Double your pleasure, double your fun, with double-mint containers!

> Did you reference the correct RFC here (finding Microsoft Point-To-Point Compression (MPPC) Protocol)? 

I did not.   https://tools.ietf.org/html/rfc2119

Dockerfile.vcsim
jnovack: complete.

Dockerfile.vcsim
```diff
@@ -1,4 +1,45 @@
+# Create a builder container
```
embano1, line 1: Can we please specify which container tag/arch this references so it's more clear when upgrading, like in the example I gave?

Dockerfile.vcsim
```diff
@@ -1,4 +1,45 @@
...
+    --disabled-password \
+    --gecos "" \
+    --home "/nonexistent" \
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid "${UID}" \
+    "${USER}"
+
+# Build
+RUN CGO_ENABLED=0 go get -u github.com/vmware/govmomi/vcsim
```
embano1, line 19: Hm, what are you expecting this to do? Go mod versioning would come into your way here...why not build as usual instead of `go get`? (see for example the Dockerfile I gave in the discussion)

Dockerfile.vcsim
```diff
@@ -1,4 +1,45 @@
...
+
+# Allow container to use latest TLS certificates
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+# Copy over appuser to run as non-root
+COPY --from=build /etc/passwd /etc/passwd
+COPY --from=build /etc/group /etc/group
+
+# Copy over the /tmp directory for golang/os.TmpDir
+COPY --chown=appuser --from=build /tmp /tmp
```
embano1, line 33: nit: if you don't mind, let's create a dedicated `tmp` directory and copy this over in the build sage. Background: one does not know what certain RUN/build steps in the build stage might put in `/tmp` over time thus unintentionally sneaking in files or folders in the final image.

Dockerfile.vcsim
```diff
@@ -1,4 +1,45 @@
...
+
+# Run all commands as non-root
+USER appuser:appuser
+
+# Application runs on local port
+EXPOSE 8989
```
embano1, line 39: QQ: do you think we should hard-code or make this one configurable, e.g. via ENV?

Dockerfile.vcsim
jnovack: Yes, sorry, I misunderstood that.

Dockerfile.vcsim
jnovack: > what are you expecting this to do?

This downloads and builds the application.

> .why not build as usual instead of go get?

I'm building this inside the `build` container.  Then copying the binary to the `scratch` container.  I don't have `go` on my machine, and I don't expect Dockerhub to or anyone who wants to build this to have it either.

It looks like `Dockerfile.govc` is building on the host, and then copying the binary into the new container.

Am I missing a point you are trying to make?  I'm sorry, I haven't had my coffee yet.  Can you explain further?

Dockerfile.vcsim
jnovack: valid.

Dockerfile.vcsim
jnovack: If the goal is to test vcsim, then testing changing ports is a valid test.

Dockerfile.vcsim
jnovack: I see what you mean.  Resolved.

jnovack:
All comments have been resolved with the exception of the user-defined port.

The scratch container does not invoke a shell, thus cannot expand `${VCSIM_PORT}`.  I consider this a non-issue because this is an image.  Need it on a different port?  Map to the image port 8989. (e.g. `docker run ... -p 1234:8989 ...`)

dougm (approved):
Thanks @jnovack , lgtm.

@embano1 ?

embano1:
Sorry, please hold this one. This change is great, but it will break the current goreleaser process. Will talk with @jnovack about some additional changes which would be great to get in one PR.

jnovack:
Not a problem, I'm happy to join a slack/teams/discord channel/call if necessary.

embano1:
Perfect, that'd be easier. Background: we need to work on the goreleaser stuff and also govc Dockerfile to not break the release process but getting  these great improvements in. Can I reach you on the public govmomi VMware Code Slack?

embano1:
Summary from our quick call:

- [ ] Use build ARG for injecting Git Commit ID (`flags.GitVersion={{.Version}}` linker flag) to container
- [ ] rename `github` section in goreleaser to `tap`
- [ ] apply changes for `Dockerfile.govc`
- [ ] build/release verification so we don't break current release flow

cc/ @dougm 
Idea is to give users the possibility to build their own Docker images from Dockerfiles we provide for vcsim/govc. That's exactly Justin's use case (no Go/goreleaser installed locally). With the upcoming changes in this PR this will be possible and goreleaser process will use the updated Dockerfiles so we don't break the existing build/release process. We'll make minor quality improvements to the Docker images produced (labels, BUILD related information, etc.) and will mimic the goreleaser artifact creation inside the container's build phase (linker flags, etc.).

jnovack:
There's a fundamental problem with our discussion this morning.  GoReleaser's **Docker** module does not do what we want it to do.  It's not designed for it.

It appears a docker container temporary directory structure is built from a specific build (in this case `linux/amd64`.  So when the docker container temporary directory structure is created, it ALREADY has the binary.  In our case, this is a problem, because the binary is the SAME NAME as the directory.

```
   ⨯ release failed after 42.10s error=failed to link artifact:
    link /Users/jnovack/Source/govmomi/dist/vcsim_linux_amd64/vcsim dist/goreleaserdocker648450622/vcsim: file exists
```
image

So, I can't copy in the `vcsim/` directory structure, because the `binaries` section is building a binary called `vcsim`.

Any answer here seems to be a hack workaround (building a dummy binary, or renaming the binary name).  The only other solution is to use a [custom publisher](https://goreleaser.com/customization/publishers/), but again, that seems like a work-around.  The real issue is that there is no `clean: true` statement for a `dockers` instance.

jnovack:
This looks to be addressed in goreleaser/goreleaser#1890 and tagged as **wontfix**.

embano1:
Sorry for the trouble @jnovack. Looks like our idea does not easily work with `goreleaser`. But I think there's an easy workaround we could use to satisfy both use cases:

- Use a single `Dockerfile.goreleaser` for the `govc/vcsim` binaries and release as one Docker image (`vmware/vcsim`) with `goreleaser`
- Provide additional `Dockerfile`s for `vcsim` and `govc` in each directory so users can build their own images if needed

Having Github Actions enabled now on this repo, we could later automate the release process by using the `goreleaser` Github Action.

WDYT @dougm ?

dougm:
@embano1 I'm ok with whatever you guys think will work best here.  I've yet to dig into goreleaser at all and haven't been using docker for govc or vcsim myself.

jnovack:
> Use a single `Dockerfile.goreleaser` for the `govc`/`vcsim` binaries and release as one Docker image (`vmware/vcsim`) with `goreleaser`.

You don't have to put both binaries in the same file.  That's not a blocker.  We can have two images, each with one binary.  To me, one binary per image just feels more normal (see [my post above](https://github.com/vmware/govmomi/pull/2310#issuecomment-794725197)).  

I think I'm getting hung up on the need to have a separate `Dockerfile` for the user.  At best, it can be in a `deployments/` directory.  If `vmware/vcsim` and `vmware/govc` containers are published, I'm going to be using the published versions, not my own.  The only reason I would want to build it myself is if I'm making changes, and even then, I should be making them `goreleaser`/CI/CD compliant.

I counter propose with:

* Creating a `deployments/` directory ([golang-standards/project-layout compliant](https://github.com/golang-standards/project-layout)) with `Dockerfile.vcsim` and `Dockerfile.govc` for the people.
* Creating a `Dockerfile.goreleaser` for each binary `govc`/`vcsim`, and updating `goreleaser.yml` as appropriate for `linux/amd64` (just to stub, we can add more os/arch manifests later).

embano1:
> one binary per image just feels more normal

Ack-ed.

> I'm going to be using the published versions

Oh, sorry. Understood you actually wanted "full" `Dockerfile`s instead of using the provided images. Even better then!
So let's go ahead and create an updated `Dockerfile` per binary that continues to work with `goreleaser` and clean up the `goreleaser` configuration file while we're at it.

jnovack:
@embano1 Umm.. what happened here? This looks like a glitch in the matrix.

embano1:
> Umm.. what happened here? This looks like a glitch in the matrix.

Yeah, might be a temporary (flaky) glitch :) I'll keep an eye on it, but not related to your changes from what I see.

Dockerfile.govc
```diff
@@ -1,4 +1,43 @@
...
+COPY --from=build /etc/group /etc/group
+
+# Copy over the /tmp directory for golang/os.TmpDir
+COPY --chown=appuser --from=build /temporary-tmp-directory /tmp
+
+# Copy application from external build
+COPY govc /govc
+
+# Set entrypoint to application with container defaults
+ENTRYPOINT ["/govc"]
```
embano1, line 43: Should we use `CMD` or "enforce" `ENTRYPOINT`? (I know it can be overwritten, just asking from UX perspective)

.goreleaser.yml
```diff
@@ -12,6 +12,9 @@ builds:
     - 386
     - arm64
     - mips64le
+  ignore:
```
embano1, line 15: QQ: why do we need this?

.goreleaser.yml
```diff
@@ -98,23 +104,31 @@ brews:
...
 dockers:
-- image_templates: 
-  - "vmware/govc:{{ .Tag }}"
-  - "vmware/govc:v{{ .Major }}"
-  - "vmware/govc:v{{ .Major }}.{{ .Minor }}"
-  - "vmware/govc:latest"
-  goos: linux
-  goarch: amd64
-  dockerfile: Dockerfile.govc
-  binaries:
```
embano1, line 109: I think we should keep `id` (was binaries) to reference the specific binary used in that image (clarity).

Dockerfile.govc
jnovack: There's only one binary in the image. There's nothing else to do.

.goreleaser.yml
jnovack:
```
• DEPRECATED: skipped darwin/arm64 build on Go < 1.16 for compatibility, check https://goreleaser.com/deprecations/#builds-for-darwinarm64 for more info.
```

.goreleaser.yml
jnovack: You are right!  Fixed.

Which also corrects the first concern of `CMD` vs `ENTRYPOINT`.

.goreleaser.yml
```diff
@@ -132,3 +136,66 @@ dockers:
...
+    goarch: arm64
+    build_flag_templates:
+      - "--pull"
+      - "--label=org.opencontainers.image.created={{.Date}}"
+      - "--label=org.opencontainers.image.title={{.ProjectName}}"
+      - "--label=org.opencontainers.image.revision={{.FullCommit}}"
+      - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--label=org.opencontainers.image.url=https://github.com/vmware/govmomi/vcsim"
+      - "--platform=linux/arm64/v8"
+docker_manifests:
```
embano1, line 169: So we're going multi-arch images as well :)

.goreleaser.yml
```diff
@@ -132,3 +136,66 @@ dockers:
...
+  - name_template: vmware/govc:{{ .Tag }}
+    image_templates:
+    - vmware/govc:{{ .Tag }}
```
embano1, line 172: can we please also append `amd64` to (all of) these images for consistency and so it's clear in the manifest we're building for `amd` and `arm64`?

.goreleaser.yml
embano1: @dougm so you're aware: we're introducing multi-arch container builds which requires a specific Docker experimental flag. This is not an issue and "standard" in Github actions for goreleaser. But since we've not yet fully automated the release process (via actions), I wanted to bring this to your attention.

.goreleaser.yml
```diff
@@ -110,6 +110,8 @@ dockers:
       - "vmware/govc:v{{ .Major }}.{{ .Minor }}"
       - "vmware/govc:latest"
     dockerfile: Dockerfile.govc
+    binaries:
```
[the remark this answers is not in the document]
jnovack: dammit. 

I started drinking too early today.. or not soon enough.. 🤔

.goreleaser.yml
embano1: We don't need this as we're building with Go v1.16 in the upcoming action.

embano1:
Agree @jnovack. Let's break this up into smaller chunks.
I tried to build via `goreleaser` and it failed due to several issues with the `.goreleaser` file (not your fault).

May I suggest to only incorporate the Dockerfile changes and drop the rest? I'll create a follow up PR for the goreleaser stuff.

Thanks for your contribution and sorry for the back and forth due to the cleanup required...

jnovack:
@embano1 Wait, this last commit failed?  Your tone sounds like we're abandoning the PR. Maybe I'm misunderstanding.

This commit, `babe1a9`, is my submission to merge this PR.  This should be working as it should've whacked all the moles.

embano1:
Sorry, what I meant was that we don't include any changes to `.goreleaser` as it's merely broken (not by you) and needs a refactoring for the upcoming release automation changes. Spend some time yesterday in troubleshooting and fixing the `goreleaser` bits and it would be too much of a change for this PR.

Since you agreed that using Docker images built by this repo (instead of your own), I was suggesting to move ahead (only) with the changes in the corresponding `Dockerfile`s for vcsim/govc.

Your work will be the foundation for the upcoming release automation changes.

WDYT?

jnovack:
Deal.

LGTM?

Dockerfile.govc
embano1: The issue is with usability. If you directly want to run `govc about` without an interactive shell with that image you would have to use:

```bash
docker run -it --rm vmware/govc:v0.24 about
```

Instead of 

```bash
docker run -it --rm vmware/govc:v0.24 govc about
```

The latter IMHO is more user-friendly, especially when it comes to figuring out why passing `govc about` is actually not running that sub-command.

embano1 (approved):
Thank you!
LGTM

PR #2315, opened 2021-03-10
Added UpdateServiceMessage to Session Manager

ScottHolden:
Added an extra method to the Session Manager to expose the `UpdateServiceMessage` method.

It was already defined as part of the vim25 methods & types, just missing from the session manager wrapper.

Let me know if there is anything else I need to submit with this PR.


dougm (approved):
Thanks @ScottHolden !

Issue #2308, opened 2021-03-05
vslm ExtendDisk and InflateDisk only support the host-based tasks and not the vCenter tasks


protochron:
It looks like when #2067 went in, the vslm `ExtendDisk` and `InflateDisk` methods only implemented the calls to `HostExtendDisk_Task` and `HostInflateDisk_Task` so they do not support calling  `ExtendDisk_Task` and `InflateDisk_Task` like many of the other methods in  `vslm.LocalObjectManager`.

It seems like both calls are missing something resembling:

```go
if m.isVC {
	res, err := methods.ExtendDisk_Task(ctx, m.c, &req)
	if err != nil {
		return nil, err
	}

	return object.NewTask(m.c, res.Returnval), nil
}
```

I've confirmed that `methods.ExtendDisk_Task` works if you build up the task manually with a `*vim25.Client`, so it seems like there shouldn't be a reason not to add support for both types of task unless there's some subtlety in the API that I'm missing.

dougm:
Sounds right to me @protochron , do you want to submit a PR to add those?

cc @yashdesai93

dougm:
Fixed in #2321 (thanks @protochron)

PR #2321, opened 2021-03-19
Add support for calling vCenter for VLSM ExtendDisk and InflateDisk

protochron:
As outlined in #2308, it looks like when #2067 was committed it only
supported calling `ExtendDisk` and `InflateDisk` on ESXi hosts and not
against vCenter.

This fixes the problem by adding the same logic used by other VSLM
methods to switch between vCenter and ESXi tasks.


dougm (approved):
Looks good, thanks @protochron

Issue #2322, opened 2021-03-22
Events example broken


embano1:
Flag arguments are not accepted, inconsistent use of GOVC_URL vs GOVMOMI_URL env vars.

PR #2323, opened 2021-03-22
Fix events example

embano1:
Fixes err variable shadowed in example.go which caused the examples to
fail silently on some errors, e.g. TLS issues when not passing
`-insecure`.

Minor documentation improvements on the example and start stream from
-10min instead of -1h to reduce load on vCenter backend.

Closes: #2322 

### DIFF
--- a/examples/events/main.go
+++ b/examples/events/main.go
@@ -30,9 +30,11 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
-// example use: go run main.go -url $GOVC_URL -b 8h -f VmEvent
 func main() {
-	begin := flag.Duration("b", time.Hour, "Begin time") // default BeginTime is 1h ago
+	// example use against simulator: go run main.go -b 8h -f
+	// example use against vCenter with optional event filters:
+	// go run main.go -url $GOVMOMI_URL -insecure $GOVMOMI_INSECURE -b 8h -f VmEvent UserLoginSessionEvent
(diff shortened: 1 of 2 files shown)


examples/events/main.go
```diff
@@ -30,11 +30,14 @@ import (
...
-// example use: go run main.go -url $GOVC_URL -b 8h -f VmEvent
 func main() {
-	begin := flag.Duration("b", time.Hour, "Begin time") // default BeginTime is 1h ago
+	// example use against simulator: go run main.go -b 8h -f
+	// example use against vCenter with optional event filters:
+	// go run main.go -url $GOVMOMI_URL -insecure $GOVMOMI_INSECURE -b 8h -f VmEvent UserLoginSessionEvent
+	begin := flag.Duration("b", 10*time.Minute, "Begin time") // default BeginTime is 10min ago
 	end := flag.Duration("e", 0, "End time")
 	follow := flag.Bool("f", false, "Follow event stream")
+	flag.Parse()
```
dougm, line 40: This is called already within `examples.Run`: https://github.com/vmware/govmomi/blob/c1c1989ea3ece09ecd01e0c327542146abebdb2b/examples/examples.go#L133

embano1: Thx, but only with this change the flags actually worked for me...hm...

embano1: Update: found the issue which is caused by an unhandled error in `example.go`. Will update PR to fix this.

dougm (approved):
Thanks @embano1

PR #2327, opened 2021-03-24
Add the vSAN stretched cluster reference.

jpeach:
Add a well-known ManagedObjectReference constant to allow access
to the vSAN stretched cluster management system.

Signed-off-by: James Peach <jpeach@vmware.com>


dougm (approved):
Thanks @jpeach

PR #2329, opened 2021-03-25
govc: remove device.boot -firmware default

dougm:
Avoid reverting from efi to bios when device.boot is used without specifiying the -firmware option


embano1 (approved):
LGTM

Issue #2235, opened 2021-01-10
vcsim has support for vsan endpoints


agllno:
Does the vcsim have support for vsan endpoints like VsanHelthPort, Pbmport and VsanPerfQuery?

If yes, then we can use replay capabilities to capture these data?

agllno:
Any update on this?

dougm:
vcsim does not have support for vsan endpoints currently, other than cns:
https://github.com/vmware/govmomi/blob/master/cns/simulator/simulator.go

Not sure if anyone's looked at implementing it yet, cc @gh05tn0va @divyenpatel @SandeepPissay

dougm:
We don't support those currently, and I don't any experience with them.  If you could provide more details such as API doc links, client usage examples (in any language), etc., that may help.

PR #2337, opened 2021-03-31
vcsim: add vsan simulator

dougm:
cns/simulator implements a subset of the vsan namespace.
Add vsan/simulator to implement other vsan related methods,
starting with the ConvertToStretchedCluster method.

Add cluster.stretch bats test.

Issue #2235

### DIFF
--- a/govc/test/cluster.bats
+++ b/govc/test/cluster.bats
@@ -269,3 +269,16 @@ _EOF_
   memory=$(govc cluster.usage -json DC0_C0 | jq -r .Memory.Summary.Usage)
   [ "$memory" = "34.3" ]
 }
+
+@test "cluster.stretch" {
+  vcsim_env -host 4
+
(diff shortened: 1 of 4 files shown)


dougm:
cc @jamiemonserrate

jpeach (approved):
I'm not that familiar with the simulator, but this overall looks pretty plausible to me.

When running on macOS, the bats tests have some incompatibilities (e.g BSD-derived xargs doesn't have a -r option), but the cluster.stretch test passed.

dougm:
Thanks for reviewing and trying it out @jpeach
Yeah, some of the bats tests depend on gnu, current workaround is to install: https://github.com/vmware/govmomi/blob/master/govc/test/README.md#darwin-macosx
Then we just prefer the gnu version: https://github.com/vmware/govmomi/blob/798dbec558a3da82b2e7f759d0f394608b954782/govc/test/test_helper.bash#L21-L24

PR #2343, opened 2021-04-01
govc: add library.update command

### LINKED ISSUE #2149: rename library objects
I don't know if I'm missing something, but is there a way to rename templates in a content library with govc?   rename.object cannot find something that can be listed via library.ls

dougm:
Fixes #2149

### DIFF
--- a/govc/library/update.go
+++ b/govc/library/update.go
@@ -1,0 +1,77 @@
+/*
+Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package library
+
+import (
+	"context"
+	"flag"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/vapi/library"
+)
+
+type update struct {
+	*flags.ClientFlag
+
+	library.Library
+}
+
+func init() {
+	cli.Register("library.update", &update{})
+}
+
+func (cmd *update) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
+	cmd.ClientFlag.Register(ctx, f)
+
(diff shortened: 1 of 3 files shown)


govc/library/update.go
```diff
@@ -0,0 +1,77 @@
...
+
+func init() {
+	cli.Register("library.update", &update{})
+}
+
+func (cmd *update) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
+	cmd.ClientFlag.Register(ctx, f)
+
+	f.StringVar(&cmd.Name, "n", "", "Library name")
```
embano1, line 42:
```suggestion
	f.StringVar(&cmd.Library.Name, "n", "", "Library name")
```

embano1: nit: consistency...instead of my suggestion, feel free to use **one** of both, embedded or named field.

govc/library/update.go
```diff
@@ -0,0 +1,77 @@
...
+		return err
+	}
+
+	m := library.NewManager(c)
+	dst, err := flags.ContentLibrary(ctx, c, f.Arg(0))
+	if err != nil {
+		return err
+	}
+
+	cmd.ID = dst.ID
```
embano1, line 73:
```suggestion
	cmd.ID = dst.Library.ID
```

govc/test/library.bats
```diff
@@ -50,7 +50,14 @@ load test_helper
...
 
-  run govc library.rm /my-content
+  run govc library.update -n new-content my-content
+  assert_success
+
+  run govc library.info new-content
+  assert_success
+  assert_matches "$id"
+
+  run govc library.rm /new-content
```
embano1, line 60: interesting: `rm` needs `/` whereas the other commands don't?

govc/library/update.go
embano1: Question: are we good with these two parameters for now or do we need to support more field update operations?

vapi/library/library.go
```diff
@@ -212,6 +212,20 @@ func (c *Manager) PublishLibrary(ctx context.Context, library *Library, subscrip
 	return c.Do(ctx, url.Request(http.MethodPost, spec), nil)
 }
 
+// UpdateLibrary can update one or both of the tag Description and Name fields.
+func (c *Manager) UpdateLibrary(ctx context.Context, l *Library) error {
+	spec := struct {
```
embano1, line 217: This is not going to work with the most recent VAPI version. Note the breaking change from the previous version you use: https://developer.vmware.com/docs/vsphere-automation/latest/content/api/content/library/library_id/patch/

govc/test/library.bats
dougm: The library.* commands all use the same library/finder package, the first `/` is optional in all cases.

govc/library/update.go
dougm: I've had this patch sitting around for a while, but iirc, the API (and UI) didn't support updating the other fields.

embano1 (approved):
LGTM

PR #2350, opened 2021-04-05
simulator: Wait until VM creation completes before adding to folder

theaugy:
If something like finder is used to retrieve the list of VMs in a
folder, it is possible that the list may include VMs that are not yet
fully constructed.

In this case, property.Wait() might observe theoretically-impossible
values, like <nil> for a VM's powerState.

This can cause TestRace (race_test.go) to hang.

The solution is to wait until the VM is fully initialized before adding
it to the folder. Because some initialization code expects the VM to
have a real MO reference, we have to use Registry.reference(vm) to
assign the new VM a reference (this is normally called by the
add-to-folder code; so thankfully it is idempotent).


theaugy:
Sorry, looks like the commit message got messed up when "interpreted" by the github formatter. `<nil>` was removed from this sentence:
```
In this case, property.Wait() might observe theoretically-impossible
values, like <nil> for a VM's powerState.
```

dougm (approved):
lgtm, thanks @theaugy

PR #2351, opened 2021-04-06
simulator.Context locking, Async Tasks, and Task Delay

theaugy:
This PR is broken into 4 commits:

- Tie simulator locking to the Context
- Make simulator tasks asynchronous
- Add support for task delay
- Ensure "hand off" from handler func to async task goroutine

It may be easiest to review each commit individually!

# Tie simulator locking to the Context

When the simulator dispatches a call, it locks the handler object, calls the handler function, then unlocks.
This is fine when the handler function doesn't need to lock anything else, and so long as the handler function does all of its work before completing. However, there are paths (like VM creation) where the handler function *does* lock other objects, and asynchronous tasks are closer to how a real vCenter would behave.

This commit replaces the prior lock elision policy of "if I'm locking the handler object, skip the lock" (see `Context.WithLock()`) with a scheme where all locks are associated with the request's `simulator.Context`. If a path tries to acquire a lock that the Context already holds, a reference to the lock is tracked, but the `.Lock()` is elided. This effectively means that locks are "shared" for a particular Context.

# Make simulator tasks asynchronous

Async tasks are closer to real vCenter behavior, and they are necessary to enable task-level delay (ie, simulating a long-running task).  This change essentially runs the existing `task.Run()` code in a goroutine. It also requires Context arguments for any handler functions that spawn a task, because all tasks must know which Context they're running in (to enable locking; see previous commit).

There were many places where synchronous tasks were implicitly assumed. So these have been fixed.

The simulator unit tests were run tens of thousands of times to shake out race conditions. Some of those have been fixed in other PRs. There remain racy tests that were racy before this PR, including

- TestRaceDestroy is just an inherently racy test, and fails sometimes.
- TestRace deadlocks every ~20k runs. My hunch is that this is a side effect of the inter-request (now inter-Context) lock ordering issues inherent to the simulator's overall approach to locking.
- TestSessionManagerLoginExtension
- ExampleHandlerREST: I have once observed a keepalive `handler.Stop()` and `handler.Start()` deadlock. This seems to be possible if the `Start()` goroutine fires while `Stop()` is waiting for it to finish:
  - `Stop()` holds the handler mutex while it waits on the waitGroup
  - `Start()` goroutine needs to acquire the handler mutex to re-enter `Start()` (`send()` ends up here)
  - The fact that the `Start()` goroutine ends up calling `Start()` feels like it may be part of the problem, but I haven't investigated fully.
- TestHandlerREST: Discussed in PR comments. Scheduling/timing anomaly, perhaps?

# Add support for task delay

Adds 2 tests: one to test the prior "API level" delay (which simulates a slow-to-respond vCenter), and one to test the new "task level" delay (which simulates a task taking a long time).

A `TaskDelay` var allows customization of the task delay; it can be configured globally or per-method, much like API level delays.

# Ensure "hand off" from handler func to async task goroutine

This commit guarantees that a lock held by the handler func (ie, the lock on the handler itself) is held by the async goroutine without the possibility of someone else getting the lock in between.

Without this behavior, there could be confusing Time-of-check vs. Time-of-use bugs, where the handler func checks some lock-protected data, spawns a task based on that data, but by the time the task runs, that data has changed.

### DIFF
--- a/cns/simulator/simulator.go
+++ b/cns/simulator/simulator.go
@@ -62,7 +62,7 @@ type CnsVolumeManager struct {
 
 const simulatorDiskUUID = "6000c298595bf4575739e9105b2c0c2d"
 
-func (m *CnsVolumeManager) CnsCreateVolume(ctx context.Context, req *cnstypes.CnsCreateVolume) soap.HasFault {
+func (m *CnsVolumeManager) CnsCreateVolume(ctx *simulator.Context, req *cnstypes.CnsCreateVolume) soap.HasFault {
 	task := simulator.CreateTask(m, "CnsCreateVolume", func(*simulator.Task) (vim25types.AnyType, vim25types.BaseMethodFault) {
 		if len(req.CreateSpecs) == 0 {
 			return nil, &vim25types.InvalidArgument{InvalidProperty: "CnsVolumeCreateSpec"}
@@ -160,7 +160,7 @@ func (m *CnsVolumeManager) CnsCreateVolume(ctx context.Context, req *cnstypes.Cn
 
 	return &methods.CnsCreateVolumeBody{
 		Res: &cnstypes.CnsCreateVolumeResponse{
-			Returnval: task.Run(),
+			Returnval: task.Run(ctx),
 		},
 	}
 }
@@ -237,7 +237,7 @@ func (m *CnsVolumeManager) CnsQueryAllVolume(ctx context.Context, req *cnstypes.
 	}
 }
 
-func (m *CnsVolumeManager) CnsDeleteVolume(ctx context.Context, req *cnstypes.CnsDeleteVolume) soap.HasFault {
+func (m *CnsVolumeManager) CnsDeleteVolume(ctx *simulator.Context, req *cnstypes.CnsDeleteVolume) soap.HasFault {
 	task := simulator.CreateTask(m, "CnsDeleteVolume", func(*simulator.Task) (vim25types.AnyType, vim25types.BaseMethodFault) {
 		operationResult := []cnstypes.BaseCnsVolumeOperationResult{}
 		for _, volumeId := range req.VolumeIds {
@@ -259,13 +259,13 @@ func (m *CnsVolumeManager) CnsDeleteVolume(ctx context.Context, req *cnstypes.Cn
 
 	return &methods.CnsDeleteVolumeBody{
 		Res: &cnstypes.CnsDeleteVolumeResponse{
-			Returnval: task.Run(),
+			Returnval: task.Run(ctx),
 		},
 	}
 }
 
 // CnsUpdateVolumeMetadata simulates UpdateVolumeMetadata call for simulated vc
-func (m *CnsVolumeManager) CnsUpdateVolumeMetadata(ctx context.Context, req *cnstypes.CnsUpdateVolumeMetadata) soap.HasFault {
+func (m *CnsVolumeManager) CnsUpdateVolumeMetadata(ctx *simulator.Context, req *cnstypes.CnsUpdateVolumeMetadata) soap.HasFault {
 	task := simulator.CreateTask(m, "CnsUpdateVolumeMetadata", func(*simulator.Task) (vim25types.AnyType, vim25types.BaseMethodFault) {
 		if len(req.UpdateSpecs) == 0 {
 			return nil, &vim25types.InvalidArgument{InvalidProperty: "CnsUpdateVolumeMetadataSpec"}
@@ -291,13 +291,13 @@ func (m *CnsVolumeManager) CnsUpdateVolumeMetadata(ctx context.Context, req *cns
 	})
 	return &methods.CnsUpdateVolumeBody{
 		Res: &cnstypes.CnsUpdateVolumeMetadataResponse{
-			Returnval: task.Run(),
+			Returnval: task.Run(ctx),
 		},
 	}
 }
 
 // CnsAttachVolume simulates AttachVolume call for simulated vc
-func (m *CnsVolumeManager) CnsAttachVolume(ctx context.Context, req *cnstypes.CnsAttachVolume) soap.HasFault {
+func (m *CnsVolumeManager) CnsAttachVolume(ctx *simulator.Context, req *cnstypes.CnsAttachVolume) soap.HasFault {
 	task := simulator.CreateTask(m, "CnsAttachVolume", func(task *simulator.Task) (vim25types.AnyType, vim25types.BaseMethodFault) {
 		if len(req.AttachSpecs) == 0 {
 			return nil, &vim25types.InvalidArgument{InvalidProperty: "CnsAttachVolumeSpec"}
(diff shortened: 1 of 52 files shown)


theaugy:
What's up is good to go, but I have one more tweak I'd like to make: I want to guarantee that the task "inherits" the lock, if it's held, from the goroutine/Context that calls `Run()`.
As written, there is no guarantee that the `Run()` goroutine will get the lock before `Service.call()`'s `WithLock()` call returns and releases, which could lead to reasonable-looking handler code experiencing ToCToU bugs.

I will be sure to check this PR before pushing; if reviews are underway, and if it's preferred, maybe I'll make that an extra commit instead of force-pushing again, or even put it in a separate PR.

theaugy:
last failure was a pre-existing data race: `searchDatastore.Run()` in host_datastore_browser.go modifies the Task while it's running, and that can create a data race with `task.Wait()` (because `WaitForUpdatesEx()` polls the task's state).

theaugy:
```
2021-04-07T02:48:08.6723681Z     handler_test.go:154: Expected i > j, got i: 9, j: 9
2021-04-07T02:48:08.6724490Z --- FAIL: TestHandlerREST (0.70s)
```
I ran this test x2000 to see if I could get a repro, but unfortunately, no joy.
Will re-push to rerun tests.

embano1:
Thank you @theaugy 

> Looking around for how to re-run the checks now that this is out of draft mode.

We don't use a bot but Github actions. Every push (incl. empty commits and force-push) to a PR automatically triggers the runs.

> I ran this test x2000 to see if I could get a repro, but unfortunately, no joy.

Sorry for that. Facing the same issues and tbh I currently have no idea why this is happening randomly. It might be due to underlying IO/network/scheduling glitches or ulimits which throw off some tests and are hard to reproduce. But never saw it with other repos/Github Actions that often.

theaugy:
I appreciate the commiseration, @embano1 :)

Perhaps you would also recognize this last failure in the govc-test portion:

```
ok 91 library.deploy vmtx
ok 92 library.pubsub


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```
(full log at https://api.travis-ci.org/v3/job/766162998/log.txt)

I haven't seen this before, and to be honest I haven't been able to run the govc-tests locally. Perhaps I should open an issue about that to see if I can get it figured out, and maybe I'll be able to figure out what's going on.

> It might be due to underlying IO/network/scheduling glitches or ulimits which throw off some tests and are hard to reproduce

That seems plausible, but I haven't worked extensively with Actions so I can't offer much help.
I did investigate the possibility that a task, somewhere in the keepalive path, was being waited on, forcing a stall for 20ms while updates are collected (see simulator/property_collector.go), but my reading of TestHandlerREST led me to believe there are no tasks proper (`count.Send` is nonblocking except for the mutex).

Perhaps the issue is what you describe, and it is exacerbated by a freak mutex contention between `count.Send` and one of the calls to `i.Value()`. Maybe switching `count` to atomics might help? I obviously can't get it to happen enough to test it, though.

theaugy:
cc @dougm 
This is ready for a proper look. I'm not sure if I should be pinging anyone else; please let me know (or cc them) if so!

embano1:
> I haven't seen this before, and to be honest I haven't been able to run the govc-tests locally. Perhaps I should open an issue about that to see if I can get it figured out, and maybe I'll be able to figure out what's going on.

I've seen this too in my PR for the migration from Travis to GHA. I could not reproduce this locally but good to know you also see it on Travis bc I thought it's a fundamental issue with GHA. On OSX you'll run into xarg issues running the `make govc-test` target. I usually run these in a Linux container (requires make, go, curl, jq IIRC).

I will have to defer to @dougm review your changes and verify the govc tests pass.

theaugy:
>  I usually run these in a Linux container

That's how I've been running them. I'm on linux, but the `-i` install of govc is a no-go on my system, so I run in a container that is essentially `ubuntu:latest` + golang + make + gcc + git, and:
```
RUN echo 'export PATH=$PATH:/usr/local/go/bin' > /root/.profile
CMD . /root/.profile && cd /src && make test
```
I then run like:
```
govmomi $ docker run -it --volume=$(pwd):/src $image
```
It hangs at:
```
(cd govc/test && ./vendor/github.com/sstephenson/bats/libexec/bats -t .)
1..222
```
`ps auxf` shows 2 leaf processes: `cat` and `cat /tmp/govc-test-`. I am guessing that `/tmp/govc-test-` is a log (it is a FIFO according to `ls -l`), and cat is just to get output to stdout. I haven't had a chance to investigate beyond that, but maybe this tells you something?

dougm (approved):
I've been wanting to fix the locking story, rather than continue to bandaid..  This is just awesome, huge thank you @theaugy for tackling this!  Really appreciate how you outlined the details and broke into separate commits.
Your new lock elision policy and implementation looks great to me.  And I had totally punted on async Tasks, great to have this now and integrated with the Delay mechanism.  I'd also been wanting to pass `Context` to all methods so we could move away from the `Map` package var (internally at least), so your PR gets us closer to that too.

Regarding the govc bats test failure "No output has been received in the last 10m0s...", I've only seen that in Travis in the past and have never been able to reproduce locally.
If this is still happening with GH Actions, if there's a timeout hook where we could `killall -QUIT vcsim`, the stack traces may shed some light.

I also run the govc tests locally rather than in a container, but can try to look at that too.  We do use a fifo, just to get the random vcsim port: https://github.com/vmware/govmomi/blob/cd87fd467d905bbd9e94e59e3e98238db51f3785/govc/test/test_helper.bash#L55-L59

theaugy:
Woo! Glad you like it!

I have a couple thoughts on some of the topics you brought up:

> fix the locking story

The locking story is hopefully improved a bit, but there's still "lock ordering issues inherent to the simulator's overall approach to locking" that Would Be Nice(tm) to fully resolve. In most cases, the "courtesy" locking around `Call()` is appropriate and sufficient, but handlers that have to lock something else might deadlock with other lock-something-else handlers. If we want to enforce a particular lock ordering (inventory hierarchy parent-to-child, or child-to-parent, for instance), the courtesy locking needs to either go away, be configurable, or be optional, because in principle only the handler knows what locks it will need to take. My first guess is that opt-out + $lock_order_policy is the right balance between simplicity (majority of handlers don't need to change) and correctness (handler has full knowledge of what it needs to do, and can ensure correct locks in correct order).

----

Another correctness problem is due to the fact that `Context.WithLock()` always uses `Map`, which could be problematic because not all MOs are owned by `Map`, so you lose custom sync.Locker behavior, or worse, have 2 mutexes for the same MO. I added comments about it (in sum: this is A Bad Thing but it works for now), but it would be much better to take the owning registry as an argument to `Context.WithLock()`, which should solve that issue. I shied away from that to keep this PR from getting even bigger ;)

Maybe combining `Registry.Get()` with `WithLock()` somehow might be the most ergonomic way to do it. Or we could embed the owning registry in the MO somehow (is this already in there somewhere?) and have `WithLock()` figure out the right Registry on its own.

----

Anyway, thanks for merging! Let me know if you'd like me to write up issues for any of these.

dougm:
Agreed, there are still issues to resolve. I had definitely opted for convenience/simplicity with the courtesy locking, but right it can get tricky when having to lock other objects.  Issues/PRs to further improve would be great, thanks again @theaugy !

PR #2353, opened 2021-04-08
export simulator.task.Wait

theaugy:
Handy to have for external code that works with simulator.Task

### DIFF
--- a/simulator/entity_test.go
+++ b/simulator/entity_test.go
@@ -60,7 +60,7 @@ func TestRename(t *testing.T) {
 		}).(*methods.Rename_TaskBody).Res.Returnval
 
 		task := Map.Get(id).(*Task)
-		task.wait()
+		task.Wait()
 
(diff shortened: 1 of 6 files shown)


theaugy:
First push had a data race between here:
```
func (f *PropertyFilter) DestroyPropertyFilter(ctx *Context, c *types.DestroyPropertyFilter) soap.HasFault {
        body := &methods.DestroyPropertyFilterBody{}

        RemoveReference(&f.pc.Filter, c.This) <-- f.pc.Filter
```
and here:
```
func (pc *PropertyCollector) apply(ctx *Context, update *types.UpdateSet) types.BaseMethodFault {
        for _, ref := range pc.Filter { <-- pc.Filter
```

I reckon we need a lock on `f.pc.Filter` for that `RemoveReference()` call.

dougm (approved):
Thanks @theaugy

PR #2358, opened 2021-04-13
vcsim: add moid value mapping mappings

dougm:
cluster moid values have "domain" prefix in real VC, not "clustercomputeresource"

folder moid values have "group" prefix.


embano1 (approved):
LGTM

PR #2359, opened 2021-04-14
Logging related updates

dougm:
* govc: log invalid NetworkMapping.Name with import.ova command
* vcsim: put verbose logging behind '-trace' flag
* govc: revert pretty print pruning optimization

### DIFF
--- a/govc/importx/ovf.go
+++ b/govc/importx/ovf.go
@@ -164,6 +164,21 @@ func (cmd *ovfx) Map(op []Property) (p []types.KeyValue) {
 	return
 }
 
+func (cmd *ovfx) validateNetwork(e *ovf.Envelope, net Network) {
+	var names []string
+
+	if e.Network != nil {
+		for _, n := range e.Network.Networks {
+			if n.Name == net.Name {
+				return
+			}
+			names = append(names, n.Name)
+		}
+	}
+
+	_, _ = cmd.Log(fmt.Sprintf("Warning: invalid NetworkMapping.Name=%q, valid names=%s\n", net.Name, names))
+}
+
 func (cmd *ovfx) NetworkMap(e *ovf.Envelope) ([]types.OvfNetworkMapping, error) {
 	ctx := context.TODO()
 	finder, err := cmd.DatastoreFlag.Finder()
@@ -176,6 +191,7 @@ func (cmd *ovfx) NetworkMap(e *ovf.Envelope) ([]types.OvfNetworkMapping, error) 
 		if m.Network == "" {
 			continue // Not set, let vSphere choose the default network
 		}
+		cmd.validateNetwork(e, m)
 
 		var ref types.ManagedObjectReference
 
--- a/simulator/simulator.go
+++ b/simulator/simulator.go
@@ -129,6 +129,12 @@ func Fault(msg string, fault types.BaseMethodFault) *soap.Fault {
 	return f
 }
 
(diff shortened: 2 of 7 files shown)


govc/importx/ovf.go
```diff
@@ -164,6 +164,21 @@ func (cmd *ovfx) Map(op []Property) (p []types.KeyValue) {
 	return
 }
 
+func (cmd *ovfx) validateNetwork(e *ovf.Envelope, net Network) {
+	var names []string
+
+	if e.Network != nil {
+		for _, n := range e.Network.Networks {
+			if n.Name == net.Name {
+				return
```
embano1, line 173: Is the following scenario possible?

- `e.Network.Networks` contains multiple elements, e.g. "net-1", "net-2"
- `net` is "net-2"
- the first validation (171) would fail and append to `names`
- the second validation (in loop 171) succeeds and the code returns

In this case we don't want to print the warning (179), correct? Only if there's no match found at all?

govc/importx/ovf.go
```diff
@@ -176,6 +191,7 @@ func (cmd *ovfx) NetworkMap(e *ovf.Envelope) ([]types.OvfNetworkMapping, error)
 		if m.Network == "" {
 			continue // Not set, let vSphere choose the default network
 		}
+		cmd.validateNetwork(e, m)
```
embano1, line 194: So we continue code execution and only print the warning then?

simulator/simulator.go
```diff
@@ -129,6 +129,12 @@ func Fault(msg string, fault types.BaseMethodFault) *soap.Fault {
 	return f
 }
 
+func logPrintf(format string, v ...interface{}) {
```
embano1, line 132: To be honest, when I saw the `logPrintf()` changes above I thought you missed the `.` in between the calls :) 

Suggestion: instead of using this function, why not use a custom io.Writer `Tracer` and pass it to `log.SetOutput` in the condition logic here: https://github.com/vmware/govmomi/blob/df08d4b21fec16b02d2d1b985d316c3762fc3b6c/vcsim/main.go#L104

This would ease extensibility and avoid "forgetting" to call the right log* statement.
Example: https://play.golang.org/p/Y94Z17wxm8V

govc/importx/ovf.go
dougm: Right, since it may not cause any errors, plus an imported vm's vNICs could be reconfigured post-import before being powered on.  We do similar here with warnings from the API: https://github.com/vmware/govmomi/blob/9af23072767a6d3093c000464d015120e6c6d20c/govc/importx/ovf.go#L279

simulator/simulator.go
dougm: Sounds good, I'll push that change.

govc/importx/ovf.go
embano1: You just wrote the unit test :D Thx for clarification

embano1 (approved):
LGTM

embano1:
Do you want to squash before merging? I can only squash/merge but not squash and create merge commit as we usually do.

dougm:
In general, I always squash changes that are made during the review.  Here I squashed the `logPrintf` -> `tracef` into the original commit.  Having multiple logical commits in a PR is fine, each of these could have been their own PR, I just batched into one with a "logging" theme.  Make it cleaner to revert, say for example if we wanted to remove just the ovf warning commit.

Issue #2293, opened 2021-02-24
The VM Cloning Process Has Some Unexpected Issues


serversci:
I'm able to clone a VM by the following means:

```
#!/bin/bash

main()
{
   export GOVC_INSECURE=true
   export GOVC_PASSWORD='password'
   export GOVC_URL='https://vcenter.company.com'
   export GOVC_USERNAME='username@domain'
   export GOVC_DATACENTER='datacenter'
   export GOVC_FOLDER='/datacenter/vm/folder/subfolder'
   export GOVC_DATASTORE_CLUSTER='datastorecluster'

   local SrcVM="$1"
   local DstVMName="$2"
   local DstVMIP="$3"
   local DstVMGateway="$4"
   local DstVMMask="$5"
   local DstVMTZ="$6"
   local DstVMDNSServers="$7"

   govc vm.clone -cluster=datacentercluster -vm="$SrcVM" -on=false "$DstVMName"

   local DstVMMAC
   DstVMMAC="$(
      govc vm.info -json "$DstVMName" |
      jq -r '.VirtualMachines[].Config.Hardware.Device[] | select(.MacAddress != null) | .MacAddress'
   )"

   local -a CustomizationOptions=(
      "-vm=$DstVMName"
      "-mac=$DstVMMAC"
      "-ip=$DstVMIP"
      "-netmask=$DstVMMask"
      "-DstVMGateway=$DstVMGateway"
      "-dns-server=$DstVMDNSServers"
   )

   govc vm.customize "${CustomizationOptions[@]}"
   govc vm.power -on "$DstVMName"
}

main "$@"; exit
```

Here are the outstanding issues:

1. The **/etc/resolv.conf** is being populated with `nameserver` entries only.

- How is **/etc/resolv.conf** populated with `search` entries?

2. The **/etc/hosts** has an entry with dots in the hostname removed. For example: `10.20.30.40 vm1mycompanycom` where I would have expected `10.20.30.40 vm1.mycompany.com`. Similarly, the **/etc/sysconfig/network** file has the line `HOSTNAME=vm1mycompanycom` where I would have expected `HOSTNAME=vm1.mycompany.com`.

- How do I prevent the dots from being removed from hostname entries as shown above?

3. The **/etc/hosts** file has an IPv6 line entered: `::1 vm1mycompanycom`. However, the VM has IPv6 disabled.

- How do I prevent IPv6 entries from being added to the **/etc/hosts** file?

[...]

PR #2363, opened 2021-04-15
govc: add vm.customize -dns-suffix flag

dougm:
Issue #2293

### DIFF
--- a/govc/test/vm.bats
+++ b/govc/test/vm.bats
@@ -1036,7 +1036,7 @@ load test_helper
   run govc vm.customize -vm DC0_H0_VM0 -type Linux
   assert_failure # no -ip specified
 
-  run govc vm.customize -vm DC0_H0_VM0 -ip 10.0.0.45 -netmask 255.255.0.0 -type Linux
+  run govc vm.customize -vm DC0_H0_VM0 -ip 10.0.0.45 -netmask 255.255.0.0 -type Linux -dns-server 1.1.1.1 -dns-suffix example.com
   assert_success
 
   run govc vm.power -on DC0_H0_VM0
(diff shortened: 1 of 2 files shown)


govc/test/vm.bats
```diff
@@ -1036,7 +1036,7 @@ load test_helper
   run govc vm.customize -vm DC0_H0_VM0 -type Linux
   assert_failure # no -ip specified
 
-  run govc vm.customize -vm DC0_H0_VM0 -ip 10.0.0.45 -netmask 255.255.0.0 -type Linux
+  run govc vm.customize -vm DC0_H0_VM0 -ip 10.0.0.45 -netmask 255.255.0.0 -type Linux -dns-server 1.1.1.1 -dns-suffix example.com
```
embano1, line 1039: Can you provide a test which show/validates multiple (ie list)?

govc/vm/customize.go
```diff
@@ -69,8 +70,10 @@ func (cmd *customize) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.gateway = nil
 	f.Var(&cmd.netmask, "netmask", "Netmask")
 	cmd.netmask = nil
-	f.Var(&cmd.dnsserver, "dns-server", "DNS server")
+	f.Var(&cmd.dnsserver, "dns-server", "DNS server list")
 	cmd.dnsserver = nil
+	f.Var(&cmd.dnssuffix, "dns-suffix", "DNS suffix list")
+	cmd.dnssuffix = nil
```
embano1, line 76: Is that needed?

dougm: Not normally, but @gressd has some code that embeds some govc commands (including this one), needs to clear the array after each call.

govc/test/vm.bats
dougm: In real vCenter these values would land inside the VM, files such as /etc/resolv.conf, /etc/hosts, etc.  The API (and hence vcsim) doesn't have access to them.

embano1 (approved):
LGTM

PR #2364, opened 2021-04-16
docs: update for 0.25 release

dougm:


govc/CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # changelog
 
+### 0.25.0 (2021-04-16)
```
embano1, line 3: the new version behavior (-l option) is missing

embano1 (changes requested):
Do we also generate a changelog for vcsim?

govc/CHANGELOG.md
dougm: I'll add that.  There's other stuff missing too.  Need a better way that going back through git history for future releases.
I didn't bother with the top-level govmomi/CHANGELOG.md this time and have never maintained one for vcsim.

embano1 (approved):
LGTM

PR #2375, opened 2021-04-20
toolbox: move process management to its own package

dougm:
This allows vcsim to use the process manager without depending on the backdoor RPC package.


embano1 (approved):
LGTM

PR #2382, opened 2021-04-21
Fix VM Guest test and vet warnings

embano1:
The test was actually broken due to incorrect use of waitgroup (one defer was missing). I removed all concurrency to simplify the test.

Closes: #2381

### DIFF
--- a/object/vm_guest_test.go
+++ b/object/vm_guest_test.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/simulator"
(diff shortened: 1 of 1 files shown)


object/vm_guest_test.go
```diff
@@ -48,25 +48,19 @@ func TestVirtualMachineWaitForIP(t *testing.T) {
 			t.Errorf("expected v6 ip, but %q is v4", ip)
 		}
 
-		var wg sync.WaitGroup
-		wg.Add(1)
-		go func() {
-			wg.Done()
-			ip, err = vm.WaitForIP(ctx, true)
```
dougm, line 55: We wanted to test:
- call to `vm.WaitForIP(ctx, true)` would block, waiting for an IPv4 address
- update `guest.ipAddress` would trigger a PropertyCollector update, causing WaitForIP to return

The first `wg.{Add,Done,Wait}` is so we don't Map.Update the ipAddress until the go routine is running.
The second waits for the go routine to return after WaitForIP returns.

How was the test breaking?  I think I can see how the first might race, where Map.Update could get called before the go routine calls WaitForIP (property.Wait underneath).  Maybe a good old fashioned sleep would be better?
```diff
modified   object/vm_guest_test.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/simulator"
@@ -51,16 +52,15 @@ func TestVirtualMachineWaitForIP(t *testing.T) {
 		var wg sync.WaitGroup
 		wg.Add(1)
 		go func() {
-			wg.Done()
+			defer wg.Done()
+
 			ip, err = vm.WaitForIP(ctx, true)
 			if err != nil {
 				t.Fatal(err)
 			}
-			wg.Done()
 		}()
-		wg.Wait()
+		time.Sleep(time.Millisecond * 100)
 
-		wg.Add(1)
 		simulator.Map.WithLock(simulator.SpoofContext(), obj.Reference(), func() {
 			simulator.Map.Update(obj, []types.PropertyChange{
 				{Name: "guest.ipAddress", Val: "10.0.0.1"},
```

embano1: I see, so the test was set up to simulate concurrency and blocking behavior. Got it. Yeah, you already saw the issue with the code (the missing defer caused a false positive, ie test always passed). The issue I faced was that my go vet (Go 1.16) gave me linter errors due to t.Fatal being called in the goroutine (which is incorrect by the spec). I'll revert the changes and use your proposed approach which fixes the concurrency issue. Just need to make the linter happy :)

embano1: I inverted the logic (to make linter happy) and simplified the test structure a bit, still adhering to your test case I hope.

dougm (approved):
lgtm

PR #2384, opened 2021-04-21
adding rancher to projects and reorganizing in alpha order

luthermonson:
simple readme update to add Rancher to the projects list as well as reorganize them all in alpha ordering


dougm (approved):
Thanks @luthermonson

PR #2397, opened 2021-04-28
Set RoundTripper in ssoadmin.NewClient

dougm:
Commit 4ed615f6 added RoundTripper to support govc '-verbose', with govc sso.* commands setting RoundTripper,
but NewClient needs to set the default.


embano1 (approved):
LGTM

PR #2414, opened 2021-05-05
Isolate the SSO bats test that times out

akutz:
This patch isolates the govc test for `sso.bats` -- the test that often times out when run as part of GitHub actions. This change is to determine whether or not the issue is specific to `sso.bats` or if the timeout occurs at that point during the 200+ test run for the uber-bats test. This patch also excludes the `sso.bats` test from the uber run.

To enable `sso.bats`, it must be run with `SSO_BATS=1`. The make target `govc-test-sso` can be used to run `sso.bats`.

### DIFF
--- a/.github/workflows/govmomi-govc-tests.yaml
+++ b/.github/workflows/govmomi-govc-tests.yaml
@@ -45,3 +45,51 @@ jobs:
 
       - name: Run govc Tests
         run: make govc-test
+
+  govc-tests-sso:
+    name: Run govc Tests - sso
+    strategy:
+      matrix:
+        go-version: ["1.15", "1.16"]
+        # tests randomly hang on ubuntu-20.04 go v1.16
+        platform: ["ubuntu-18.04"]
+
+    runs-on: ${{ matrix.platform }}
+    timeout-minutes: 20
+
+    steps:
+      - name: Set up Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+        id: go
+
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Run govc Tests - sso
+        run: make govc-test-sso
+
+  govc-tests-sso-assert-cert:
+    name: Run govc Tests - sso - assert cert
+    strategy:
+      matrix:
+        go-version: ["1.15", "1.16"]
+        # tests randomly hang on ubuntu-20.04 go v1.16
+        platform: ["ubuntu-18.04"]
+
+    runs-on: ${{ matrix.platform }}
+    timeout-minutes: 20
+
+    steps:
+      - name: Set up Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+        id: go
+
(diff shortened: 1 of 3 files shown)


akutz:
Hi @dougm / @embano1,

I do not think we need to merge this. We can just keep resending it to see if we can cause a timeout in either the original `govc-tests` or new `govc-tests-sso` action.

dougm:
I was looking at adding a signal handler to get vcsim's stacktrace on timeout and noticed this output from vcsim during the test:
```
2021/05/05 10:16:04 http: TLS handshake error from 127.0.0.1:50294: remote error: tls: bad certificate
```
This is triggered by `govc about.cert`.  But we also use `about.cert` in cli.bats, session.bats, tags.bats
Not sure if it's related, but would be curious if the test isn't hanging with this PR, would sso.bats hang if enabled again and just this section removed:
```diff
modified   govc/test/sso.bats
@@ -30,9 +30,9 @@ load test_helper
   run govc sso.service.ls -t sso:sts -U
   assert_success "$sts"
 
-  cert=$(govc about.cert -show | grep -v CERTIFICATE | tr -d '\n')
-  trust=$(govc sso.service.ls -json -t sso:sts | jq -r .[].ServiceEndpoints[].SslTrust[0])
-  assert_equal "$cert" "$trust"
+  # cert=$(govc about.cert -show | grep -v CERTIFICATE | tr -d '\n')
+  # trust=$(govc sso.service.ls -json -t sso:sts | jq -r .[].ServiceEndpoints[].SslTrust[0])
+  # assert_equal "$cert" "$trust"
 
   govc sso.service.ls -t cs.identity | grep com.vmware.cis | grep -v https:
   govc sso.service.ls -t cs.identity -l | grep https:
```

.github/workflows/govmomi-govc-tests.yaml
```diff
@@ -45,3 +45,51 @@ jobs:
 
       - name: Run govc Tests
         run: make govc-test
+
+  govc-tests-sso:
```
embano1, line 49: I think we can simplify this workflow file into one job using matrix tests, e.g:

⚠️ please verify as I'm typing from my head :)

```yaml
jobs:
  govc-tests:
    name: Run govc Tests
    strategy:
      matrix:
        go-version: ["1.15", "1.16"]
        platform: ["ubuntu-18.04"]
        cmd: ["govc-test", "govc-test-sso-assert-cert"]
        experimental: [false]
        include:
          - go-version: "1.15"
            platform: "ubuntu-18.04"
            cmd: "govc-test-sso"
            experimental: true
          - go-version: "1.16"
            platform: "ubuntu-18.04"
            cmd: "govc-test-sso"
            experimental: true

    runs-on: ${{ matrix.platform }}
    continue-on-error: ${{ matrix.experimental }}
    timeout-minutes: 10

    steps:
      - name: Set up Go ${{ matrix.go-version }}
        uses: actions/setup-go@v2
        with:
          go-version: ${{ matrix.go-version }}
        id: go

      - name: Check out code
        uses: actions/checkout@v2

      - name: Run ${{ matrix.cmd }}
        run: make ${{ matrix.cmd }}
```

embano1:
Closing this as discussed. See: #2426

PR #2415, opened 2021-05-05
Use "vcsim uuidgen" for bats tests

akutz:
This patch fixes an issue where some bats tests fail because the `uuidgen` command is not present. The `vcsim` binary includes `uuidgen` for this reason, and thus the two locations where `uuidgen` was used directly have been updated to call `vcsim uuidgen` instead.


embano1 (approved):
LGTM

dougm (approved):

PR #2417, opened 2021-05-05
Update govc test docs with act

akutz:
This patch updates the README for the govc tests to indicate how users can run the tests like they would be run by GitHub Actions with the `act` command.


embano1 (approved):
LGTM

dougm (approved):
Thanks @akutz

PR #2426, opened 2021-05-07
fix: Isolate SSO govc tests

embano1:
Isolate `govc` SSO-related tests as separate workflows, marked as experimental so they don't fail the run in case of cancellation.

Note: Only running the SSO-related tests with one go version (1.15) for now and timeout of 3m.
Kudos to @akutz for pinning the issue down to the SSO tests and making the relevant Makefile/tests changes.

Closes: #2425


embano1:
Ready for review @dougm. The checks now show in detail which govc related test failed/was cancelled, making it easier in PR reviews.

dougm (approved):

PR #2428, opened 2021-05-07
vcsim: untie datastore capacity from local fs

### LINKED ISSUE #2390: vcsim should provide a way to set the initial capacity for a datastore
When running `vcsim`, I've noticed that when you're looking at cluster-level usage, the amount of free space can vary depending on where you're running `vcsim`! Doing some digging, it looks like `vcsim` makes a syscall to the host system to get the amount of available disk space and uses that as the amount of unused capacity for the underlying datastore. Normally this isn't an issue, but if you happen to run `vcsim` on a machine without a lot of free disk space then it limits what you can do with the simulator.

One way around the issue is to add more datastores, but that isn't really representative of the environment we're trying to simulate. It would be better if there was a way to set the amount of initial capacity in the datastore as either a parameter on the `vcsim` command or as an exposed parameter in `simulator.Model`.

dougm:
Using the current local filesystem to set datastore capacity is not useful and vcsim uses very little actual disk space.
The simulator.VirtualMachine impl already adjusts datastore FreeSpace when VirtualDisks, Snapshots, etc, are added and removed.
This change removes the os Statfs and sets capacity to 10TB by default.

The default disk size for VMs created by the Model has been increased from 1K to 10G

Fixes #2390


embano1:
Hey @dougm sorry for my delayed review. Will do thorough review early next week, but good work!

QQ: looks like the usage of this env var `VCSIM_DATASTORE_CAPACITY` is not documented (only in this PR). How about adding this to `vcsim` docs?

simulator/datastore.go
```diff
@@ -65,19 +66,15 @@ func parseDatastorePath(dsPath string) (*object.DatastorePath, types.BaseMethodF
...
 	}
 
 	info := ds.Info.GetDatastoreInfo()
 
-	now := time.Now()
-
-	info.Timestamp = &now
-	info.MaxMemoryFileSize = info.FreeSpace
-	info.MaxFileSize = info.FreeSpace
+	info.Timestamp = types.NewTime(time.Now())
```
embano1, line 77: For testing (perhaps in future) and deterministic behavior, should we make time mockable, e.g. via global timeFn or a clock (https://github.com/benbjohnson/clock)?

simulator/host_datastore_system.go
```diff
@@ -83,11 +84,28 @@ func (dss *HostDatastoreSystem) add(ctx *Context, ds *Datastore) *soap.Fault {
...
-	browser.Datastore = dss.Datastore
-	ds.Browser = Map.Put(browser).Reference()
+	if Map.Get(ds.Self) == nil {
+		browser := &HostDatastoreBrowser{}
+		browser.Datastore = dss.Datastore
+		ds.Browser = Map.Put(browser).Reference()
 
-	folderPutChild(ctx, folder, ds)
+		var capacity units.ByteSize
+		b := os.Getenv("VCSIM_DATASTORE_CAPACITY")
```
embano1, line 93: Can you please document `"VCSIM_DATASTORE_CAPACITY"` in the README?

simulator/host_datastore_system.go
```diff
@@ -83,11 +84,28 @@ func (dss *HostDatastoreSystem) add(ctx *Context, ds *Datastore) *soap.Fault {
...
+		if b == "" {
+			b = "1TB"
+		}
+		capacity.Set(b)
```
embano1, line 97: Ignoring error is fine here?

simulator/host_datastore_system.go
```diff
@@ -83,11 +84,28 @@ func (dss *HostDatastoreSystem) add(ctx *Context, ds *Datastore) *soap.Fault {
...
+
+		ds.Summary.Capacity = int64(capacity)
+		ds.Summary.FreeSpace = ds.Summary.Capacity
+
+		info := ds.Info.GetDatastoreInfo()
```
embano1, line 102: `info` is shadowed here, intentionally?

simulator/datastore.go
dougm: Not sure we need testing of the time value in this case, other than just `Timestamp != nil`

simulator/host_datastore_system.go
dougm: Not sure we should use an env var here at all.  Tempted to leave it hardcoded as `1TB` (give or take) and find a better way to configure the default if we find we need to.  Any thoughts @protochron ?

simulator/host_datastore_system.go
dougm: Ah yeah, we should just `panic()` on error here I think.

simulator/host_datastore_system.go
dougm: Whoops, no, I'll remove this line.

simulator/datastore.go
embano1: Fine with me.

simulator/host_datastore_system.go
protochron: I'm fine with a hardcoded value, though maybe use something like `10TB` and call it a day?

protochron:
Sorry for the delayed response on my end, I was on vacation for awhile.

I think the approach here looks sound. Being able to modify the default datastore size via an environment variable would be useful, but possibly not needed as long as the default size isn't too small.

dougm:
@protochron hope you had a good vacation!  Sounds good to me, 10TB it is.  We can address the use case(s) of being able to change that if/when it comes up.   cc @embano1

embano1 (approved):

embano1:
Thx @dougm, please just update the PR description to reflect the last changes.

dougm:
@embano1 good catch, updated commit message too

embano1 (approved):
LGTM

Issue #1403, opened 2019-03-08
vSAN support in govc CLI


cormachogan:
This is a new feature request.

I would like to see support added to govc for vSAN.  I would like govc CLI to be able to do the following:

For the whole vSAN datastore:
1. Display the number of hosts in the vSAN cluster
2. Display the number of vSAN components per host. 
3. Display the size of the vSAN datastore, as well as consumed space/remaining space.
4. Display vSAN health overview - green/yellow/red
5. Display detailed information on a specific health check
6. Display which services have been configured cluster wide (deduplication, compression, encryption, iSCSI service, Performance Service)

For individual vSAN objects:
1. Display policy of a vSAN object (VM, VMDK)
2. Display compliance state of a vSAN object against its policy
3. Display the components which make up an object, including their location and state
4. Display size/consumed space
5. Display location in datastore, including folder

Attached are some screenshots taken from a vSphere client which has vSAN enabled to show some of the information that can be seen in the UI. First there are some vSAN health details (summary/overview, followed by a detailed via of one health check)

![vsan-health-summary](https://user-images.githubusercontent.com/18258500/54024058-812ecc00-418e-11e9-833a-407a0ff11d38.png)
![vsan-health-detailed](https://user-images.githubusercontent.com/18258500/54024066-855ae980-418e-11e9-88c1-d41cd95a4675.png)

In this UI view, we can see which cluster wide services have been enabled/disabled:

![vsan services state](https://user-images.githubusercontent.com/18258500/54024112-b3d8c480-418e-11e9-8742-139684a36f8e.png)

Because all objects on vSAN are made up of components, and because a VM has multiple objects, it is very useful to see all objects, and all components of a VM. The UI provide a view which shows this, and include the state of the components as well as their location:

![vsan object](https://user-images.githubusercontent.com/18258500/54024261-2fd30c80-418f-11e9-85ce-f392f3e65ae9.png)

There are a lot of other things that we could include, but I think this makes a good starting point.

For reference material, there is a good **Cookbook for vSAN + PowerCLI** here:
https://storagehub.vmware.com/section-assets/powercli-cookbook-for-vsan

There are an API references for **vSAN health**:
https://code.vmware.com/apis/398/vsan#/doc/vim.cluster.VsanVcClusterHealthSystem.html

Some useful PowerCLI cmdlets:

[...]

PR #2432, opened 2021-05-10
govc: add vsan info and change commands

### LINKED ISSUE #1965: Implement VsanUnmapConfig in cluster.change
We would like the ability to set VsanUnmapConfig (see (https://bit.ly/3feVfOR & https://bit.ly/3b5jlZ8) in govc, at a cluster level (probably in cluster.change).

This setting is not exposed in the vSphere UI and currently the only way for a user to set this is either via RVC (see rvc.rvc-git/lib/rvc/modules/vsan.rb#8244) or using PowerCLI:

`Get-Cluster -name <cluster> | Set-VsanClusterConfiguration -GuestTrimUnmap $true`

It would be great if we could do this directly in govc. 

Note that we can set this host-by-host using:
`govc host.option.set --host.ip=<host IP> VSAN.GuestUnmap 1`

And then use a for-loop across the hosts to set:
`for host in $(govc find . -type h -datastore $(govc find -i datastore -name vsanDatastore) | cut -d'/' -f5);do echo $host; govc host.option.set --host.ip=$host VSAN.GuestUnmap 1;done`

But it would be preferable to have feature parity with PowerCLI (to be able to do this on a cluster-level with one command)

dougm:
- vsan.change command adds the ability to set VsanUnmapConfig

- vsan.info command adds the ability to inspect vSAN cluster config

- vcsim: add support for VsanClusterGetConfig and VsanClusterReconfig methods

Fixes #1965

Co-authored-by: Tanay Kothari <tkothari@vmware.com>
Co-authored-by: Doug MacEachern <dougm@vmware.com>

### DIFF
--- a/govc/main.go
+++ b/govc/main.go
@@ -98,6 +98,7 @@ import (
 	_ "github.com/vmware/govmomi/govc/vm/rdm"
 	_ "github.com/vmware/govmomi/govc/vm/snapshot"
 	_ "github.com/vmware/govmomi/govc/volume"
+	_ "github.com/vmware/govmomi/govc/vsan"
 )
 
 func main() {
--- a/govc/test/vsan.bats
+++ b/govc/test/vsan.bats
@@ -1,0 +1,24 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+@test "vsan.change" {
+  vcsim_env -cluster 2
+
+  run govc vsan.info -json DC0_C0
+  assert_success
+  config=$(jq .Clusters[].Info.UnmapConfig <<<"$output")
+  assert_equal null "$config"
+
+  run govc vsan.change DC0_C0
+  assert_failure # no flags specified
+
+  run govc vsan.change -unmap-enabled DC0_C0
+  assert_success
+
+  run govc vsan.info -json DC0_C0
+  assert_success
+
+  config=$(jq .Clusters[].Info.UnmapConfig.Enable <<<"$output")
+  assert_equal true "$config"
+}
--- a/govc/vsan/change.go
+++ b/govc/vsan/change.go
@@ -1,0 +1,99 @@
+/*
+Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vsan
+
+import (
+	"context"
+	"flag"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/vsan"
+	"github.com/vmware/govmomi/vsan/types"
+)
+
+type change struct {
+	*flags.DatacenterFlag
+
+	unmap *bool
+}
+
+func init() {
+	cli.Register("vsan.change", &change{})
+}
+
+func (cmd *change) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.DatacenterFlag, ctx = flags.NewDatacenterFlag(ctx)
+	cmd.DatacenterFlag.Register(ctx, f)
+
+	f.Var(flags.NewOptionalBool(&cmd.unmap), "unmap-enabled", "Enable Unmap")
+}
+
+func (cmd *change) Usage() string {
+	return "CLUSTER"
+}
+
+func (cmd *change) Description() string {
+	return `Change vSAN configuration.
+
+Examples:
+  govc vsan.change -unmap-enabled ClusterA # enable unmap
+  govc vsan.change -unmap-enabled=false ClusterA # disable unmap`
+}
+
+func (cmd *change) Run(ctx context.Context, f *flag.FlagSet) error {
+	vc, err := cmd.Client()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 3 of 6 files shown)


dougm:
@darkmesh-b  can you review?

darkmesh-b:
> @darkmesh-b can you review?

A few points:

* govc vsan.info returns the correct values of either enabled or disabled, when enabled via RVC
* govc vsan.change does **not** change the cluster unmap attribute (tested with vsan 7u2 build-17694817 and 7u2a build-17920168)
* The help file for govc vsan.change needs to be updated (shows the same as 'vsan.info')
* We should probably either change the name of the verb, or add a more specific value to show that we are changing the 'unmap' parameter (e.g. vsan.change.unmap or vsan.unmap) as vsan.change is ambiguous

dougm:
Thanks @darkmesh-b .  The various govc change (vm.change, cluster.change, etc) commands map to a reconfig method and command flags map to config spec fields.  I added doc for vsan.change with an example:
```
govc vsan.change -unmap-enabled ClusterA # enable unmap
```
Did you use the `-unmap-enable` flag?  Otherwise, `VsanClusterReconfig` won't make the change.

I've used rvc from GitHub before, but it has since been archived.  Would be interested to see rvc's vsan functionality, is there a version included with vCenter or somewhere else?

darkmesh-b:
Right, thanks for the clarification @dougm; both enabling and disabling now work fine:

` ./govc vsan.change -unmap-enabled=0
[14-05-21 15:04:11] Updating vSAN...OK

./govc vsan.info
Path:             /Datacenter/host/New Cluster
  Enabled:        true
  Unmap Enabled:  false

./govc vsan.change -unmap-enabled
[14-05-21 15:04:49] Updating vSAN...OK

./govc vsan.info
Path:             /Datacenter/host/New Cluster
  Enabled:        true
  Unmap Enabled:  true`

We should probably change the output behaviour when no arguments are given (to the help page?) - and as mentioned change the name to avoid ambiguity

RVC is indeed included within vCenter: once in the bash shell, `rvc localhost` gets you connected

dougm:
@darkmesh-b yes, we could add validation to the command so at least 1 flag is set.
I see there's quite a few rvc vsan commands, anything else you think would be useful here?  Some other ideas mentioned in #1403

darkmesh-b:
@dougm any/all of those [mentioned](https://github.com/vmware/govmomi/issues/1403) there would be great, I think Cormac has it covered pretty well. 

Cormac has a good [reference guide](https://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/products/vsan/vmware-ruby-vsphere-console-command-reference-for-virtual-san.pdf) to RVC (although a little old), which is useful. Some highlights below --

The rvc command `vsan.cluster_info ` cycles through each host, providing some useful information, including if vsan is enabled on the host, the UUIDs, disk sizes/use, fault domains, the vmk adapter used for vSAN and services enabled. This caters for points 1.1 & 1.6 on Cormac's list. There's quite a lot of info here:

```
Host: w1-pe-poc-esx-012.eng.vmware.com
  Product: VMware ESXi 7.0.2 build-17630552
  vSAN enabled: yes
  Cluster info:
    Cluster role: agent
    Cluster UUID: 5272f746-9a83-3b86-0f92-10951a6f816d
    Node UUID: 607ed6cd-f90d-5e38-f781-ac1f6b54a5f0
    Member UUIDs: ["607ed6bd-0f2c-fd02-8567-ac1f6b54a600", "607edb29-426c-cd6c-a5e0-ac1f6b54a370", "607ed6cd-f90d-5e38-f781-ac1f6b54a5f0", "607ed6bd-4c64-5008-7d9f-ac1f6b54a4c0"] (4)
  Node evacuated: no
  Storage info:
    Auto claim: no
    Disk Mappings:
      Cache Tier: Local NVMe Disk (t10.NVMe____INTEL_SSDPED1K375GA_____________________000133902E5CD2E4) - 349 GB, v14
      Capacity Tier: Local NVMe Disk (t10.NVMe____INTEL_SSDPE2KX020T7_____________________00002DA954E4D25C) - 1863 GB, v14
      Capacity Tier: Local NVMe Disk (t10.NVMe____INTEL_SSDPE2KX020T7_____________________000034C0CDE4D25C) - 1863 GB, v14
      Cache Tier: Local NVMe Disk (t10.NVMe____INTEL_SSDPED1K375GA_____________________000133901C5CD2E4) - 349 GB, v14
      Capacity Tier: Local NVMe Disk (t10.NVMe____INTEL_SSDPE2KX020T7_____________________0000326C27E4D25C) - 1863 GB, v14
      Capacity Tier: Local NVMe Disk (t10.NVMe____INTEL_SSDPE2KX020T7_____________________000034C12DE4D25C) - 1863 GB, v14
  FaultDomainInfo:
    Not configured
  NetworkInfo:
    Adapter: vmk2 (172.21.4.12)
  Data efficiency enabled: no
  Encryption enabled: no
```

The command `vsan.host_info` gives the same output for an individual host.

One of the most useful commands is `vsan.obj_status_report -t` which gives a very detailed list on each VM and object, showing the number of objects and healthy components, as well as any unassociated objects. This would cater for point 1.2 and 2.3

The `vsan.disks_stats` command shows the hosts/disks usage, thus covers point 2 with more detail as well as point 1.2 and 2.4

`vsan.health.health_summary` shows a health summary similar to the UI; covers points 1.4, albeit without colours. 

Simply performing an `ls` on the datastore location gives the size and usage of the vsanDatastore (covering point 1.3), e.g.:

```
/localhost/Datacenter/datastores> ls
0 datastore1: 12.35GB 12.2%
1 datastore1 (1): 12.35GB 12.2%
2 datastore1 (2): 12.35GB 12.2%
3 datastore1 (3): 12.35GB 17.6%
4 vsanDatastore: 32006.23GB 3.8%
```

dougm:
Thanks @darkmesh-b , we added validation to check that the flag is set.  We only have 1 flag currently, but will do the same as more are added.  And thank you for the vsan/rvc references, that'll be most helpful as we looking to expand govc/vsan functionality.

darkmesh-b (approved):

PR #2435, opened 2021-05-11
govc: Add support for getting the VC proxy and no-proxy configuration

matonev:
govc: Add support for getting the VC proxy and no-proxy configuration
Introduced the vAPI bindings and the implementation for those

govc output:
```
HTTP proxy:          Enabled
                     Server:    http://10.186.17.218
                     Port:      80
                     Username:  user
HTTPS proxy:         Disabled
FTP proxy:           Enabled
                     Server:  http://google.com
                     Port:    21
No Proxy addresses:  localhost, 127.0.0.1
```

### DIFF
--- a/vapi/appliance/networking/proxy.go
+++ b/vapi/appliance/networking/proxy.go
@@ -47,16 +47,16 @@ type Proxy struct {
 	Enabled  bool   `json:"enabled,omitempty"`
 }
 
-// ProxyConfig represents configuration for vcenter proxy.
-type ProxyConfig struct {
+// ProxyList represents configuration for vcenter proxy.
+type ProxyList struct {
 	Ftp   Proxy `json:"ftp,omitempty"`
 	Http  Proxy `json:"http,omitempty"`
 	Https Proxy `json:"https,omitempty"`
 }
 
-// Proxy returns all Proxy configuration.
-func (m *Manager) ProxyConfig(ctx context.Context) (*ProxyConfig, error) {
-	var res ProxyConfig
+// ProxyList returns all Proxy configuration.
+func (m *Manager) ProxyList(ctx context.Context) (*ProxyList, error) {
+	var res ProxyList
 	var rawRes []struct {
 		Key   string
 		Value Proxy
@@ -83,7 +83,7 @@ func (m *Manager) ProxyConfig(ctx context.Context) (*ProxyConfig, error) {
 }
 
 // NoProxy returns all excluded servers for proxying.
-func (m *Manager) NoProxyConfig(ctx context.Context) ([]string, error) {
+func (m *Manager) NoProxy(ctx context.Context) ([]string, error) {
 	r := m.Resource(applianceNoProxyConfigPath)
 	var res []string
 	return res, m.Do(ctx, r.Request(http.MethodGet), &res)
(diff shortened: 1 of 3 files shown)


dougm (commented):
Looks great, thanks @matonev !  Just a few minor comments/suggestions.

vapi/appliance/networking/proxy.go
```diff
@@ -0,0 +1,90 @@
...
+
+// ProxyConfig represents configuration for vcenter proxy.
+type ProxyConfig struct {
+	Ftp   ProtocolProxyConfig `json:"ftp,omitempty"`
+	Http  ProtocolProxyConfig `json:"http,omitempty"`
+	Https ProtocolProxyConfig `json:"https,omitempty"`
+}
+
+// Proxy returns all Proxy configuration.
+func (m *Manager) Proxy(ctx context.Context) (ProxyConfig, error) {
```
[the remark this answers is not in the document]
matonev: Thanks, renamed the `NoProxy` to `NoProxyConfig` as well

vapi/appliance/networking/proxy.go
```diff
@@ -0,0 +1,90 @@
...
+		case "ftp":
+			res.Ftp = c.Value
+		}
+	}
+
+	return &res, nil
+}
+
+// NoProxy returns all excluded servers for proxying.
+func (m *Manager) NoProxyConfig(ctx context.Context) ([]string, error) {
```
dougm, line 86: Let's leave this as you had it, `NoProxy`, since it matches the [API name](https://developer.vmware.com/docs/vsphere-automation/v7.0U2-deprecated/appliance/rest/appliance/networking/noproxy/get/) + return values.

vapi/appliance/networking/proxy.go
```diff
@@ -0,0 +1,90 @@
...
+type Proxy struct {
+	Server   string `json:"server,omitempty"`
+	Port     int    `json:"port,omitempty"`
+	Username string `json:"username,omitempty"`
+	Password string `json:"password,omitempty"`
+	Enabled  bool   `json:"enabled,omitempty"`
+}
+
+// ProxyConfig represents configuration for vcenter proxy.
+type ProxyConfig struct {
```
dougm, line 51: Wondering if we should name this `ProxyList` to match [the API](https://developer.vmware.com/docs/vsphere-automation/v7.0U2-deprecated/appliance/rest/appliance/networking/proxy/get/)
While this structure isn't a list itself, it does capture the data returned by the `ProxyList` API.

And same with the method below.  While the structure differs from what the API returns, I don't imagine new protocols being added to the list. And if they are, we can just add to this structure.

vapi/appliance/networking/proxy.go
```diff
@@ -0,0 +1,90 @@
...
+	Ftp   Proxy `json:"ftp,omitempty"`
+	Http  Proxy `json:"http,omitempty"`
+	Https Proxy `json:"https,omitempty"`
+}
+
+// Proxy returns all Proxy configuration.
```
dougm, line 57: Also make sure the doc comment matches the type/method name.

dougm (commented):
Thanks @matonev , just want to fixup the naming a bit more to align with the API naming.

vapi/appliance/networking/proxy.go
matonev: Agree. Renamed the `struct` and the method

dougm (approved):
Awesome, thanks @matonev !

Issue #2446, opened 2021-05-19
Executing file on windows vm using govc


rahul14sep:
I want to execute powershell script on window  vm using govc, I tried using `govc guest.run ` but it looks to execute file in /bin/bash shell, Is there any way i could tell govc to use windows command prompt?

Command I tried
govc guest.run -vm VC-ATM-ICLONE -l='username:password' 'powershell C:\windows_network_refresh.ps1'

and got below error
govc: ServerFaultCode: File /bin/bash was not found

Is there any other way I could execute file on windows vm.

dougm:
What does the following report for your vm?
```console
% govc object.collect vm/VC-ATM-ICLONE guest.guestFamily
```
The govc code uses cmd.exe when `windowsGuest`, otherwise bash: https://github.com/vmware/govmomi/blob/13851af3d26288ec5760970b5b3ae3fd298d014d/guest/toolbox/client.go#L120-L133

rahul14sep:
govc object.collect vm/VC-ATM-ICLONE guest
Output:
guest.toolsStatus                      types.VirtualMachineToolsStatus           toolsOk
guest.toolsVersionStatus               string                                    guestToolsCurrent
guest.toolsVersionStatus2              string                                    guestToolsSupportedNew
guest.toolsRunningStatus               string                                    guestToolsRunning
guest.toolsVersion                     string                                    11269
guest.toolsInstallType                 string                                    guestToolsTypeMSI
guest.guestId                          string                                    
guest.guestFamily                      string                                    otherGuestFamily
guest.guestFullName                    string                                    Windows Server 2019, 64-bit  (Build 17763)

Its not populating correct information for guest family because we are using ESX 6.7 which does not support win2k19 OS.

For such cases can we have optional parameter in guest.run to override shell path , where we could specify which 
 shell to use (shell="c:\\Windows\\System32\\cmd.exe" or shell="/bin/bash") something like that .

Like we have option in pyvmomi
ps = vim.vm.guest.ProcessManager.ProgramSpec(programPath=r'C:\Windows\System32\cmd.exe', arguments='/c powershell C:\windows_network_refresh.ps1')

dougm:
Ah ok. With #2451 you can use the same ProgramSpec in your example or the new guest.run `-w` flag to set guestFamily to Windows, which makes `cmd.exe /c` the default way to run the program

rahul14sep:
This will help. Thanks @dougm for adding this support.

[...]

PR #2451, opened 2021-05-20
govc: fix default guest.run path for unsupported Windows guests

dougm:
- Add '-w' flag to guest.run

Closes: #2446

### DIFF
--- a/govc/vm/guest/guest.go
+++ b/govc/vm/guest/guest.go
@@ -88,12 +88,18 @@ func (flag *GuestFlag) Toolbox() (*toolbox.Client, error) {
 
 	family := ""
 	var props mo.VirtualMachine
-	err = vm.Properties(context.Background(), vm.Reference(), []string{"guest.guestFamily"}, &props)
+	err = vm.Properties(context.Background(), vm.Reference(), []string{"guest.guestFamily", "guest.toolsInstallType"}, &props)
 	if err != nil {
 		return nil, err
 	}
 	if props.Guest != nil {
 		family = props.Guest.GuestFamily
+		if family == string(types.VirtualMachineGuestOsFamilyOtherGuestFamily) {
+			if props.Guest.ToolsInstallType == string(types.VirtualMachineToolsInstallTypeGuestToolsTypeMSI) {
+				// The case of Windows version not supported by the ESX version
+				family = string(types.VirtualMachineGuestOsFamilyWindowsGuest)
+			}
+		}
 	}
 
 	return &toolbox.Client{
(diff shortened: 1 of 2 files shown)


embano1:
Thx Doug, I'm fine with the code changes but was wondering why we don't allow to overwrite the SHELL but depend on the OS type instead? Wouldn't changing the SHELL be more flexible instead of introducing that new windows specific parameter?

dougm:
@embano1 we can override if an absolute path is given, for example:
```console
% govc guest.run -vm foo /usr/local/bin/any-program ...
```
The os family based `bash -c` and `cmd.exe /c` defaults when non-absolute path is given is just for convenience.
We could live without the new `-w` flag, in this case that would mean having to specify:
```console
% govc guest.run -vm foo c:\\Windows\\System32\\cmd.exe /c powershell c:\\bar.ps1
```
Instead of:
```console
% govc guest.run -vm foo -w powershell c:\\bar.ps1
```

embano1:
Thx for confirming my assumption that this is possible today, but adding `-w` is a convenience  feature. Do you foresee other combinations or more flags where this would be needed or is Windows the exception here?

dougm:
Yeah, windows is the exception.  Instead of adding `-w`, we could instead do the change below, toolsInstallType would only ever be `MSI` on Windows guests. What do you think?

```diff
modified   govc/vm/guest/guest.go
@@ -88,12 +88,18 @@ func (flag *GuestFlag) Toolbox() (*toolbox.Client, error) {
 
 	family := ""
 	var props mo.VirtualMachine
-	err = vm.Properties(context.Background(), vm.Reference(), []string{"guest.guestFamily"}, &props)
+	err = vm.Properties(context.Background(), vm.Reference(), []string{"guest.guestFamily", "guest.toolsInstallType"}, &props)
 	if err != nil {
 		return nil, err
 	}
 	if props.Guest != nil {
 		family = props.Guest.GuestFamily
+		if family == string(types.VirtualMachineGuestOsFamilyOtherGuestFamily) {
+			if props.Guest.ToolsInstallType == string(types.VirtualMachineToolsInstallTypeGuestToolsTypeMSI) {
+				// The case of Windows version not supported by the ESX version
+				family = string(types.VirtualMachineGuestOsFamilyWindowsGuest)
+			}
+		}
 	}
 
 	return &toolbox.Client{
```

embano1:
Nice, I like that much better! No (breaking) change for user and it will behave accordingly to what users would expect (see issue).

dougm:
ok, done @embano1 👍

embano1 (approved):
LGTM

PR #2468, opened 2021-05-27
vcsim: Sets "ChangeTrackingSupported" property of the simulated VM

amanpaha:
## Description

This change checks the device type and sets ChangeTrackingSupported property for the simulated VM accordingly.  
Before enabling change tracking for VM, we validates if change tracking is supported or not by checking the Capability.ChangeTrackingSupported property. While writing unit test we are using vcsim for getting simulated VM but we cant test our functionality after checking ChangeTrackingSupported, since it is always nil.

## How Has This Been Tested?

Ran virtual_machine_test.go (package simulator) to see if this property is populated or not.

### DIFF
--- a/simulator/virtual_machine.go
+++ b/simulator/virtual_machine.go
@@ -160,6 +160,8 @@ func NewVirtualMachine(ctx *Context, parent types.ManagedObjectReference, spec *
 	vm.Runtime.ConnectionState = types.VirtualMachineConnectionStateConnected
 	vm.Summary.Runtime = vm.Runtime
 
+	vm.Capability.ChangeTrackingSupported = types.NewBool(changeTrackingSupported(spec))
+
 	vm.Summary.QuickStats.GuestHeartbeatStatus = types.ManagedEntityStatusGray
 	vm.Summary.OverallStatus = types.ManagedEntityStatusGreen
 	vm.ConfigStatus = types.ManagedEntityStatusGreen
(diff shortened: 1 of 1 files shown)


embano1:
Thank you @amanpaha for your contribution! I‘ll have @dougm do the code review, but could you please squash and sign your commit? +1 for using the  `vcsim:` prefix in the commit title.

simulator/virtual_machine.go
```diff
@@ -2237,3 +2239,32 @@ func allSnapshotsInTree(tree []types.VirtualMachineSnapshotTree) []types.Managed
 
 	return result
 }
+
+func changeTrackingSupported(spec *types.VirtualMachineConfigSpec) *bool {
+	supported := false
+	for _, device := range spec.DeviceChange {
+		if dev, ok := device.GetVirtualDeviceConfigSpec().Device.(*types.VirtualDisk); ok {
+
+			if _, ok := dev.Backing.(*types.VirtualDiskFlatVer2BackingInfo); ok {
```
dougm, line 2248: Thanks @amanpaha , your first commit used a `switch dev.Backing.(type)`, which is what I would suggest here.  I think `changeTrackingSupported` can return `bool` instead of `*bool`, then each case can `return true` and the default case `return false`.  Then use the `NewBool` helper to convert to a pointer:
```go
vm.Capability.ChangeTrackingSupported = types.NewBool(changeTrackingSupported(spec))
```
What do you think?

amanpaha: Yes, definitely looks good, thanks for the suggestion.
I have made the change, should I resolve the conversation or reviewer will do it after verifying that the correct changes were made.

dougm (approved):
Looks good, thanks @amanpaha !

Issue #2495, opened 2021-06-16
[BUG] vcsim: The spec.location.host SOAP parameter is ignored when cloning


jpillon:
When using the SOAP API to clone a VM on a cluster with a targeted host, the target host is ignored and the new VM "runs" on the same host as the  cloned one. 

**To Reproduce**

1. Start the simulator
2. Perform a SOAP **CloneVM_Task** request on a cluster while defining the host parameter in the VirtualMachineRelocateSpec

**Expected behavior**
The VM should be "running" on the given host

**Affected version**
Found on v0.24.1, but the "missing" code persists on master

PR #2496, opened 2021-06-16
Rightly take the host parameter into account while cloning a VM on a cluster

jpillon:
## Description

Just set the Dest Host properly in the clone task

Closes: #2495

## Type of change

- [ X] Bug fix (non-breaking change which fixes an issue)
  not work as expected)

## How Has This Been Tested?

- [X ] Test with packer (after a bug fix in it)

## Checklist:

- [X] My code follows the CONTRIBUTION [guidelines](./../CONTRIBUTING.md) of
  this project

### DIFF
--- a/simulator/virtual_machine.go
+++ b/simulator/virtual_machine.go
@@ -1701,8 +1701,15 @@ func (vm *VirtualMachine) CloneVMTask(ctx *Context, req *types.CloneVM_Task) soa
 			pool = vm.ResourcePool
 		}
 	}
+
+	destHost := vm.Runtime.Host
+
+	if req.Spec.Location.Host != nil {
+		destHost = req.Spec.Location.Host
+	}
+
 	folder, _ := asFolderMO(Map.Get(req.Folder))
-	host := Map.Get(*vm.Runtime.Host).(*HostSystem)
+	host := Map.Get(*destHost).(*HostSystem)
 	event := vm.event()
 
 	ctx.postEvent(&types.VmBeingClonedEvent{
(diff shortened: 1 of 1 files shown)


simulator/virtual_machine.go
```diff
@@ -1701,8 +1701,16 @@ func (vm *VirtualMachine) CloneVMTask(ctx *Context, req *types.CloneVM_Task) soa
 			pool = vm.ResourcePool
 		}
 	}
+
+	DestHost := vm.Runtime.Host
```
dougm, line 1705: Non-exported variable names should start with lowercase, can you rename to `destHost` ?

simulator/virtual_machine.go
```diff
@@ -1701,8 +1701,16 @@ func (vm *VirtualMachine) CloneVMTask(ctx *Context, req *types.CloneVM_Task) soa
...
+
+	if req.Spec.Location.Host != nil {
+		hostRef := req.Spec.Location.Host.Reference()
```
dougm, line 1708: `req.Spec.Location.Host` is already a pointer, so there's no need for `hostRef` here, can do the following instead:
```go
	if req.Spec.Location.Host != nil {
		destHost = req.Spec.Location.Host
	}
```

dougm (commented):
Thanks @jpillion !  Looks good, 2 minor suggestions

dougm:
@jpillon one other request, can you prefix the commit message with `vcsim: ` ?  This helps the CHANGELOG generator:
https://github.com/vmware/govmomi/blob/master/CONTRIBUTING.md#format-of-the-commit-message

```
vcsim: take the host parameter into account while cloning a VM on a cluster
```

simulator/virtual_machine.go
jpillon: Commit amended ;-)

simulator/virtual_machine.go
jpillon: Sorry for that, I'm not extremely familiar with Go.
I amended my commit for that (since I also needed to change the commit message)

dougm (approved):
Thanks @jpillon

Issue #2314, opened 2021-03-09
Please add NumUplinkPorts= parameter to dvs.create 


haydonryan:
Hi There,

I noticed that when using PowerShell you are able to specify the number of uplink ports on a Distributed Virtual Switch.  Would it be possible to add that to govc as well, please?

Thanks in advance.

dougm:
In general, it is possible for govc to provide the same functionality as PowerCLI, since both use the same APIs.  But we as maintainers of govc don't know the entire API landscape, so any additional details/context you can provide would be most helpful.  Such as:
- example PowerCLI command including input/output
- links to any relevant PowerCLI docs for the command(s)

haydonryan:
I'm not surprised I tried to look through the vSphere API and instantly got lost on how big it is.  Thinking of possibly doing a PR myself but that could take a while, since I'm not super advanced with golang and it's been a while..

I was following Will Lamb's guide setting up Tanzu, but trying to convert it to govc:
`New-VDSwitch -Server $vc -Name $VDSName -Location (Get-Datacenter -Name $DatacenterName) -NumUplinkPorts 1`
https://github.com/lamw/vsphere-with-tanzu-homelab-scripts/blob/4396f568dd8d35966cc310b60dd1e5e56dd47e16/setup_vcsa.ps1#L35

dougm:
If you want to create a PR that'd be great, here's some info on how to start.  That PowerCLI snippet helps a lot.  I ran vcsim like so:
```console
% vcsim -l :8989 -trace-file vcsim.log
```
Created the DVS like so:
```console
PS /root> Connect-VIServer -Server 10.0.0.241 -User user -Password 'pass' -Port 8989
PS /root> New-VDSwitch -Name VDS -Location (Get-Datacenter -Name DC0) -NumUplinkPorts 1
```

From vcsim.log, `xmlformat` of the CreateDVS_Task request:
```xml
  <CreateDVS_Task xmlns="urn:vim25">
   <_this type="Folder">folder-6</_this>
   <spec>
    <configSpec xsi:type="VMwareDVSConfigSpec">
     <name>VDS</name>
     <uplinkPortPolicy xsi:type="DVSNameArrayUplinkPortPolicy">
      <uplinkPortName>dvUplink1</uplinkPortName>
     </uplinkPortPolicy>
     <contact/>
    </configSpec>
   </spec>
  </CreateDVS_Task>
```

Doing that in govc would be something like the following:
```diff
[...]

PR #2499, opened 2021-06-23
adding flag to control number of uplink ports when creating a new distributed switch

haydonryan:
## Description

When creating distributed and nested switches, you will often want to provide only a single uplink port, or in the case of a 4 port nic, utilize 4 ports.  This feature was missing from dvs.create.

I also added a local check to ensure that you don't try and create a DVS with 0 or invalid ports, as vSphere won't let you do that.

No dependencies

Closes: #2314

## Type of change

  not work as expected)

## How Has This Been Tested?
- I used the code that  @dougm provided in the issue, and put modified the code. I compared what I received in the simulator with the output Doug provided by running the Powershell command.

- I then powered up my homelab and tested this against VSphere 7.0 u2 and created a DVS with 1 uplink port, 3, 1030. 
Added a default of 2 ports, and added the CLI flags to override.  Tested again with no flag, 1,3,1030.

**Test Configuration**:
* Toolchain: go version go1.14.2 linux/amd64

## Checklist:

  this project
- [ ] I have added tests that prove my fix is effective or that my feature works  - I can't find where to add tests to this?
- [ ] New and existing unit tests pass locally with my changes - I can't find where to add tests to this? ran go test but didnt' see any tests (other than main) run.


govc/dvs/create.go
```diff
@@ -34,6 +34,8 @@ type create struct {
 	configSpec *types.VMwareDVSConfigSpec
 
 	dProtocol string
+
+	numUplinkPorts int32
```
dougm, line 38: Can use `int` here instead of `int32`, to avoid the later conversion?

govc/dvs/create.go
```diff
@@ -49,9 +51,12 @@ func (cmd *create) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.DVSCreateSpec.ConfigSpec = cmd.configSpec
 	cmd.DVSCreateSpec.ProductInfo = new(types.DistributedVirtualSwitchProductSpec)
 
+	cmd.numUplinkPorts = 2 // Set Default Uplink ports to 2
```
dougm, line 54: I think the default should be '0' to maintain the existing behavior, what do you think?

govc/dvs/create.go
```diff
@@ -49,9 +51,12 @@ func (cmd *create) Register(ctx context.Context, f *flag.FlagSet) {
...
+
 	f.StringVar(&cmd.ProductInfo.Version, "product-version", "", "DVS product version")
 	f.Var(flags.NewInt32(&cmd.configSpec.MaxMtu), "mtu", "DVS Max MTU")
 	f.StringVar(&cmd.dProtocol, "discovery-protocol", "", "Link Discovery Protocol")
+	f.Var(flags.NewInt32(&cmd.numUplinkPorts), "num-uplinks", "Number of Uplinks")
```
dougm, line 59: Looks like you can just use `f.IntVar(&cmd.numUplinkPorts, ...)` here.  We only need to use `f.Var` + `flags.NewX` for spec fields sent over the wire.  In this case `cmd.numUplinkPorts` is just a local counter.

govc/dvs/create.go
```diff
@@ -94,6 +99,18 @@ func (cmd *create) Run(ctx context.Context, f *flag.FlagSet) error {
...
+	numUplinkPorts := int(cmd.numUplinkPorts)
+
+	if numUplinkPorts > 0 {
+		var policy types.DVSNameArrayUplinkPortPolicy
+		for i := 0; i < numUplinkPorts; i++ {
+			policy.UplinkPortName = append(policy.UplinkPortName, fmt.Sprintf("Uplink %d", i+1))
+		}
+		cmd.configSpec.UplinkPortPolicy = &policy
+	} else {
+		return errors.New("Must specify more than one uplink port")
```
dougm, line 111: This isn't required to create a DVS though, is it?  Seems like we can keep the default as `0`, without erroring here.

dougm (commented):
Thanks @haydonryan !  No need to add attribution for my part of the patch, but there are some docs on how-to here: https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

One other request, if you could change the commit message to include the 'govc: ' prefix, that'll help the CHANGELOG generator: https://github.com/vmware/govmomi/blob/master/CONTRIBUTING.md#format-of-the-commit-message

See CHANGELOG.md for other examples:
`govc: add dvs.create '-num-uplinks' flag`

govc/dvs/create.go
haydonryan: My only concern around that is someone reading the code will see cmd.numUplinkPorts = 0 and assume that this creates zero ports when it actually skips the if block and creates 2 uplinks.

govc/dvs/create.go
haydonryan: Sounds Good, but how about we use uint since the number should never be negative?

govc/dvs/create.go
haydonryan: I was trying to have this cover if someone put in a negative number or hardcoded zero.  Originally I wasn't sure if there was an upper limit on the number of uplinks, but there doesn't appear to be.  Happy to remove if not appropriate.

govc/dvs/create.go
dougm: Sure, you can do that and use `flag.Uint()`

govc/dvs/create.go
dougm: I think it's ok to leave it as: `f.UintVar(&cmd.numUplinkPorts, "num-uplinks", 0, "Number of Uplinks")`
And let the `CreateDVS` method do what it does by default.

govc/dvs/create.go
dougm: Looks like if you use `f.UintVar(...)`, then the Go flag package will complain about a negative value, and `0` value will just result in vSphere's default behavior as the command does currently.

PR #2526, opened 2021-07-29
vcsim: use 'domain-c' prefix for cluster moids

dougm:
## Description

Real vCenter uses a `domain-c` prefix, rather than just `domain-`

## Type of change

  not work as expected)

## How Has This Been Tested?

Adjusted test in object.bats

## Checklist:

  this project

### DIFF
--- a/simulator/registry.go
+++ b/simulator/registry.go
@@ -45,8 +45,9 @@ var refValueMap = map[string]string{
 	"VirtualMachineSnapshot":         "snapshot",
 	"VmwareDistributedVirtualSwitch": "dvs",
 	"DistributedVirtualSwitch":       "dvs",
-	"ClusterComputeResource":         "domain",
+	"ClusterComputeResource":         "domain-c",
 	"Folder":                         "group",
+	"StoragePod":                     "group-p",
 }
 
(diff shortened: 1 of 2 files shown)


simulator/registry.go
```diff
@@ -114,7 +114,10 @@ func typeName(item mo.Reference) string {
 // valuePrefix returns the value name prefix of a given object
 func valuePrefix(typeName string) string {
 	if v, ok := refValueMap[typeName]; ok {
-		return v
```
zach96guan, line 117: Should we just return `v` here? As the prefix check targets "domain-c".

zach96guan (approved):
Thanks for the fix! Overall LGTM.

yuyin002 (dismissed):

simulator/registry.go
dougm: We only want to return `v` if it already contains a `-` (domain-c being the only case currently). Otherwise we want to append a `-`.

zach96guan: Thanks for the clarification! That makes sense to me.

HakanSunay (approved):
🚢

PR #2540, opened 2021-08-09
vcsim: avoid race when fetching object Locker

dougm:
## Description

The use of an interface method (mo.Reference) to get an object's lock can cause a read race.
The race detector doesn't follow what happens in the interface impl (most impls just return the `Self` field).
Prefer the mo 'Self' field via reflection to avoid this race.

## Type of change

  not work as expected)

## How Has This Been Tested?

Existing tests pass, in particular simulator/race_test.go after removing the special case for use of `*ListView.Self` in the `Registry.locker()` method.

## Checklist:

  this project


fdawg4l (approved):

zach96guan (approved):
Thanks for taking on the fix! lgtm.

Issue #2538, opened 2021-08-05
[BUG] Error processing attribute "type" with value when the vm name has a colon


gtmtnbiker:
**Describe the bug**
When running 
./govc fields.set  TemplateName foobar bsmithadm-LINUX-oeiac_ds6-2-2021-08-04T20:36:15Z

I get the output:
./govc: ServerFaultCode: 
Error processing attribute "type" with value "bsmithadm-LINUX-oeiac_ds6-2-2021-08-04T20"

But this works fine:
./govc fields.set -add -type VirtualMachine TemplateName foobar vm-rstjean

**To Reproduce**

1.  govc fields.add -type VirtualMachine TemplateName
1. Create a virtual machine named bsmithadm-LINUX-oeiac_ds6-2-2021-08-04T20:36:15Z
1. .govc fields.set  TemplateName foobar bsmithadm-LINUX-oeiac_ds6-2-2021-08-04T20:36:15Z

**Expected behavior**
It should not produce an error and set the field TemplateName to foobar for the vm  bsmithadm-LINUX-oeiac_ds6-2-2021-08-04T20:36:15Z

**Affected version**
v0.24.0, v0.26.0

**Screenshots/Debug Output**
```
centos76-clearcase-bsmith(bsmith): ./govc fields.set -verbose=true TemplateName foobar bsmithadm-LINUX-oeiac_ds6-2-2021-08-04T20:36:15Z
RetrieveProperties(propertyCollector, []types.PropertyFilterSpec)...
...CustomFieldsManager field: `govc object.collect -dump CustomFieldsManager:CustomFieldsManager field`

RetrieveProperties(propertyCollector, []types.PropertyFilterSpec)...
...group-d1 name: Datacenters

RetrieveProperties(propertyCollector, []types.PropertyFilterSpec)...
...datacenter-900399 name: Marlborough vRA
...datacenter-21     name: BedfordESX

SetField(CustomFieldsManager, bsmithadm-LINUX-oeiac_ds6-2-2021-08-04T20:36:15Z, 5307, foobar)...
&soap.Fault{
    XMLName: xml.Name{Space:"http://schemas.xmlsoap.org/soap/envelope/", Local:"Fault"},
    Code:    "ServerFaultCode",
[...]

PR #2556, opened 2021-08-30
govc: Fix incorrect MoRef parsing

embano1:
## Description

Fixes incorrect string arg parsing in MoRef handling logic inside `govc` where a non-MoRef argument (e.g. `vm:12:03:04`) was interpreted as MoRef making the `govc` action fail.

Closes: #2538

## Type of change

  not work as expected)

## How Has This Been Tested?

Existing tests pass with this internal change.

**Test Configuration**:
* Toolchain:
* SDK:
* (add more if needed)

## Checklist:

  this project


embano1:
@dougm your proposal seems to do the trick, thx!

I just have two questions based on your proposal:

1) It seems we don't have to touch the same method on `*PermissionsFlag` (see below), IMHO this code is still fine with that change
1) If the provided argument is not of type `MoRef`, `elements, err := finder.ManagedObjectList(ctx, arg)` is used which returns multiple MoRefs for the same object (e.g. ref from `host` folder and ref from `vm` folder) making the action take unnecessarily long. Assuming this is a known "issue" and could be optimized with deduplication logic?

`ManagedObjects()` on `*PermissionsFlag`

```go
func (f *PermissionFlag) ManagedObjects(ctx context.Context, args []string) ([]types.ManagedObjectReference, error) {
	if !f.asRef {
		return f.DatacenterFlag.ManagedObjects(ctx, args)
	}

	var refs []types.ManagedObjectReference

	for _, arg := range args {
		var ref types.ManagedObjectReference
		if ref.FromString(arg) {
			refs = append(refs, ref)
		} else {
			return nil, fmt.Errorf("invalid moref: %s", arg)
		}
	}

	return refs, nil
}
```

dougm:
Thanks @embano1 , for 1 ok to leave as-is for now, but I would also be fine with just removing the `-i` flag from those commands and just use `DatacenterFlag.ManagedObjects` directly in the permissions commands.

On 2,
> returns multiple MoRefs for the same object

Not sure what you mean?  It may return multiple MoRefs for the same input _arg_, such as `/dc1/vm/k8s-*`

Can we also add a test, something like below which currently fails w/o your PR:
```bash
  run govc vm.create "my:vm"
  assert_success

  run govc object.collect my:vm name
  assert_success my:vm
```

embano1:
> Not sure what you mean?

Given this command (with the patch included) `govc fields.set TemplateName foobar test-vm:03:12Z` I can see in the debugger that this line `elements, err := finder.ManagedObjectList(ctx, arg)` in `func (flag *DatacenterFlag) ManagedObjects` returns two elements for a single arg (`test-vm:03:12Z`): 

`Path = /vcqaDC/vm/test-vm:03:12Z` and `Path = /vcqaDC/host/cls/10.78.209.130/test-vm:03:12Z`.

dougm:
> `Path = /vcqaDC/vm/test-vm:03:12Z` and `Path = /vcqaDC/host/cls/10.78.209.130/test-vm:03:12Z`.

Oh, I was not aware of that, we should look at fixing it separately. Workaround would be to specify full path as the arg `/vcqaDC/vm/test-vm:03:12Z` or even relative to DC `vm/test-vm:03:12Z`

The finder methods for specific types (e.g. `VirtualMachineList`) limit the search to the type's folder already (vm,host,datastore,network)

embano1:
Tests added

dougm (approved):
Thanks @embano1

Issue #2570, opened 2021-09-02
[BUG] Data race in unprotected global variable `simulator.Map`


awsaba:
**Describe the bug**

A data race error is causing our CI testing that uses the `simulator` to fail about 1/10 of the time. The following error was able to be isolated using the govmomi/simulator tests in this repository with additional iterations.

The data race appears related to the global `simulator.Map` variable being reinitialized.

**To Reproduce**

1. Increase the test count and include `-race`. This command limits the testing to just `simulator` to isolate the issue.
```
GORACE=atexit_sleep_ms=10000 /usr/local/bin/go test -count 100 \
  -race \
  -timeout 10m \
  -v  \
  ./simulator/... | tee out.txt
```
1.  The a data race is reported in the output. Example is below.

Note: While I was able to reproduce this every time, a coworker was unable to locally. In their environment tests would time out consistently even with the timeout raised to 15m.

**Expected behavior**

No data races detected.

**Affected version**

This is the most recent version: v0.26.1-0.20210809200900-293f924ae9f4 and reproduced as above on the master branch.

**Screenshots/Debug Output**

```
 WARNING: DATA RACE
 Write at 0x000002b30538 by goroutine 148:
   github.com/vmware/govmomi/simulator.NewServiceInstance()
       /xxx/govmomi/simulator/service_instance.go:37 +0x20f
   github.com/vmware/govmomi/simulator.(*Model).Create()
       /xxx/govmomi/simulator/model.go:469 +0x17a
   github.com/vmware/govmomi/simulator.TestWaitForUpdatesOneUpdateCalculation()
       /xxx/govmomi/simulator/property_collector_test.go:470 +0x3bd
  testing.tRunner()
       /usr/local/Cellar/go/1.16.6/libexec/src/testing/testing.go:1193 +0x202

Previous read at 0x000002b30538 by goroutine 138:
   github.com/vmware/govmomi/simulator.(*Task).Run.func1()
       /xxx/govmomi/simulator/task.go:125 +0x507
```

**Additional context**

I have a patch to submit as a PR that addresses the race directly which may not completely address the issue for tests which are depending on the global values of `Map`, but the`govmomi/simulator` and our tests are passing.

PR #2571, opened 2021-09-02
vcsim: Fix Data race in unprotected simulator.Map

awsaba:
Close: #2750

## Description

There is data race when reinitialized the global `simulator.Map` during tests. This changes is to wrap access and re- initialization using singleton style functions with a corresponding mutex.

Closes: #2570

## Type of change

  not work as expected)

## How Has This Been Tested?

- [ ] `make check`: 2 tests FAIL also fail on `master`
- [ ] Testing the simulator package directly with `-race -count 100`: 

```
GORACE=atexit_sleep_ms=10000 /usr/local/bin/go test -count 100 \
  -race \
  -timeout 10m \
  -v  \
  ./simulator/... | tee out.txt
```

**Test Configuration**:

* Toolchain: go 1.17
* SDK:
* (add more if needed)

## Checklist:

  this project

### DIFF
--- a/cns/simulator/simulator.go
+++ b/cns/simulator/simulator.go
@@ -70,11 +70,13 @@ func (m *CnsVolumeManager) CnsCreateVolume(ctx *simulator.Context, req *cnstypes
 			return nil, &vim25types.InvalidArgument{InvalidProperty: "CnsVolumeCreateSpec"}
 		}
 
+		vimMap := simulator.Map()
+
 		operationResult := []cnstypes.BaseCnsVolumeOperationResult{}
 		for _, createSpec := range req.CreateSpecs {
 			staticProvisionedSpec, ok := interface{}(createSpec.BackingObjectDetails).(*cnstypes.CnsBlockBackingDetails)
 			if ok && staticProvisionedSpec.BackingDiskId != "" {
-				datastore := simulator.Map().Any("Datastore").(*simulator.Datastore)
+				datastore := vimMap.Any("Datastore").(*simulator.Datastore)
 				volumes, ok := m.volumes[datastore.Self]
 				if !ok {
 					volumes = make(map[cnstypes.CnsVolumeId]*cnstypes.CnsVolume)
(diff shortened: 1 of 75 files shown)


awsaba:
cc @protochron

dougm:
Thank you @awsaba !  I'm behind on govmomi issues+PRs, but will get to fully reviewing this soon.  The other part of the hold up is me kicking myself for adding the package var (`Map`) to begin with.  The `simulator.Map` seemed like a convenience originally, but long term we'd like to get rid of the package var.  Will share some more thoughts on that soon.  cc @akutz

awsaba:
> What do others think?

@dougm @akutz Thanks for the insight, this discussion has really helped.

>  We would need to make the Context param required. There are also methods that currently include Context in their signature, but still use Map directly, instead of ctx.Map, but those would be simple to fix up.

Looking at the error from #2571 and the other common ones that we see in CI, that method at least is one of these that takes a `Context` but doesn't use `ctx.map`. Smaller changes to change that and some locking in our fixtures around repeated calls to `Simulator.Create` may be enough to reduce how often it hits, so I'll be happy to close this PR and start with some smaller changes around that as a path forward while some of the higher level planning happens.

dougm:
> > We would need to make the Context param required. There are also methods that currently include Context in their signature, but still use Map directly, instead of ctx.Map, but those would be simple to fix up.
> 
> Looking at the error from #2571 and the other common ones that we see in CI, that method at least is one of these that takes a `Context` but doesn't use `ctx.map`. Smaller changes to change that and some locking in our fixtures around repeated calls to `Simulator.Create` may be enough to reduce how often it hits, so I'll be happy to close this PR and start with some smaller changes around that as a path forward while some of the higher level planning happens.

Thanks @awsaba , I think it would be good to move forward with both if you're up for it:
- make the Context param required for method impls
- use ctx.Map instead of Map directly in those methods

Might be better to have those as 2 PRs, or more, up to you. I think we'll want to do this, regardless of if a separate "v2" comes to be, as we'll reuse much of "v1". Hope to see you in the community zoom when we get that setup!

awsaba:
Closing in favor of multiple PRs to use `ctx.Map` and require it where necessary and further discussion on bigger breaking changes.

PR #2590, opened 2021-09-15
vcsim: Add TenantManager support in simulator

nikhaild:
## Description

Add simple implementation of TenantManager in simulator to support
calls for marking, unmarking, retrieving service provider entities.

Adding support for TenantManager API (mark/unmark/retrieve "service provider" object)
in vcsim would help develop components that make use of TenantManager API functionality.

Closes: issue-2587

## Type of change

  not work as expected)

## How Has This Been Tested?
- [x] TestTenantManagerVPX
   - Added a new unit test that tests various API inputs for the TenantManager mark/unmark/retrieve APIs.
- [x] TestTenantManagerForOldClients
  - Added a new unit test to verify old vim client won't see TenantManager moref entry in ServiceContent but new vim clients would see it.
```
$ go test -v -run "TestTenantManager*" -failfast -v github.com/vmware/govmomi/simulator
=== RUN   TestTenantManagerForOldClients
--- PASS: TestTenantManagerForOldClients (0.38s)
=== RUN   TestTenantManagerVPX
--- PASS: TestTenantManagerVPX (0.38s)
PASS
ok  	github.com/vmware/govmomi/simulator	0.767s
```
- [x] Ran all `simulator` unit tests.
```
$ go test -failfast github.com/vmware/govmomi/simulator
ok  	github.com/vmware/govmomi/simulator	40.581s
```
- [x] Ran `govc-test` make target.
```
$ make govc-test
…
ok 172 vcsim powercli
…
ok 184 vcsim ovftool
…
ok 225 vm.customize
ok 226 volume.ls
ok 227 vsan.change

$ echo $?
0
```

**Test Configuration**:
* Toolchain: go version go1.17.1 linux/amd64
* SDK:
* (add more if needed)

## Checklist:

  this project

### DIFF
--- a/simulator/vpx/service_content.go
+++ b/simulator/vpx/service_content.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2017-2021 VMware, Inc. All Rights Reserved.
+Copyright (c) 2017 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -83,5 +83,4 @@ var ServiceContent = types.ServiceContent{
 	HealthUpdateManager:         &types.ManagedObjectReference{Type: "HealthUpdateManager", Value: "HealthUpdateManager"},
 	FailoverClusterConfigurator: &types.ManagedObjectReference{Type: "FailoverClusterConfigurator", Value: "FailoverClusterConfigurator"},
 	FailoverClusterManager:      &types.ManagedObjectReference{Type: "FailoverClusterManager", Value: "FailoverClusterManager"},
-	TenantManager:               &types.ManagedObjectReference{Type: "TenantTenantManager", Value: "TenantManager"},
 }
--- a/simulator/vpx/service_content.go
+++ b/simulator/vpx/service_content.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+Copyright (c) 2017-2021 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -83,4 +83,5 @@ var ServiceContent = types.ServiceContent{
 	HealthUpdateManager:         &types.ManagedObjectReference{Type: "HealthUpdateManager", Value: "HealthUpdateManager"},
 	FailoverClusterConfigurator: &types.ManagedObjectReference{Type: "FailoverClusterConfigurator", Value: "FailoverClusterConfigurator"},
 	FailoverClusterManager:      &types.ManagedObjectReference{Type: "FailoverClusterManager", Value: "FailoverClusterManager"},
+	TenantManager:               &types.ManagedObjectReference{Type: "TenantTenantManager", Value: "TenantManager"},
 }
--- a/object/tenant_manager.go
+++ b/object/tenant_manager.go
@@ -1,0 +1,78 @@
+/*
+Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package object
+
+import (
+	"context"
+
+	"github.com/vmware/govmomi/vim25"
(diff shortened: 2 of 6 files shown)


simulator/vpx/service_content.go
```diff
@@ -83,4 +83,5 @@ var ServiceContent = types.ServiceContent{
 	HealthUpdateManager:         &types.ManagedObjectReference{Type: "HealthUpdateManager", Value: "HealthUpdateManager"},
 	FailoverClusterConfigurator: &types.ManagedObjectReference{Type: "FailoverClusterConfigurator", Value: "FailoverClusterConfigurator"},
 	FailoverClusterManager:      &types.ManagedObjectReference{Type: "FailoverClusterManager", Value: "FailoverClusterManager"},
+	TenantManager:               &types.ManagedObjectReference{Type: "TenantTenantManager", Value: "TenantManager"},
```
dougm, line 86: Thanks @nikhail !  The change looks good overall, but one issue is having this field set for client versions < vim 6.9.1, which may error when calling `RetrieveServiceContent` since this field isn't included in older vmodl versions.  Not an issue for Go clients, but Java and python will error.  We've punted on model versioning so far, but may need to consider it here.  The untested patch below may do the trick (we can also add helpers for future use).  Or a workaround would be to leave the field as nil by default in vcsim, then on the client side either set the field in tests or handle the case in the client code when `TenantManager` is nil.  

```diff
modified   simulator/service_instance.go
@@ -17,6 +17,7 @@ limitations under the License.
 package simulator
 
 import (
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
@@ -69,10 +70,17 @@ func NewServiceInstance(ctx *Context, content types.ServiceContent, folder mo.Fo
 	return s
 }
 
-func (s *ServiceInstance) RetrieveServiceContent(*types.RetrieveServiceContent) soap.HasFault {
+func (s *ServiceInstance) RetrieveServiceContent(ctx *Context, _ *types.RetrieveServiceContent) soap.HasFault {
+	content := s.Content
+	action := ctx.req.Header.Get("SOAPAction")
+	version := strings.Split(action, "/")[1]
+	if version >= "6.9.1" {
+		content.TenantManager = &types.ManagedObjectReference{Type: "TenantTenantManager", Value: "TenantManager"}
+	}
+
 	return &methods.RetrieveServiceContentBody{
 		Res: &types.RetrieveServiceContentResponse{
-			Returnval: s.Content,
+			Returnval: content,
 		},
 	}
 }
```

nikhaild:
> Thanks @nikhail . Realizing that there are 2 ways to get ServiceContent, the method and the property collector. I think we should spend more time thinking about how to do the versioning in a general purpose way. For now can you revert the versioning related changes from this PR? We can merge your PR and address the versioning before the next release. Let's keep your test as part of the PR, just add a `t.Skip("needs vmodl versioning")` to the beginning of the test. Does this sound ok to you?
Thanks @dougm , makes sense, I've removed the client version based setting of TenantManager moref in service_content.go (also it was only setting it for `RetrieveServiceContent()` path, not for property collector path, since modifying simulator's service content state on an API call looked iffy, yeah, would need a better approach to handle "view of service content" based on client version, on all paths, not just specific API path.

> I'd also rather avoid pulling in the go-version dep, we have some code for that already, just needs to be refactored into a package for use outside of govc: https://github.com/vmware/govmomi/blob/master/govc/flags/version.go#L32
Ah, that's better (than adding a full go module dep), thanks for pointing out that code, we can refactor that when dealing with the vmodl version handling later.

nikhaild:
Hmm.. the unit simulator tests failed here with this snippet:
https://github.com/vmware/govmomi/runs/4072561378?check_suite_focus=true
```=== RUN   TestSessionManagerLoginExtension
2120
2021/11/01 20:22:45 http: TLS handshake error from 127.0.0.1:40704: tls: first record does not look like a TLS handshake
2121
    session_manager_test.go:264: Post "https://sdkTunnel:8089/sdk?GOVMOMI_TUNNEL_PROXY_PORT=40563": net/http: TLS handshake timeout
2122
--- FAIL: TestSessionManagerLoginExtension (10.03s)
```
However I don't fully understand why, re-running that unit test locally runs fine for me:
```
$ pwd
/home/ndeshpande/go/src/github.com/vmware/govmomi/simulator
$ go test -v -run TestSessionManagerLoginExtension -failfast
=== RUN   TestSessionManagerLoginExtension
--- PASS: TestSessionManagerLoginExtension (0.01s)
PASS
ok  	github.com/vmware/govmomi/simulator	0.022s
```

dougm (approved):
Thanks @nikhail , lgtm!

dougm:
> Hmm.. the unit simulator tests failed here with this snippet: https://github.com/vmware/govmomi/runs/4072561378?check_suite_focus=true
> 
```
>     session_manager_test.go:264: Post "https://sdkTunnel:8089/sdk?GOVMOMI_TUNNEL_PROXY_PORT=40563": net/http: TLS handshake timeout
> 2122
> --- FAIL: TestSessionManagerLoginExtension (10.03s)
```
This test flakes, unrelated to your change, re-ran and it passed.

Issue #2488, opened 2021-06-08
extraconfig does not handle large content


briner:
**Describe the bug**

We are trying to use govc to load some heavy data. This works well through the html5 webui.  The content is 394'888 caracter long

**To Reproduce**

```bash
# generate a very long extraconfig
cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 394888  | head -n 1 > long_extraconfig
# try to use it
govc vm.change -vm soca-lab-govc-test -e test.extraconfig.with.long.extraconfig=$(long_extraconfig)
  # -bash: /home/beef/bin/govc: Argument list too long
```

As you can see, we got an error from the bash, as the extraconfig is way too long

**Expected behavior**
It could be a good idea to provide this lengthy content through a file instead of using the shell to do so
```bash
govc vm.change -vm soca-lab-govc-test -extraconfig-file test.extraconfig.with.long.extraconfig=./long_extravars
```
**Affected version**
```bash
govc version
  # govc 0.25.0
```

fortinj66:
Same issue loading large ignite files for Fedora CoreOS

dougm:
Good suggestion @briner , we can add a `-extraconfig-file` flag.  Just need to use a different `flag.Value`
 interface impl. 
 The current one is here: https://github.com/vmware/govmomi/blob/5dace8100c85cafd69ac7f4148549aaa6cf4ea76/govc/vm/change.go#L37-L44

dougm:
@tanayk2610 existing tests here: https://github.com/vmware/govmomi/blob/5dace8100c85cafd69ac7f4148549aaa6cf4ea76/govc/test/vm.bats#L87

Okeanos:
I also just ran into this problem. The root cause is the `ARG_MAX` definition (see [here](https://www.in-ulm.de/~mascheck/various/argmax/) for more details) which varies wildly between different operating systems. macOS for example has a generous `getconf ARG_MAX` value of `262144 bytes` whereas Windows stops at a lousy `32000 bytes`.

Being able to load from file would be a huge blessing and make this way more robust and reproducible across platforms.

RyanMillerC:
Also hitting this with large base64 encoded files that upload fine through HTML5 GUI but fail with `Argument list too long` due to `ARG_MAX` through govc.

Okeanos:
I don't want to sound annoying but could we get some information on when this change might be expected?

embano1:
[...]

PR #2596, opened 2021-09-26
govc: Add feature to read file contents for ExtraConfig

Okeanos:
## Description

The `vm.change` command allows setting extra values via `-e`, however, they have to be inline and are, hence, subject to the `getconf ARG_MAX` value of the local shell. To allow larger contents, e.g. for Fedora CoreOS Ignition configs, providing a file path and processing it internally is required.

Closes: #2488

## Type of change

  not work as expected)

## How Has This Been Tested?

I added some bats tests in `govc/test/vm.bats` similar to the existing extraConfig tests. Don't have a (local) ESX instance I can use to run the tests, though 🤷‍♂️ . Also I am not sure whether I can simply turn on the the GitHub Actions and have them run the tests for me. Would it be possible for you guys to add a pull request workflow that does run all tests?

## Checklist:

  this project

### DIFF
--- a/govc/vm/change.go
+++ b/govc/vm/change.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


Okeanos:
Sorry about the linting error; thought I had my editor set to auto format.

govc/vm/change.go
```diff
@@ -168,6 +198,10 @@ func (cmd *change) Run(ctx context.Context, f *flag.FlagSet) error {
 		cmd.VirtualMachineConfigSpec.ExtraConfig = cmd.extraConfig
 	}
 
+	if len(cmd.extraConfigFile) > 0 {
+		cmd.VirtualMachineConfigSpec.ExtraConfig = cmd.extraConfigFile
```
dougm, line 202: Looks like we can only use `-e` or `-f` with the same command invocation, as any `-f` would overwrite `-e` args here. How about the change below so both flags can be used?  I don't think original `len(cmd.extraConfig)` was needed.

```diff
modified   govc/vm/change.go
@@ -194,13 +194,7 @@ func (cmd *change) Run(ctx context.Context, f *flag.FlagSet) error {
 		return flag.ErrHelp
 	}
 
-	if len(cmd.extraConfig) > 0 {
-		cmd.VirtualMachineConfigSpec.ExtraConfig = cmd.extraConfig
-	}
-
-	if len(cmd.extraConfigFile) > 0 {
-		cmd.VirtualMachineConfigSpec.ExtraConfig = cmd.extraConfigFile
-	}
+	cmd.VirtualMachineConfigSpec.ExtraConfig = append(cmd.extraConfig, cmd.extraConfigFile...)
 
 	setAllocation(&cmd.CpuAllocation)
 	setAllocation(&cmd.MemoryAllocation)
```

dougm (commented):
Thanks @Okeanos ! Just one suggestion, lgtm otherwise

govc/vm/change.go
Okeanos: Applied your change :)

dougm (approved):
Great, thanks @Okeanos , we can merge when CI passes

Issue #2650, opened 2021-11-05
[BUG] Data race in simulator.EventManager.PostEvent


awsaba:
**Describe the bug**

**To Reproduce**

1. Run tests with higher count: `TEST_COUNT=10 make test`
1. The data race will occasionally be detected on different tests.

**Expected behavior**

No data race detected.

**Affected version**

This occurs on current master@14d8083c9ca296ce3c3e6d4fe21e455119f56db5, with go 1.16 and 1.17.

**Screenshots/Debug Output**

This run it occurred on TestDataCenterPowerOnMultiVMs, but other common tests include Test

```
=== RUN   TestDatacenterPowerOnMultiVMs
==================
WARNING: DATA RACE
Read at 0x00c0001c2d68 by goroutine 106:
  github.com/vmware/govmomi/simulator.(*EventManager).PostEvent()
      /Users/asaba/projects/govmomi/simulator/event_manager.go:159 +0x6b
  github.com/vmware/govmomi/simulator.(*Context).postEvent.func1()
      /Users/asaba/projects/govmomi/simulator/session_manager.go:390 +0xeb
  github.com/vmware/govmomi/simulator.(*Registry).WithLock()
      /Users/asaba/projects/govmomi/simulator/registry.go:630 +0x5e
  github.com/vmware/govmomi/simulator.(*Context).WithLock()
      /Users/asaba/projects/govmomi/simulator/session_manager.go:382 +0xe4
  github.com/vmware/govmomi/simulator.(*Context).postEvent()
      /Users/asaba/projects/govmomi/simulator/session_manager.go:388 +0x18
  github.com/vmware/govmomi/simulator.(*powerVMTask).Run()
      /Users/asaba/projects/govmomi/simulator/virtual_machine.go:1461 +0x85c
  github.com/vmware/govmomi/simulator.(*powerVMTask).Run-fm()
      /Users/asaba/projects/govmomi/simulator/virtual_machine.go:1437 +0x44
  github.com/vmware/govmomi/simulator.(*Task).Run.func1()
      /Users/asaba/projects/govmomi/simulator/task.go:112 +0x9e

Previous write at 0x00c0001c2d68 by goroutine 25:
  github.com/vmware/govmomi/simulator.(*EventManager).PostEvent()
      /Users/asaba/projects/govmomi/simulator/event_manager.go:159 +0x8a
  github.com/vmware/govmomi/simulator.(*Context).postEvent.func1()
[...]

PR #2651, opened 2021-11-05
Fixes data race in simulator.EventManager.PostEvent

awsaba:
## Description

This PR introduces independent `simulator.Context` for the multiple tasks spawned by `PowerOnMultiVM`.

Closes: #2650

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [ ] `make test`
- [ ] `TEST_COUNT=10 make test`

## Checklist:

  this project

### DIFF
--- a/simulator/datacenter.go
+++ b/simulator/datacenter.go
@@ -163,10 +163,21 @@ func (dc *Datacenter) PowerOnMultiVMTask(ctx *Context, req *types.PowerOnMultiVM
 
 		for _, ref := range req.Vm {
 			vm := ctx.Map.Get(ref).(*VirtualMachine)
+
+			// This task creates multiple subtasks which violates the assumption
+			// of 1:1 Context:Task, which results in data races in objects
+			// like the Simulator.Event manager. This is the minimum context
+			// required for the PowerOnVMTask to complete.
+			taskCtx := &Context{
+				Context: ctx.Context,
+				Session: ctx.Session,
+				Map:     ctx.Map,
+			}
+
 			// NOTE: Simulator does not actually perform any specific host-level placement
 			// (equivalent to vSphere DRS).
-			ctx.WithLock(vm, func() {
-				vmTaskBody := vm.PowerOnVMTask(ctx, &types.PowerOnVM_Task{}).(*methods.PowerOnVM_TaskBody)
+			taskCtx.WithLock(vm, func() {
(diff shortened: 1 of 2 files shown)


dougm (commented):
Thanks @awsaba .  The `EventManager` is locked by default when `PostEvent` is called via the API, using the same locking mechanism applied to all other managed objects: https://github.com/vmware/govmomi/blob/14d8083c9ca296ce3c3e6d4fe21e455119f56db5/simulator/simulator.go#L217-L219

And when `postEvent` is called internally, we lock the EventManager:
https://github.com/vmware/govmomi/blob/14d8083c9ca296ce3c3e6d4fe21e455119f56db5/simulator/session_manager.go#L387-L392

The problem with #2650 is with the tasks being async and the `Context` being shared between the task goroutines. Hence why waiting on each task to complete avoids the race (as mentioned here https://github.com/vmware/govmomi/issues/2650#issuecomment-962141039)

In the `PowerOnMultiVMTask` method `Context:Task` ratio is `1:N`, whereas with other API methods `Context:Task` is `1:1`, so that's why we don't hit this race elsewhere.

We could leave `PowerOnMultiVMTask` as-is and do explicit locking in `PostEvent` as your PR does. But if we do that, EventManager should also opt-out of the default locking mechanism, as PropertyCollector does for example: https://github.com/vmware/govmomi/blob/14d8083c9ca296ce3c3e6d4fe21e455119f56db5/simulator/property_collector.go#L41
And `postEvent` would no longer need to use `WithLock`. What do you think?

awsaba:
I was hoping the only thing that needed to be synchronized would be the `(P|p)ostEvent` methods that are approximating a message queue, but looking at the other methods that would be need to use the lock on the EventManager manager struct, I think it would create more potential for deadlocks. This call to `fillPage` for example:

https://github.com/vmware/govmomi/blob/master/simulator/event_manager.go#L388-L402

```go
func (c *EventHistoryCollector) SetCollectorPageSize(ctx *Context, req *types.SetCollectorPageSize) soap.HasFault {
	body := new(methods.SetCollectorPageSizeBody)
	size, err := validatePageSize(req.MaxCount)
	if err != nil {
		body.Fault_ = err
		return body
	}

	c.size = size
	c.page = list.New()
	ctx.WithLock(c.m, c.fillPage)

	body.Res = new(types.SetCollectorPageSizeResponse)
	return body
}
```

I've changed the branch to create independent `Context`s per-task which resolves the issue for `PowerOnMultiVM`.

dougm (approved):
lgtm, thanks @awsaba

PR #2682, opened 2021-12-10
vcsim: add support for cloning HostSystems

dougm:
## Description

The simulator.ClusterComputeResource.AddHost method uses the compiled in esx.HostSystem template by default,
as does simulator.AddStandaloneHost.
With this change, if the HostConnectSpec.UserName is that of a HostSystem.Name in the given Cluster or
of a standalone host, use that HostSystem as the template.

## Type of change

  not work as expected)
- [ ] Build related change

## Checklist:

  this project

### DIFF
--- a/simulator/host_system.go
+++ b/simulator/host_system.go
@@ -58,11 +58,15 @@ func NewHostSystem(host mo.HostSystem) *HostSystem {
 	hs.Summary.Runtime = &hs.Runtime
 	hs.Summary.Runtime.BootTime = &now
 
+	// shallow copy Summary.Hardware, as each host will be assigned its own .Uuid
 	hardware := *host.Summary.Hardware
 	hs.Summary.Hardware = &hardware
 
-	info := *esx.HostHardwareInfo
-	hs.Hardware = &info
+	if hs.Hardware == nil {
+		// shallow copy Hardware, as each host will be assigned its own .Uuid
+		info := *esx.HostHardwareInfo
+		hs.Hardware = &info
+	}
 
 	cfg := new(types.HostConfigInfo)
 	deepCopy(hs.Config, cfg)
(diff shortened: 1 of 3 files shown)


simulator/host_system.go
```diff
@@ -61,8 +61,10 @@ func NewHostSystem(host mo.HostSystem) *HostSystem {
 	hardware := *host.Summary.Hardware
 	hs.Summary.Hardware = &hardware
 
-	info := *esx.HostHardwareInfo
-	hs.Hardware = &info
+	if hs.Hardware == nil {
+		info := *esx.HostHardwareInfo
+		hs.Hardware = &info
+	}
```
embano1, line 67: question: could this be simplified to:

```suggestion
	if hs.Hardware == nil {
		hs.Hardware = *esx.HostHardwareInfo
	}
```

embano1: Ah, sorry seems like this is supposed to be a copy of the object. Any DeepCopy concerns on that struct (haven't checked the struct fields)?

embano1 (approved):
LGTM (just one question).

Regarding the commit title it seems it could be reworded to "support" instead of "supporting"

simulator/host_system.go
dougm: Shallow copy is enough, the only field we ever change is the Uuid: https://github.com/vmware/govmomi/blob/afc67f589a3d1bcb066ac2a0724a1cace1b3f4a9/simulator/host_system.go#L103

I'll add a comment on that w/ the next push.

dougm:
> Regarding the commit title it seems it could be reworded to "support" instead of "supporting"

Whoops, will fix

embano1 (approved):
LGTM

PR #2688, opened 2021-12-21
Helpers for enc/dec VIM objects to/from XML

akutz:
## Description

This patch provides several new helper functions for encoding/decoding VIM objects to/from XML:

* The decoder functions can return multiple VIM objects from a `string`, `byte` slice, `io.Reader`, or `xml.Decoder`, all without needing to know the type of the objects in advance!
* The encoder functions produce output that is consumable by other vSphere SDKs for Java and Python.
* The file [`util_examples_test.go`](https://github.com/akutz/govmomi/blob/feature/xml-helpers/vim25/xml/util/util_examples_test.go) includes two examples for using the encoder and decoder functions.
* The file [`util_test.go`](https://github.com/akutz/govmomi/blob/feature/xml-helpers/vim25/xml/util/util_test.go) provides 100% code coverage!
* The directory [`vim25/xml/util/examples`](https://github.com/akutz/govmomi/blob/feature/xml-helpers/vim25/xml/util/examples) includes code and documentation that illustrates how to use the new helper functions to share VIM objects between Go and Python.

Closes: `na`

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [x] Unit tests
- [x] Example tests
- [x] Documented (with code) use cases with Go and Python

## Checklist:

  this project

### DIFF
--- a/vim25/xml/util/examples/README.md
+++ b/vim25/xml/util/examples/README.md
@@ -1,0 +1,75 @@
+# Sharing VIM objects between Golang and Python using XML
+
+This directory includes several examples of how to share VIM objects between multiple languages, such as Golang and Python, by marshaling and unmarshaling those objects to and from XML.
+
+# Requirements
+
+* Golang 1.16+
+* Python 3.9+
+* PyVmomi (use `pip3 install --user pyvmomi` to make sure you have the latest version)
+
+# Marshal a ConfigSpec to XML
+
+The following examples demonstrate how to marshal a simple `ConfigSpec` object to XML to stdout in Golang and Python:
+
+## Golang
+
+```bash
+$ go run main.go
+<obj xmlns:vim25="urn:vim25" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="vim25:VirtualMachineConfigSpec">
+  <name>go-vm</name>
+  <numCPUs>2</numCPUs>
+  <memoryMB>2048</memoryMB>
+</obj>
+```
+
+## Python
+
+```bash
+$ python3 main.py
+<?xml version="1.0" ?>
+<object xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns="urn:vim25" xsi:type="VirtualMachineConfigSpec">
+  <name>python-vm</name>
+  <numCPUs>2</numCPUs>
+  <memoryMB>2048</memoryMB>
+</object>
+```
+
+# Unmarshal XML to a ConfigSpec
+
+The following examples demonstrate how to unmarshal XML to a `ConfigSpec` and back to stdout in Golang and Python:
+
+## Golang
+
+Please note the output will reflect the name from the object created in Python, `python-vm`, which indicates the resource was:
+
+1. Marshaled to XML in Python
+1. Unmarshaled to XML in Go
+1. Marshaled _back_ to XML in Go
+
+```bash
+$ python3 main.py | go run main.go -
+<obj xmlns:vim25="urn:vim25" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="vim25:VirtualMachineConfigSpec">
+  <name>python-vm</name>
+  <numCPUs>2</numCPUs>
+  <memoryMB>2048</memoryMB>
+</obj>
+```
+
+## Python
+
+Please note the output will reflect the name from the object created in Go, `go-vm`, which indicates the resource was:
+
+1. Marshaled to XML in Go
+1. Unmarshaled to XML in Python
+1. Marshaled _back_ to XML in Python
+
+```bash
+$ go run main.go | python3 main.py -
+<?xml version="1.0" ?>
+<object xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns="urn:vim25" xsi:type="VirtualMachineConfigSpec">
+  <name>go-vm</name>
+  <numCPUs>2</numCPUs>
+  <memoryMB>2048</memoryMB>
+</object>
+```
--- a/vim25/xml/util/util.go
+++ b/vim25/xml/util/util.go
@@ -1,0 +1,243 @@
+/*
+Copyright (c) 2021 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package util implements some helper functions for marshaling/unmarshaling
+// vim objects to/from XML.
+package util
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"reflect"
+	"strings"
+
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/govmomi/vim25/xml"
+)
+
+// VimObjectsFromBytes returns all vim objects decoded from the provided bytes.
+//
+// This is a helper function that wraps VimObjectsFromDecoder. Please see
+// the documentation on VimObjectsFromDecoder for additional information.
+func VimObjectsFromBytes(v []byte) (<-chan interface{}, <-chan error) {
+	return VimObjectsFromReader(bytes.NewReader(v))
+}
+
+// VimObjectsFromString returns all vim objects decoded from the provided
+// string.
+//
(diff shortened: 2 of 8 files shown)


vim25/xml/util/examples/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Unmarshal XML to a ConfigSpec
+
+The following examples demonstrate how to unmarshal XML to a `ConfigSpec` and back to stdout in Golang and Python:
+
+## Golang
+
+Please note the output will reflect the name from the object created in Python, `python-vm`, which indicates the resource was:
+
+1. Marshled to XML in Python
```
embano1, line 46: nit: typo (also below)

vim25/xml/util/util.go
```diff
@@ -0,0 +1,241 @@
...
+		return vimTypeName
+	}
+
+	// Decode the input stream in the background, returning tokens/errors on
+	// the chanObj and chanErr channels.
+	go func() {
+
+		// Keep iterating until there are no more tokens.
+		for {
+
```
embano1, line 102: nit: remove newline (same for other places)

vim25/xml/util/util.go
```diff
@@ -0,0 +1,241 @@
...
+			if err := d.DecodeElement(vimObj, &startElement); err != nil {
+				chanErr <- err
+				break
+			}
+
+			// Send the decoded object into the channel.
+			chanObj <- vimObj
+		}
+
+		// Once there are no more tokens and/or an error occurred, the loop
```
embano1, line 161: nit: I usually put those into a `defer()` at the top of the fn for visibility.

vim25/xml/util/util.go
```diff
@@ -0,0 +1,241 @@
...
+// This is a helper function that wraps VimObjectsToEncoder. Please see
+// the documentation on VimObjectsToEncoder for additional information.
+func VimObjectsToWriter(w io.Writer, objs ...interface{}) error {
+	return VimObjectsToEncoder(xml.NewEncoder(w), objs...)
+}
+
+// VimObjectsToEncoder marshals vim objects using an xml.Encoder so the result
+// can be unmarshaled by other vSphere SDKs, such as Java, Python, Ruby, etc.
+func VimObjectsToEncoder(e *xml.Encoder, objs ...interface{}) error {
+
```
embano1, line 209: nit: unnecessary newline (I think you did this for readability?)

vim25/xml/util/util_examples_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Decode all of the objects from the XML string.
+	chanObj, chanErr := util.VimObjectsFromString(vimObjects)
+
+	// For each object we see, print its type and information about it, along
+	// with any errors that occur along the way.
+	for {
+		select {
+		case obj := <-chanObj:
+			if obj == nil {
```
embano1, line 58: IMHO this pattern can lead to a situation where the `chanErr` is not correctly drained. A closed channel always returns the zero value (causing hot loop and thus likelihood of being scheduled before the other channel has a chance to fire).

In `util.go` once the goroutine loop is done, `chanObj` is closed before `chanErr`. So this would return too early (same in `util_test.go`).

The general pattern in these cases is to instruct the `for` loop to stop on a condition variable and set the channel to `nil` once it's closed, e.g. pseudo-code:

```go
// You can edit this code!
// Click here and start typing.
package main

import "time"

func main() {
	var (
		errChan chan error
		objChan chan interface{}
	)

	go func() {
		time.Sleep(2 * time.Second)
		close(errChan)
		close(objChan)
	}()

	for objChan != nil && errChan != nil {
		select {
		case o, ok := <-objChan:
			if !ok {
				objChan = nil

			}

		case err, ok := <-errChan:
			if !ok {
				errChan = nil
			}
		}
	}
}
```

Since the receiver doesn't have control over the channel in your case (read-only), someone has to set it to `nil` or the channel must be made writeable.

akutz: Thanks for the review. I knew when I wrote the code that I was forgetting something about channels (the implicit tuple they return with the `ok bool` if the channel is closed), but I just didn't spend too much time on it. As for the other pattern you recommended:

* I do not really care too much about the draining here since it is example code and I know exactly how it will behave
* I've never honestly seen this patter, but regardless, I do not think it is necessary here
* Even if a channel is closed, it can still be drained by a reader. See this Go playground [example](https://go.dev/play/p/oXo2tduEMv0)

    ```go
    package main
    
    import (
    	"fmt"
    	"time"
    )
    
    func main() {
    	c := make(chan int)
    	go func() {
    		c <- 1
    		close(c)
    	}()
    	time.Sleep(3 * time.Second)
    	a, ok := <-c
    	fmt.Printf("%d %v\n", a, ok)
    }
    ```

    The above example will print:

    ```bash
    1 true
    ```

    Thus it is fine to close a channel if there is still data on that channel. It will be read until there is no more data.

vim25/xml/util/util.go
akutz: I prefer the newlines for readability. Unless you feel strongly, I prefer to leave them.

dougm (commented):
Thanks @akutz 
`vim25/xml` is a fork of Go's `encoding/xml` package.. we should avoid adding a package within as it'll make rebasing against upstream trickier. Ideally we'd upstream the functionality in our fork someday and `vim25/xml` will go away.

It looks like in a nutshell, this PR adds Go versions of pyvmomi's `SoapAdapter.Deserialize` and `SoapAdapter.SerializeToUnicode`.  If that's the case, how
about we add these to the `vim25/soap` package and drop the To/From helpers:
```go
func Deserialize(d *xml.Decoder) (<-chan interface{}, <-chan error) // was VimObjectsFromDecoder

func Serialize(e *xml.Encoder, objs ...interface{}) error // was VimObjectsToEncoder
```
We could name them something else (e.g. Decode/Encode) of course, but adding these 2 funcs to the `vim25/soap` package would be fine with me, rather than creating a
new util package. What do you think?

btw, I ran the main.py example with pyvmomi 7.0.3, it outputs `<object ...>` rather than `<obj ...>`. Maybe that was changed in 8.0?

vim25/xml/util/util.go
dougm: I don't feel strongly about it, but I personally do not add newlines after a line ending w/ a brace as I find the indent provides readability. Either way is fine with me though.

akutz:
Thank you for looking @dougm!

> btw, I ran the main.py example with pyvmomi 7.0.3, it outputs `<object ...>` rather than `<obj ...>`. Maybe that was changed in 8.0?

It is actually Java that uses `object`, but it was too difficult to provide a simple example using the upstream Java SDK, so I used the Python one instead :)

akutz:
> We could name them something else (e.g. Decode/Encode) of course, but adding these 2 funcs to the vim25/soap package would be fine with me, rather than creating a new util package. What do you think?

This is fine with me. I just did not want to pollute the existing package without chatting with you about it first. I'll rebase based on your suggestions.

vim25/xml/util/util.go
embano1: no all good, just wanted to double-check

vim25/xml/util/util_examples_test.go
embano1: Just looked at the code again and since you're using an unbuffered `err` channel one issue I had in mind cannot occur here. My concern was not draining a closed channel but missing an error due to runtime scheduling. Unbuffered channel and your current logic makes this impossible IMHO 👍

vim25/xml/util/util.go
```diff
@@ -0,0 +1,243 @@
...
+// the documentation on VimObjectsFromDecoder for additional information.
+func VimObjectsFromReader(r io.Reader) (<-chan interface{}, <-chan error) {
+	return VimObjectsFromDecoder(xml.NewDecoder(r))
+}
+
+// VimObjectsFromDecoder returns all vim objects from the provided decoder.
+//
+// Any decode errors are sent on the error channel.
+//
+// Both channels are closed when io.EOF is reached.
```
embano1, line 63: nit: there's another err scenario when both channels are closed. This should be documented.

vim25/xml/util/util_examples_test.go
akutz: The above [example](https://go.dev/play/p/GhNppD6ZtWL) still works with buffered channels as well 😃 

```golang
package main

import (
	"fmt"
	"time"
)

func main() {
	c := make(chan int, 2)
	go func() {
		c <- 1
		c <- 2
		close(c)
	}()
	time.Sleep(3 * time.Second)
	a, ok := <-c
	fmt.Printf("%d %v\n", a, ok)
	a, ok = <-c
	fmt.Printf("%d %v\n", a, ok)
}
```

The above example prints:

```bash
1 true
2 true
```

akutz: The point is that closing a channel does not prevent the channel from being drained.

embano1: Sorry for not being clear. With buffered channels there's this issue https://go.dev/play/p/CqFQGdiqMlv.

Run it multiple times and you'll see. I just realized that this won't happen in your case since you're using unbuffered.

akutz: I see what you are saying. This only matters if you actually want to ensure all of the information is read off all of the channels though. Either way, I get your point, but I still think it depends on the context of the situation, not whether or not the channel is buffered.

Issue #2696, opened 2022-01-07
[BUG] Assuming IPv4 Address When Signing API Request


yuyin002:
**Describe the bug**
Hostname that's an IPv6 address is expected to be surrounded by square brackets (like in URL), however govmomi uses url.Hostname() which strips those square brackets: https://github.com/vmware/govmomi/blob/master/sts/signer.go#L281

https://pkg.go.dev/net/url#URL.Hostname says:
Hostname returns u.Host, stripping any valid port number if present.
If the result is enclosed in square brackets, as literal IPv6 addresses are, the square brackets are removed from the result.

This may lead to authentication failure for services that use govmomi.

**To Reproduce**
Reproduce should behave differently for different services that use govmomi. In our case, set-ups using IPv6 pure can not handle API requests and will return Authentication failure.

**Expected behavior**
govmomi can handle both IPv4 and IPv6 addresses for signing API requests.

**Affected version**
Please provide details on the version used, e.g. release tag, commit, module version, etc.

**Screenshots/Debug Output**
If applicable, add screenshots or debug output to help explain your problem.

**Additional context**

PR #2701, opened 2022-01-11
fix: Add IPv6 support for signing HTTP request

yuyin002:
Closes: #2696

## Description
Hostname that's an IPv6 address is expected to be surrounded by square brackets (like in URL), however govmomi uses url.Hostname() which strips those square brackets: https://github.com/vmware/govmomi/blob/master/sts/signer.go#L281

https://pkg.go.dev/net/url#URL.Hostname says:
Hostname returns u.Host, stripping any valid port number if present.
If the result is enclosed in square brackets, as literal IPv6 addresses are, the square brackets are removed from the result.

This may lead to authentication failure for services that use govmomi. This change added the check for IPv6 host and add the brackets.

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?
Previously, vCenter deployed with pure IPv6 failed to authenticate dcli request; after this change, the dcli request can be successfully authenticated.

- [ ] Test Description 1
- [ ] Test Description 2

## Checklist:

  this project
NA
```
ok  	github.com/vmware/govmomi/sts	0.291s	coverage: 47.1% of statements
```

Note: This test is failing before & after this change was made
```
guest_operations_manager_test.go:48: exit status 1
--- FAIL: TestFileInfo (0.02s)
```
NA

### DIFF
--- a/sts/signer.go
+++ b/sts/signer.go
@@ -30,6 +30,7 @@ import (
 	"io"
 	"io/ioutil"
 	mrand "math/rand"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -274,11 +275,18 @@ func (s *Signer) SignRequest(req *http.Request) error {
 		}
 
 		var buf bytes.Buffer
+		host := req.URL.Hostname()
+
+		// Check if the host IP is in IPv6 format. If yes, add the opening and closing square brackets.
+		if isIPv6(host) {
+			host = fmt.Sprintf("%s%s%s", "[", host, "]")
+		}
+
 		msg := []string{
 			nonce,
 			req.Method,
 			req.URL.Path,
-			strings.ToLower(req.URL.Hostname()),
+			strings.ToLower(host),
 			port,
 		}
 		for i := range msg {
@@ -328,4 +336,12 @@ func (s *Signer) NewRequest() TokenRequest {
 		Userinfo:    s.user,
 		KeyID:       s.keyID,
 	}
+}
+
+func isIPv6(s string) bool {
+	ip := net.ParseIP(s)
+	if ip == nil {
+		return false
+	}
+	return len(ip) == net.IPv6len
 }
--- a/go.mod
+++ b/go.mod
@@ -14,4 +14,4 @@ require (
 	github.com/vmware/vmw-guestinfo v0.0.0-20170707015358-25eff159a728
 )
 
-require github.com/kr/text v0.1.0 // indirect
+require	github.com/kr/text v0.1.0 // indirect
--- a/sts/signer_test.go
+++ b/sts/signer_test.go
@@ -133,4 +133,10 @@ func TestSigner(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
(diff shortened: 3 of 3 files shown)


sts/signer.go
```diff
@@ -274,11 +274,20 @@ func (s *Signer) SignRequest(req *http.Request) error {
 		}
 
 		var buf bytes.Buffer
+		host := req.URL.Hostname()
+		// Check if the host IP is in IPv6 format. If yes, add the opening and closing square brackets.
+		// If the host IP contains ":", it is either an IPv6 address or the entire host-port input if the port is not valid.
+		// https://github.com/golang/go/blob/master/src/net/url/url.go#L1148
```
zach96guan, line 280: Nit: should we point to the method name in that file, as line number may change right away

sts/signer.go
```diff
@@ -274,11 +274,20 @@ func (s *Signer) SignRequest(req *http.Request) error {
...
+		// So we should rule out the possibility when IPv4:port is returned.
+		if strings.Contains(host, ":") && !strings.Contains(host, ".") {
+			host = fmt.Sprintf("%s%s%s", "[", host, "]")
+		}
```
zach96guan, line 284: Have some concerns here.
- need to check if this is the right way to distinguish ip format. Per discussion in https://stackoverflow.com/questions/22751035/golang-distinguish-ipv4-ipv6, seems we could use the `ParseIP` method from `net` package
- not a blocker, should we add checkers say `IsIPv4` and `IsIPv6`, and invoke in `SignRequest`, considering they would be reused somewhere else

zach96guan (dismissed):
Thanks for the fix!!

sts/signer.go
yuyin002: Thanks Zach for the review. Yes updated the comment.

sts/signer.go
yuyin002: Good point. I used `govalidator` instead, which has IsIPv6 check.

zach96guan (dismissed):
lgtm!

sts/signer.go
```diff
@@ -274,11 +279,18 @@ func (s *Signer) SignRequest(req *http.Request) error {
 		}
 
 		var buf bytes.Buffer
+		host := req.URL.Hostname()
+
+		// Check if the host IP is in IPv6 format. If yes, add the opening and closing square brackets.
+		if govalidator.IsIPv6(host) {
```
dougm, line 285: Rather that pull in a dependency on the govalidator package, we can check using Go's stdlib:
```go
func isIPv6(s string) bool {
	ip := net.ParseIP(s)
	if ip == nil {
		return false
	}
	return len(ip) == net.IPv6len
}
```

dougm (commented):
Thanks @yuyin002 , looks good. Can we just avoid adding the govalidator depenency?

sts/signer.go
yuyin002: Thanks! I removed the govalidator dependency and added the `isIPv6` function.

dougm (approved):
Thanks @yuyin002 !

Issue #2672, opened 2021-11-27
Add Appliance Shutdown APIs


pradeep288:
**Is your feature request related to a problem? Please describe.**
Currently, govmomi doesn't support all the appliance management APIs. We need to support most of these API's

**Describe the solution you'd like**
As a first step to support most of the appliance mgmt API, we need to add Appliance Shutdown APIs.
We can take this as a reference, to add the API's
https://developer.vmware.com/apis/vsphere-automation/latest/appliance/appliance/shutdown/

PR #2721, opened 2022-01-27
Add Appliance shutdown API

pradeep288:
Closes: #2672

## Description

This PR adds Appliance shutdown API's mentioned https://developer.vmware.com/apis/vsphere-automation/latest/appliance/appliance/shutdown/

Closes: #2672 

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- Tested Manually on live vCenter Server
- Added the govc-test

## Checklist:

- [X] My code follows the CONTRIBUTION [guidelines](./../CONTRIBUTING.md) of
  this project

### DIFF
--- a/vapi/simulator/simulator.go
+++ b/vapi/simulator/simulator.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+Copyright (c) 2018-2022 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -43,6 +43,7 @@ import (
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/ovf"
 	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/vapi/appliance/shutdown"
 	"github.com/vmware/govmomi/vapi/internal"
 	"github.com/vmware/govmomi/vapi/library"
 	"github.com/vmware/govmomi/vapi/rest"
@@ -91,6 +92,7 @@ type handler struct {
 	Library     map[string]*content
 	Update      map[string]update
 	Download    map[string]download
+	Config      shutdown.Config
 }
 
 func init() {
@@ -114,6 +116,7 @@ func New(u *url.URL, settings []vim.BaseOptionValue) (string, http.Handler) {
 		Library:     make(map[string]*content),
 		Update:      make(map[string]update),
 		Download:    make(map[string]download),
+		Config:      shutdown.Config{},
 	}
 
 	handlers := []struct {
@@ -155,6 +158,7 @@ func New(u *url.URL, settings []vim.BaseOptionValue) (string, http.Handler) {
 		{internal.VCenterVMTXLibraryItem + "/", s.libraryItemTemplateID},
 		{internal.VCenterVM + "/", s.vmID},
 		{internal.DebugEcho, s.debugEcho},
+		{shutdown.Path, s.shutdown},
 	}
 
 	for i := range handlers {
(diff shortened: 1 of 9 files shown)


embano1 (approved):
Thx, LGTM

vapi/simulator/simulator.go
```diff
@@ -329,6 +333,16 @@ func OK(w http.ResponseWriter, val ...interface{}) {
 	StatusOK(w, s)
 }
 
+// OKWithNoContent responds with http.StatusNoContent.
+func OKWithNoContent(w http.ResponseWriter) {
```
dougm, line 337: Thanks @pradeep288 , lgtm overall. Just a nit here, these are two functions that do the same thing and I think we can just use the `Status` func instead of either: 
```diff
modified   vapi/simulator/simulator.go
@@ -333,16 +333,6 @@ func OK(w http.ResponseWriter, val ...interface{}) {
 	StatusOK(w, s)
 }
 
-// OKWithNoContent responds with http.StatusNoContent.
-func OKWithNoContent(w http.ResponseWriter) {
-	OKStatusNoContent(w)
-}
-
-// OKStatusNoContent responds with http.StatusNoContent.
-func OKStatusNoContent(w http.ResponseWriter) {
-	Status(w, http.StatusNoContent)
-}
-
 // BadRequest responds with http.StatusBadRequest and json encoded vAPI error of type kind.
 // For use with "/rest" endpoints where the response is a "value" wrapped structure.
 func BadRequest(w http.ResponseWriter, kind string) {
@@ -2138,7 +2128,7 @@ func (s *handler) shutdown(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Query().Get(shutdown.Action) {
 		case shutdown.Cancel:
 			delayShutdown(0)
-			OKWithNoContent(w)
+			Status(w, http.StatusNoContent)
 		case shutdown.Reboot:
 			var spec shutdown.Spec
 			if s.decode(r, w, &spec) {
@@ -2146,7 +2136,7 @@ func (s *handler) shutdown(w http.ResponseWriter, r *http.Request) {
 				s.Config.Reason = spec.Reason
 				s.Config.Action = shutdown.Reboot
 				delayShutdown(spec.Delay)
-				OKWithNoContent(w)
+				Status(w, http.StatusNoContent)
 			}
 		case shutdown.PowerOff:
 			var spec shutdown.Spec
@@ -2155,7 +2145,7 @@ func (s *handler) shutdown(w http.ResponseWriter, r *http.Request) {
 				s.Config.Reason = spec.Reason
 				s.Config.Action = shutdown.PowerOff
 				delayShutdown(spec.Delay)
-				OKWithNoContent(w)
+				Status(w, http.StatusNoContent)
 			}
 		default:
 			http.NotFound(w, r)
```

dougm (changes requested):

vapi/simulator/simulator.go
pradeep288: @dougm thanks for reviewing. The reason for creating a new function is to be consistent with the current pattern.

Currently, the pattern is

OK -> StatusOK -> Status 

to keep consistent, I added a new function and my pattern looks like

OKWithNoContent -> OKStatusNoContent -> Status


What's your opinion, shall I change as per your suggestion?

dougm: `vapi.OK` is a helper for the `/rest` endpoint (with `Value` encoding)  and `StatusOK` was added for the 7.0 `/api` endpoint response encoding. In the case of `http.StatusNoContent` there's no response body, so adding another helper function isn't needed.  You could even just call `w.WriteHeader(http.StatusNoContent)`, instead of using `Status`.  In other words, we don't need a helper for each HTTP response code, we've only added helpers for those that response with an encoded response body.

dougm: And looking at this closer, we don't need the recently added `Status` method that was recently refactored out of `StatusOK`: https://github.com/vmware/govmomi/pull/2722

pradeep288: so suggestion is to use
`w.WriteHeader(http.StatusNotFound)` instead of `Status(w, http.StatusNoContent)`

dougm: Yes

pradeep288: Changed `OKWithNoContent(w)` to `w.WriteHeader(http.StatusNotFound)`

vapi/simulator/simulator.go
```diff
@@ -2105,3 +2109,48 @@ func (s *handler) vmID(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	switch r.Method {
+	case http.MethodGet:
+		OK(w, s.Config)
+	case http.MethodPost:
+		switch r.URL.Query().Get(shutdown.Action) {
+		case shutdown.Cancel:
+			delayShutdown(0)
+			w.WriteHeader(http.StatusNotFound)
```
dougm, line 2131: These should be `http.StatusNoContent` instead of NotFound, right?

pradeep288: Sorry, Now i have updated to `w.WriteHeader(http.StatusNoContent)`

dougm (approved):
Thanks @pradeep288 !

PR #2668, opened 2021-11-24
govc: add support for supervisor services deploy

rikatz:
Signed-off-by: Ricardo Katz <rkatz@vmware.com>

## Description

Initial work on supporting vSphere Namespaces supervisor services.

Right now it's possible to:
* List
* Enable
* Disable
* Remove
* Create

TODO: Write VCSIM things

## Type of change

Please mark options that are relevant:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [ ] This change requires a documentation update
- [ ] Build related change

## How Has This Been Tested?

Tested with a real cluster. Still will check to write some unit tests

## Checklist:

- [ ] My code follows the CONTRIBUTION [guidelines](./../CONTRIBUTING.md) of
  this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged

PR #2722, opened 2022-01-28
vcsim: revert vapi.Status() method

dougm:
## Description

In PR #2668 we refactored the 'StatusOK' method into a 'Status' method, which we don't need without a response body,
as w.WriteHeader can be used directly instead.

### DIFF
--- a/vapi/simulator/simulator.go
+++ b/vapi/simulator/simulator.go
@@ -295,10 +295,10 @@ func (s *handler) DetachTag(id vim.ManagedObjectReference, tag vim.VslmTagEntry)
 	return nil
 }
 
-// Status responds with custom HTTP status and json encoded val if given.
+// StatusOK responds with http.StatusOK and encodes val, if specified, to JSON
 // For use with "/api" endpoints.
-func Status(w http.ResponseWriter, httpStatus int, val ...interface{}) {
-	w.WriteHeader(httpStatus)
+func StatusOK(w http.ResponseWriter, val ...interface{}) {
+	w.WriteHeader(http.StatusOK)
 	if len(val) == 0 {
 		return
 	}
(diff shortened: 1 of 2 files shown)


dougm:
See comments: https://github.com/vmware/govmomi/pull/2721#discussion_r794176022

vapi/simulator/simulator.go
```diff
@@ -291,10 +291,10 @@ func (s *handler) DetachTag(id vim.ManagedObjectReference, tag vim.VslmTagEntry)
 	return nil
 }
 
-// Status responds with custom HTTP status and json encoded val if given.
+// StatusOK responds with http.StatusOK and json encoded val if given.
```
embano1, line 294:
```suggestion
// StatusOK responds with http.StatusOK and encodes val, if specified, to JSON
```

embano1: I found the description to be confusing as I assumed val has to be JSON encoded (which is not true). You might want to clean up the other methods too.

vapi/simulator/simulator.go
```diff
@@ -306,12 +306,6 @@ func Status(w http.ResponseWriter, httpStatus int, val ...interface{}) {
...
 
-// StatusOK responds with http.StatusOK and json encoded val if given.
-// For use with "/api" endpoints.
-func StatusOK(w http.ResponseWriter, val ...interface{}) {
-	Status(w, http.StatusOK, val...)
-}
-
 // OK responds with http.StatusOK and json encoded val if given.
 // For use with "/rest" endpoints where the response is a "value" wrapped structure.
 func OK(w http.ResponseWriter, val ...interface{}) {
```
embano1, line 311: nit: perhaps in the future we can revisit this API as it‘s confusing to have StatusOK and OK functions. I understand from the comment that the difference is in the response wrapping.

dougm: Agreed!

embano1 (commented):
nit: while you are at it, can you please also update the doc string for `func OK` regarding the JSON encoding? Then feel free to squash so we don‘t need another approval round here :)

PR #2747, opened 2022-02-09
fix: rest.Client.LoginByToken invalid signature

dougm:
## Description

PR #2701 (ebeaa71b) added IPv6 support for signing HTTP request,
but was treating all IPs as v6.

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

Added tests and ran against real vCenter using `govc session.login -r ...`

## Checklist:

  this project

### DIFF
--- a/sts/client_test.go
+++ b/sts/client_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/vmware/govmomi/session"
 	"github.com/vmware/govmomi/ssoadmin"
 	"github.com/vmware/govmomi/ssoadmin/types"
+	"github.com/vmware/govmomi/vapi/rest"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/methods"
 	"github.com/vmware/govmomi/vim25/soap"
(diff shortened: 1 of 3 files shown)


yuyin002:
> @yuyin002 can you review? My suggestion of using `len(ip) == net.IPv6len` was wrong, see: https://stackoverflow.com/questions/22751035/golang-distinguish-ipv4-ipv6/22752227
> 
> Also when using a FQDN, `isIPv6` will always return false.. is that an issue for you or would the IPv6 address always be used in your use-case?

Thanks for the fix @dougm, looks good to me! `when using a FQDN, `isIPv6` will always return false.` I think this should be expected. We use `isIPv6` to identify IPv6 IP only, because we need to add the square brackets around the IPv6 IP if that's the host. With FQDN, we shouldn't worry about the square brackets.

Seems like I don't have the permission to approve this MR though..

dougm:
Thanks @yuyin002 , added you to the team so you should be able to approve now.

kay-ge (approved):
LGTM, thanks for fixing.

yuyin002 (approved):

Issue #2806, opened 2022-04-05
How to set disk IOPS limit using govc?


embano1:
### Discussed in https://github.com/vmware/govmomi/discussions/2801

<div type='discussions-op-text'>

<sup>Originally posted by **Heiko-san** March 28, 2022</sup>
Hi,

What I want to achieve is to set the IOPS limit for the disk of an existing vm.
I mean the setting that goes into vm's config as "sched.scsi0:0.throughputCap=xy", which you set in GUI by "edit settings" -> go to the disk -> "Limit IOPs"  -> if you switch to "user defined" you can enter any number ...

- I found `govc vm.disk.change` but it seems I can only set disk size there.
- I see the disk using `govc device.info -vm myvm disk-1000-0`
- There is also a `govc disk.ls`, but it doesn't show anything

How can I set this limit?

Update:
- I found this, but doesn't seem to work either (even if it doesn't show any errors): `govc vm.change -vm myvm -e sched.scsi0:0.throughputCap=123`</div>

PR #2807, opened 2022-04-05
govc: Edit disk storage IO

embano1:
## Description

Allow setting disk IO limits (or `-1` for unlimited). vCenter validates values and returns error message when a limit is invalid, e.g. `-disk.io.limit 0`.

Closes: #2806

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [x] Various `govc` runs against vCenter 6.7U3 setting/removing (invalid) disk IO limits

```console
./govc vm.disk.change -vm /vcqaDC/vm/vm-with-disk -disk.key 2000 -disk.io.limit -1
./govc object.collect -json /vcqaDC/vm/vm-with-disk config.hardware.device | jq '.[0].Val.VirtualDevice[]|select(.Key==2000)|.StorageIOAllocation'
{
  "Limit": -1,
  "Shares": {
    "Shares": 1000,
    "Level": "normal"
  },
  "Reservation": 0
}
```

## Checklist:

  this project

### DIFF
--- a/govc/vm/disk/change.go
+++ b/govc/vm/disk/change.go
@@ -41,6 +41,9 @@ type change struct {
 
 	bytes units.ByteSize
 	mode  string
+
+	// 	SIOC
+	limit *int64
 }
 
 func init() {
@@ -73,6 +76,7 @@ func (cmd *change) Register(ctx context.Context, f *flag.FlagSet) {
 	f.IntVar(&cmd.key, "disk.key", 0, "Disk unique key")
 	f.StringVar(&cmd.mode, "mode", "", fmt.Sprintf("Disk mode (%s)", strings.Join(vdmTypes, "|")))
 	f.StringVar(&cmd.sharing, "sharing", "", fmt.Sprintf("Sharing (%s)", strings.Join(sharing, "|")))
+	f.Var(flags.NewOptionalInt64(&cmd.limit), "disk.io.limit", "Disk storage IO per seconds limit (-1 for unlimited)")
 }
 
 func (cmd *change) Process(ctx context.Context) error {
@@ -151,6 +155,8 @@ func (cmd *change) Run(ctx context.Context, f *flag.FlagSet) error {
 	if int64(cmd.bytes) != 0 {
 		editdisk.CapacityInKB = int64(cmd.bytes) / 1024
 	}
+
+	editdisk.StorageIOAllocation.Limit = cmd.limit
 
 	switch backing := editdisk.Backing.(type) {
 	case *types.VirtualDiskFlatVer2BackingInfo:
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -5658,6 +5658,7 @@ Examples:
 
 Options:
   -disk.filePath=        Disk file name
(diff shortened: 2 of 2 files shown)


govc/vm/disk/change.go
```diff
@@ -152,6 +156,13 @@ func (cmd *change) Run(ctx context.Context, f *flag.FlagSet) error {
 		editdisk.CapacityInKB = int64(cmd.bytes) / 1024
 	}
 
+	current := editdisk.StorageIOAllocation.Limit
```
embano1, line 159: IIUC the pointer will never be `nil` (either `-1` or the actual custom value)

govc/vm/disk/change.go
```diff
@@ -73,6 +76,7 @@ func (cmd *change) Register(ctx context.Context, f *flag.FlagSet) {
 	f.IntVar(&cmd.key, "disk.key", 0, "Disk unique key")
 	f.StringVar(&cmd.mode, "mode", "", fmt.Sprintf("Disk mode (%s)", strings.Join(vdmTypes, "|")))
 	f.StringVar(&cmd.sharing, "sharing", "", fmt.Sprintf("Sharing (%s)", strings.Join(sharing, "|")))
+	f.Int64Var(&cmd.limit, "disk.io.limit", 0, "Disk storage IO per seconds limit (-1 for unlimited)")
```
embano1, line 79: It's ok to use `0` as default here since this value is not allowed by the API: `A specified parameter was not correct: spec.sharesInfo.limit`

embano1, line 79: It's ok to use `0` as default here since this value is not allowed by the API: `A specified parameter was not correct: spec.sharesInfo.limit`

dougm, line 79: You can use a pointer instead, so the default would actually be `nil`, see for example: https://github.com/vmware/govmomi/blob/20fe92d2f45b3573da5fb830e397a52c5460484d/govc/flags/resource_allocation_info.go#L74

dougm (commented):
`flags.NewOptionalInt64` is probably what you want to use for this flag?

embano1:
Ready for another round @dougm

govc/vm/disk/change.go
```diff
@@ -152,6 +156,13 @@ func (cmd *change) Run(ctx context.Context, f *flag.FlagSet) error {
 		editdisk.CapacityInKB = int64(cmd.bytes) / 1024
 	}
 
+	current := editdisk.StorageIOAllocation.Limit
+	desired := cmd.limit
+
+	if desired != nil && *current != *desired {
```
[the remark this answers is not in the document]
embano1: What if the value is not provided? Wouldn't `cmd.limit` be a `nil` pointer and panic on `*current` deref?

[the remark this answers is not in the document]
embano1: I had my doubts about your initial response, then thought about it and you're right: I don't need a nil check because it won't be sent to the remote in this case. The new code works and requires no nil check :)

dougm (approved):
Nice!

PR #2808, opened 2022-04-06
govc: Add CLI command cluster.module

bryanv:
## Description

Add govc cluster.module command

This is a little easier than trying to remember the vapi HTTP calls. Note that modules.ls does output the MoIDs since that is what the API returns: a later improvement might be to resolve those to their friendlier inventory names.

Closes: #(issue-number)

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

Against real VC, and added bats test

## Checklist:

  this project

### DIFF
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -44,6 +44,11 @@ but appear via `govc $cmd -h`:
  - [cluster.group.create](#clustergroupcreate)
  - [cluster.group.ls](#clustergroupls)
  - [cluster.group.remove](#clustergroupremove)
+ - [cluster.module.create](#clustermodulecreate)
+ - [cluster.module.ls](#clustermodulels)
+ - [cluster.module.rm](#clustermodulerm)
+ - [cluster.module.vm.add](#clustermodulevmadd)
+ - [cluster.module.vm.rm](#clustermodulevmrm)
  - [cluster.override.change](#clusteroverridechange)
  - [cluster.override.info](#clusteroverrideinfo)
  - [cluster.override.remove](#clusteroverrideremove)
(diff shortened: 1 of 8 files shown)


govc/USAGE.md
```diff
@@ -227,6 +227,11 @@ but appear via `govc $cmd -h`:
  - [metric.ls](#metricls)
  - [metric.reset](#metricreset)
  - [metric.sample](#metricsample)
+ - [modules.add_vms](#modulesadd_vms)
```
dougm, line 230: Thinking we should put this under `cluster` and use singular `module`:
```
cluster.module.vm.add
cluster.module.create
cluster.module.ls
cluster.module.rm
```

For `cluster.module.vm`, we could do:
```
cluster.module.vm.add
cluster.module.vm.rm
cluster.module.vm.ls
```
Or even:
```
cluster.module.vm ID -add vm...
cluster.module.vm ID -rm vm...
cluster.module.vm ID # list module VMs
```

Edit: I see that `modules.ls` impl below can list modules or module members, that's fine with me if you prefer   instead of a `module.vm.ls` command.

govc/modules/add_vms.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (cmd *addVMs) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
+	cmd.ClientFlag.Register(ctx, f)
+
+	cmd.SearchFlag, ctx = flags.NewSearchFlag(ctx, flags.SearchVirtualMachines)
+	cmd.SearchFlag.Register(ctx, f)
+
+	f.StringVar(&cmd.moduleID, "module_id", "", "Module ID")
```
dougm, line 47: Since Module ID is required, this could just be the 1st arg to the command(s). As Usage `ID VM...`
Or if using a flag is preferred, let's just name it `-id`.

govc/modules/rm_vms.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/vapi/cluster"
+	"github.com/vmware/govmomi/vim25/mo"
+)
+
+type rmVMs struct {
+	*flags.ClientFlag
```
dougm, line 31: Note that since `SearchFlag` embeds a `ClientFlag`, you can actually remove this field here and the entire `Process` function below. Rest of the code will still compile/work the same.

dougm (commented):
Thanks @bryanv ! Looks great. Just some suggestions, mainly around the package/command names.

govc/modules/add_vms.go
bryanv: Renamed to `-id`

govc/USAGE.md
bryanv: Moved to under cluster

dougm (approved):
lgtm, thanks again @bryanv

Issue #2834, opened 2022-05-05
precision time feature in VM advanced settings in govc


infraguru:
**Is your feature request related to a problem? Please describe.**
Precision Clock since 7.x for VMs command to add virtual component

govc output of the device:
Name:              precisionclock-19000
 Type:            VirtualPrecisionClock
 Label:           Precision Clock 1
 Summary:
 Key:             19000

API call vSphere API Data Object - VirtualPrecisionClock(vim.vm.device.VirtualPrecisionClock)
([link](https://developer.vmware.com/apis/1192/vsphere))

Feature was on the ESXi host level and have the capability to set this feature at host level would complete the full feature request.

**Additional context**
feature should include both options to set, host system (NTP) or (ANY)

PR #2837, opened 2022-05-05
govc: add 'device.clock.add' command

dougm:
## Description

add 'device.clock.add' command - Add precision clock device to VM

Closes: #2834

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

Tested against real vCenter + vcsim

## Checklist:

  this project

### DIFF
--- a/govc/device/clock/add.go
+++ b/govc/device/clock/add.go
@@ -1,0 +1,86 @@
+/*
+Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package floppy
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+type add struct {
+	*flags.VirtualMachineFlag
+}
+
+func init() {
+	cli.Register("device.clock.add", &add{})
+}
+
+func (cmd *add) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.VirtualMachineFlag, ctx = flags.NewVirtualMachineFlag(ctx)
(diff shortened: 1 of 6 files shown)


govc/device/clock/add.go
```diff
@@ -0,0 +1,91 @@
...
+  govc device.info clock-*`
+}
+
+func (cmd *add) Run(ctx context.Context, f *flag.FlagSet) error {
+	vm, err := cmd.VirtualMachine()
+	if err != nil {
+		return err
+	}
+
+	if vm == nil {
```
embano1, line 56: question: is this check needed, i.e. there is a case where err==nil but no vm returned?

govc/device/clock/add.go
```diff
@@ -0,0 +1,91 @@
...
+		return flag.ErrHelp
+	}
+
+	devices, err := vm.Device(ctx)
```
embano1, line 60: question: is there a reason why you have this call here? I don't see the value used (overwritten at later stage)

govc/device/clock/add.go
```diff
@@ -0,0 +1,91 @@
...
+	devices, err = vm.Device(ctx)
+	if err != nil {
+		return err
+	}
+
+	devices = devices.SelectByType(d)
+
+	name := devices.Name(devices[len(devices)-1])
+
+	fmt.Println(name)
```
embano1, line 88: nit: more user-friendly print?

govc/device/clock/add.go
dougm: Yes, see: https://github.com/vmware/govmomi/blob/be5528b51ec58f28601350e466e662727cf6c7c4/govc/flags/virtual_machine.go#L102

Although, I'm not sure why we did that instead of returning `ErrHelp`. There might have been other cases in the past where a vm flag was optional, but I only see 1 here: https://github.com/vmware/govmomi/blob/be5528b51ec58f28601350e466e662727cf6c7c4/govc/vm/option/info.go#L103

Something we could refactor in another PR.

govc/device/clock/add.go
dougm: I'll remove this, copy-n-paste from another `device/*/add.go` command, where we use `devices` to find the controller.  Example: https://github.com/vmware/govmomi/blob/be5528b51ec58f28601350e466e662727cf6c7c4/govc/device/cdrom/add.go#L72-L77

govc/device/clock/add.go
dougm: We do the same in other `device/*/add.go` commands, so you can use the `device.*.add` command output as input to other commands.  Example: https://github.com/vmware/govmomi/blob/be5528b51ec58f28601350e466e662727cf6c7c4/toolbox/toolbox-test.sh#L102-L103

embano1: ah, didn't know it was for piping :)

embano1 (approved):
LGTM

embano1 (approved):
LGTM

Issue #2843, opened 2022-05-12
Introduce VmConfigSpec field to the DeployLibraryItem() DeploymentSpec 


sreyasn:
The VmConfigSpec field was recently introduced into the DeploymentSpec of the content library Deploy Item API to facilitate specifying a vim.vm.ConfigSpec in a serialized base64 encoded XML format. 

This will allow LibraryItem's deployed to take up vm specifications from the config spec rather than defaults from the OVF.

PR #2844, opened 2022-05-13
Add content library VmConfigSpec support

dougm:
## Description

- govc: add library.deploy '-config' flag

- vcsim: add content library VmConfigSpec support

- api: add VmConfigSpec field to content library DeploymentSpec

Closes: #2843

## Type of change

  not work as expected)
- [ ] Build related change

## Checklist:

  this project

### DIFF
--- a/govc/library/deploy.go
+++ b/govc/library/deploy.go
@@ -38,6 +38,7 @@ type deploy struct {
 	*importx.OptionsFlag
 
 	profile string
+	config  string
 }
 
 func init() {
@@ -61,6 +62,10 @@ func (cmd *deploy) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.OptionsFlag.Register(ctx, f)
 
 	f.StringVar(&cmd.profile, "profile", "", "Storage profile")
+
+	if cli.ShowUnreleased() {
+		f.StringVar(&cmd.config, "config", "", "VM config spec")
+	}
 }
 
 func (cmd *deploy) Process(ctx context.Context) error {
@@ -220,6 +225,14 @@ func (cmd *deploy) Run(ctx context.Context, f *flag.FlagSet) error {
 				FolderID:       folder.Reference().Value,
 			},
 		}
+
+		if cmd.config != "" {
+			deploy.VmConfigSpec = &vcenter.VmConfigSpec{
+				Provider: "XML",
+				XML:      cmd.config,
+			}
+		}
+
 		ref, err = m.DeployLibraryItem(ctx, item.ID, deploy)
 		if err != nil {
 			return err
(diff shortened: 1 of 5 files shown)


vapi/simulator/simulator.go
```diff
@@ -1898,6 +1938,17 @@ func (s *handler) deleteVM(ref *types.ManagedObjectReference) {
 	})
 }
 
+func (s *handler) reconfigVM(ref types.ManagedObjectReference, config types.VirtualMachineConfigSpec) error {
```
embano1, line 1941: nit: `reconfigureVM` better?

govc/library/deploy.go
```diff
@@ -61,6 +63,10 @@ func (cmd *deploy) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.OptionsFlag.Register(ctx, f)
 
 	f.StringVar(&cmd.profile, "profile", "", "Storage profile")
+
+	if os.Getenv("VM_CLASS_AS_CONFIG") == "true" { // FSS
```
embano1, line 67: Question: why do we conditionally enable (define) this flag based on this env var? Is this a pattern we already do? Would it harm to always define it, since it's `""` by default?

Also: what does the comment `// FFS` stand for?

sreyasn (dismissed):
Change LGTM, Doug. Thanks again!.

govc/library/deploy.go
dougm: Feature State Switch. We use this pattern in vCenter/WCP/VM Service/Guest Cluster Service, but this is the first time in govc it'd be used for a specific flag/field.  govc does have a way to disable unreleased commands by default (see below). We'd used that for library.{checkin,checkout} for example.  This feature will be released with 8.0, but is still subject to change and could still be disabled by default?  Just wanted to add the flag for testing within govmomi itself and folks working on the feature.

https://github.com/vmware/govmomi/blob/18cdce89aae64488c0f3c76d113c3ff7c076cbb0/govc/cli/register.go#L25-L27

vapi/simulator/simulator.go
dougm: I think it's ok - unexported and the API method is actually called `ReconfigVM_Task`.. not sure why chose to spell it out in the wrapper method :shrug:

dougm: I think it's ok - unexported and the API method is actually called `ReconfigVM_Task`.. not sure why chose to spell it out in the wrapper method :shrug:

govc/library/deploy.go
dougm: On second thought, I suppose we could just use the same `GOVC_SHOW_UNRELEASED` env var here instead adding a feature specific env var.

dougm: Pushed a change that does the above, can you review again @sreyasn @embano1 ?

PR #2901, opened 2022-07-15
Ensured namespace-management works on 7.0u2+ with wider configuration options

adamfowleruk:
Added support for vDS networking and the Avi Load Balancer.
Tested on 7.0u2 and 7.0u3.
Added strongly typed fields for enums that were previously freeform strings.
Needed as part of the new terraform-provider-namespace-management project located here (not public yet): https://github.com/vmware-tanzu-labs/terraform-provider-namespace-management

### DIFF
--- a/govc/namespace/cluster/enable.go
+++ b/govc/namespace/cluster/enable.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+Copyright (c) 2020-2022 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


dougm (approved):
Awesome, lgtm @adamfowleruk . And extra thanks for the docs and notes on version diffs. Will leave this open for a bit to give others a chance to review, then we'll get it merged.

dougm:
@adamfowleruk can you squash the commits and re-include the `api:` prefix in the summary and add `Closes: #issue-num` 
These will improve the [changelog generator](https://github.com/vmware/govmomi/blob/master/CONTRIBUTING.md#format-of-the-commit-message)

adamfowleruk:
OK thanks all. I'll look at merging the commit later today/tomorrow. I'm not too worried about the govc issues today as the current implementation won't work on a real system anyway on 7.0u2+. Any thoughts on necessity of keeping govc up to date with the API would be welcome, given it only tests against the simulator (whose code is written so the govc/api passes anyway... so not a real test per-se.)

adamfowleruk:
Turns out I cannot Squash all my commits because I've been merging in changes to vmware/master whilst developing this functionality, and I cannot squash unless they're only my changes and contiguous in order. Only the final commit of mine has the changelog API compatible statements so should be fine as-is. Any thoughts?

embano1:
@adamfowleruk instead of squashing you can also use the `git reset` way to soft-reset your changes back and create a proper final commit. See here for an example: https://www.mgasch.com/2021/05/git-basics/#option-2-git-reset

adamfowleruk:
I've performed the soft reset. Please advise if it is correctly done.

dougm (approved):
lgtm, thanks @adamfowleruk !

Issue #2906, opened 2022-07-19
[BUG] vcsim: OptionsManager is racy


brakthehack:
**Describe the bug**
`OptionsManager` holds several data structures that can be updated by different threads. Since vcsim uses a single `OptionsManager` per host, it can race when two threads are updating it at the same time.

**To Reproduce**

1. Instantiate vcsim
2. Have two threads race to update `OptionsManager` Settings on the same host.
3. Eventually, observe the race using your favorite tooling.

**Expected behavior**
No races to update the same data structures in `OptionsManager`.

**Affected version**
0.27.4

**Screenshots/Debug Output**
```
WARNING: DATA RACE
Read at 0x00c00059c280 by goroutine 72:
  github.com/vmware/govmomi/simulator.(*OptionManager).QueryOptions()
      /home/worker/workspace/pkg/mod/github.com/vmware/govmomi@v0.27.4/simulator/option_manager.go:58 +0x450
  runtime.call16()
      /home/worker/workspace/linux64/src/runtime/asm_amd64.s:625 +0x48
  reflect.Value.Call()
      /home/worker/workspace/linux64/src/reflect/value.go:339 +0xd7
  github.com/vmware/govmomi/simulator.(*Service).call.func1()
      /home/worker/workspace/pkg/mod/github.com/vmware/govmomi@v0.27.4/simulator/simulator.go:218 +0x94
  github.com/vmware/govmomi/simulator.(*Registry).WithLock()
      /home/worker/workspace/pkg/mod/github.com/vmware/govmomi@v0.27.4/simulator/registry.go:625 +0x5e
  github.com/vmware/govmomi/simulator.(*Service).call()
      /home/worker/workspace/pkg/mod/github.com/vmware/govmomi@v0.27.4/simulator/simulator.go:217 +0x1acd
  github.com/vmware/govmomi/simulator.(*Service).ServeSDK()
      /home/worker/workspace/pkg/mod/github.com/vmware/govmomi@v0.27.4/simulator/simulator.go:492 +0xd0d
  github.com/vmware/govmomi/simulator.(*Service).ServeSDK-fm()
      /home/worker/workspace/pkg/mod/github.com/vmware/govmomi@v0.27.4/simulator/simulator.go:446 +0x57
  net/http.HandlerFunc.ServeHTTP()
      /home/worker/workspace/linux64/src/net/http/server.go:2047 +0x4d
  net/http.(*ServeMux).ServeHTTP()
[...]

PR #2907, opened 2022-07-19
vcsim: Synchronize calls to OptionsManager internal data structures

brakthehack:
## Description

As described in #2906, `OptionsManager` is racy as calls to its internal
data structures are unsynchronized. When multiple threads are making
calls to `OptionsManager`, they can race against each other.

This change protects these data structures with a mutex, to avoid the
race conditions.

Closes: #2906

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

Wrote a simple test and ran it with the `-race` flag. I consistently hit race conditions without my change and none with my change:

Snip:
```
	vcsim := vcsimulator.NewSimulator()
	defer vcsim.Stop()

	vcsim2 := vcsimulator.NewSimulator()
	defer vcsim2.Stop()

	t.Run("test", func(t *testing.T) {
		client := vcsim.Client.Client
		client2 := vcsim2.Client.Client
		m := object.NewOptionManager(client, client.ServiceContent.Setting.Reference())
		m2 := object.NewOptionManager(client2, client2.ServiceContent.Setting.Reference())
		ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
		defer cancel()
		for {
			select {
			case <-ctx.Done():
				return
			default:
			}
			// adding the new options to the simulated VC
			opt := fmt.Sprintf("config.%d", rand.Int())
			go m.Update(ctx, []vimtypes.BaseOptionValue{
				&vimtypes.OptionValue{Key: opt, Value: "bar"},
			})

			// adding the new options to the simulated VC
			opt2 := fmt.Sprintf("config.%d", rand.Int())
			go m2.Update(ctx, []vimtypes.BaseOptionValue{
				&vimtypes.OptionValue{Key: opt2, Value: "bar"},
			})

			m.Query(ctx, opt)
		}
```

## Checklist:

  [guidelines](https://github.com/vmware/govmomi/blob/master/CONTRIBUTING.md) of
  this project

### DIFF
--- a/simulator/option_manager.go
+++ b/simulator/option_manager.go
@@ -18,6 +18,7 @@ package simulator
 
 import (
 	"strings"
+	"sync"
 
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/simulator/esx"
@@ -30,6 +31,7 @@ import (
 
 type OptionManager struct {
 	mo.OptionManager
+	lock sync.Mutex
 }
 
 func NewOptionManager(ref *types.ManagedObjectReference, setting []types.BaseOptionValue) object.Reference {
@@ -42,6 +44,9 @@ func NewOptionManager(ref *types.ManagedObjectReference, setting []types.BaseOpt
 }
 
 func (m *OptionManager) init(r *Registry) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
 	if len(m.Setting) == 0 {
 		if r.IsVPX() {
 			m.Setting = vpx.Setting
@@ -54,6 +59,9 @@ func (m *OptionManager) init(r *Registry) {
 func (m *OptionManager) QueryOptions(req *types.QueryOptions) soap.HasFault {
 	body := &methods.QueryOptionsBody{}
 	res := &types.QueryOptionsResponse{}
+
+	m.lock.Lock()
+	defer m.lock.Unlock()
 
 	for _, opt := range m.Setting {
 		if strings.HasPrefix(opt.GetOptionValue().Key, req.Name) {
@@ -70,6 +78,7 @@ func (m *OptionManager) QueryOptions(req *types.QueryOptions) soap.HasFault {
 	return body
 }
 
+// find requires holding the lock.
 func (m *OptionManager) find(key string) *types.OptionValue {
 	for _, opt := range m.Setting {
 		setting := opt.GetOptionValue()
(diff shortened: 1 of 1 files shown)


dougm (commented):
Thanks @brakthehack .  Each vcsim ManagedObject already has its own lock by default, where the lock is held during any API invoked method call here: https://github.com/vmware/govmomi/blob/d8bd5f625fd9bd47d72159a024547541bf903e97/simulator/simulator.go#L217-L219

For example, the `m.Call()` above would be a method such as `UpdateOptions`. And `WithLock()` does a Mutex `Lock()` + `Unlock()` around the call.

I noticed the test in your issue creates 2 vcsim instances:
```
        vcsim := vcsimulator.NewSimulator()
	defer vcsim.Stop()

	vcsim2 := vcsimulator.NewSimulator()
	defer vcsim2.Stop()
```

Are you able to reproduce the issue with a single vcsim instance? If not, I wonder if we actually need to deep copy here: https://github.com/vmware/govmomi/blob/d8bd5f625fd9bd47d72159a024547541bf903e97/simulator/option_manager.go#L47

As the 2 vcsim instances my actually be racing on `vpx.Setting` ?  I will take a closer look.

simulator/option_manager.go
```diff
@@ -70,6 +78,7 @@ func (m *OptionManager) QueryOptions(req *types.QueryOptions) soap.HasFault {
 	return body
 }
 
+// find requires holding the lock.
```
akutz, line 81: Either:

* Update the GoDoc to be more clear:
    ```go
    // find is not thread-safe and requires callers to already hold either
    // a shared or exclusive lock from the OptionsManager
    ```

**or**

* Move the function into `UpdateOptions` so only it can invoke `find`, ex:
    ```go
    func (m *OptionManager) UpdateOptions(req *types.UpdateOptions) soap.HasFault {
    	m.Lock()
    	defer m.Unlock()
    
    	find := func(key string) *types.OptionValue {
    		for _, opt := range m.Setting {
    			setting := opt.GetOptionValue()
    			if setting.Key == key {
    				return setting
    			}
    		}
    		return nil
    	}
    
    	// ...
    ```

akutz (changes requested):
Thanks for the PR @brakthehack!

akutz:
For some reason @dougm 's comments didn't show up when I read this PR. GH must have been brain-farting, because I swore this PR showed up as "new."

Anyway, Doug is correct in that each MO has its own lock domain, so my suggestions should not actually be necessary. Still, they're worth considering as general guidelines for handling concurrency.

dougm:
...
> I noticed the test in your issue creates 2 vcsim instances:
> 
> ```
>         vcsim := vcsimulator.NewSimulator()
> 	defer vcsim.Stop()
> 
> 	vcsim2 := vcsimulator.NewSimulator()
> 	defer vcsim2.Stop()
> ```
...
> As the 2 vcsim instances my actually be racing on `vpx.Setting` ? I will take a closer look.

Closing this as https://github.com/vmware/govmomi/pull/3701 provides support for > 1 parallel vcsim instance in the same process.

PR #2921, opened 2022-07-31
toolbox: add hgfs OpenBSD stub

geowiwi:
Fix build on OpenBSD, add hgfs stub.

Basically I need this for telegraf, so this is untested with VMware vSphere APIs.

### DIFF
--- a/toolbox/hgfs/hgfs_openbsd.go
+++ b/toolbox/hgfs/hgfs_openbsd.go
@@ -1,0 +1,24 @@
+/*
+Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


embano1:
Should we also add a build constraint? https://pkg.go.dev/cmd/go#hdr-Build_constraints

geowiwi:
Does the boilerplate test needs fixing, or is the license header really required to look this way?

embano1:
> Does the boilerplate test needs fixing, or is the license header really required to look this way?

Yeah. Sorry for that. It requires a range, see [here](https://github.com/vmware/govmomi/blob/master/hack/boilerplate/boilerplate.go.txt) for the default. If no range is specified, it will use <current-year>-<year>.

Feel free to use the one provided in the baseline example.

geowiwi:
The failing/cancelled govc tests look unrelated to me.

dougm (approved):
lgtm, thanks @geowiwi

Issue #3004, opened 2022-12-05
[BUG] govc sso.user.ls lists only first 100 records


lapawa:
**Describe the bug**
When calling '# govc sso.user.ls' you'll receive the description of 100 user account only.
There is no way to let govc collect the next bunch of user accounts if there are more available.

**To Reproduce**

1. Prepare govc cli to login to your vCenter Server instance. E.g. by setting these environment variables 
GOVC_URL, GOVC_USERNAME and GOVC_PASSWORD

3.  Create 101 sso local user accounts. E.g. with this bash snipped:
`#
for ID in $(seq -w 101); do 
  USER="test-issue-user-$ID";
  echo "creating user '$USER'";
  govc sso.user.create -p 'TopSecret123!' "$USER";
done
`

4. Verify the complete list of users in the vSphere client.
  Administration->Single Sign On->Users and Groups
  Number of local users is shown below the right table.

5. Let govc list the sso users and count the lines of the result
`#
govc sso.user.ls | wc -l`
`100`

**Expected behavior**
 Step 4 should count all users. Not just 100.
On a fresh vCenter Server installation this might be something around 110 users.

**Affected version**
`# govc version`
`govc 0.29.0`

HakanSunay:
The ssoadmin client has `ssoadmin.Client#Limit`, which is a positive integer for the maximum number of items to return.

By default, that value is set to **100**:
https://github.com/vmware/govmomi/blob/d8e3357f206f310cba05748e23587a094f22ee71/ssoadmin/client.go#L85

lapawa:
Pull #3011 still does not fix the issue.
Using govc 0.30.1 is also limited to 100 sso objects.
When dumping the traffic by:
`govc sso.user.ls  -trace`
I can find the request with a limit parameter of 100.
`<limit>100</limit></FindPersonUsers>`
But I can't explain where this limit comes from.

dougm:
> Using govc 0.30.1 is also limited to 100 sso objects.

The fix was only on the main branch. I just cherry picked it to the release-0.30 branch and created a [0.32 release](https://github.com/vmware/govmomi/releases/tag/v0.30.2). Can you give that a try?

lapawa:
This one works for me. It reports back all available 191 users
and the request shows the new limit:
`<limit>2147483647</limit></FindPersonUsers></Body>`
Thank you

PR #3011, opened 2022-12-14
fix: Pin the result limit of ssoadmin client to its MAX

HakanSunay:
## Description

With this change, we are modifying the default result limit of the ssoadmin client from 100 to 2 147 483 647 (`math.MaxInt32`).
This is the path of least resistance to fixing the issue mentioned in #3004, another possible way is to take the value as a user input through an environment variable or flag, but that might result in API changes.

Closes: #3004 

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [x] Verified that the example in #3004 works: 
```
# 3004.sh

export GOVC_URL=<redacted>
export GOVC_USERNAME=<redacted>
export GOVC_PASSWORD=<redacted>
export GOVC_INSECURE=1

for ID in $(seq -w 101); do 
  USER="test-issue-user-$ID";
  echo "creating user '$USER'";
  go run govc/main.go sso.user.create -p 'TopSecret123!' "$USER";
done

go run govc/main.go sso.user.ls | wc -l
```

Result:
```
❯ ./3004.sh
creating user 'test-issue-user-001'
...
creating user 'test-issue-user-101'
     105
```

## Checklist:

  [guidelines](https://github.com/vmware/govmomi/blob/master/CONTRIBUTING.md) of
  this project

### DIFF
--- a/ssoadmin/client.go
+++ b/ssoadmin/client.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+Copyright (c) 2018-2022 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package ssoadmin
 
 import (
 	"context"
+	"math"
 	"path"
 	"reflect"
 	"strings"
@@ -82,7 +83,7 @@ func NewClient(ctx context.Context, c *vim25.Client) (*Client, error) {
 		Client:       sc,
 		RoundTripper: sc,
 		Domain:       "vsphere.local", // Default
-		Limit:        100,
+		Limit:        math.MaxInt32,
 	}
 	if url != Path {
 		admin.Domain = path.Base(url)


dougm (commented):
Thanks @HakanSunay . I don't recall why we chose 100 as the default limit, but iirc it needs to be something other than '0' to get any results.  I had started to look at this before your PR and had the patch below in mind, as I found tests/apps/etc doing the same in our source tree:
```diff
modified   ssoadmin/client.go
@@ -18,6 +18,7 @@ package ssoadmin
 
 import (
 	"context"
+	"math"
 	"path"
 	"reflect"
 	"strings"
@@ -82,7 +83,7 @@ func NewClient(ctx context.Context, c *vim25.Client) (*Client, error) {
 		Client:       sc,
 		RoundTripper: sc,
 		Domain:       "vsphere.local", // Default
-		Limit:        100,
+		Limit:        math.MaxInt32,
 	}
 	if url != Path {
 		admin.Domain = path.Base(url)
```

I'd also be ok with having it be configurable, but don't want to change the sso method signatures. The govc commands can just set the `ssoadmin.Client.Limit` as above. Would also prefer not to add a flag to the sso commands. This case could just be a set-and-forget env var, such as:
```console
$ export GOVC_SSOADMIN_LIMIT=20000
$ govc sso.group.ls
$ ...
```
What do you think?

HakanSunay:
> Thanks @HakanSunay . I don't recall why we chose 100 as the default limit, but iirc it needs to be something other than '0' to get any results. I had started to look at this before your PR and had the patch below in mind, as I found tests/apps/etc doing the same in our source tree:
> 
> ```diff
> modified   ssoadmin/client.go
> @@ -18,6 +18,7 @@ package ssoadmin
>  
>  import (
>  	"context"
> +	"math"
>  	"path"
>  	"reflect"
>  	"strings"
> @@ -82,7 +83,7 @@ func NewClient(ctx context.Context, c *vim25.Client) (*Client, error) {
>  		Client:       sc,
>  		RoundTripper: sc,
>  		Domain:       "vsphere.local", // Default
> -		Limit:        100,
> +		Limit:        math.MaxInt32,
>  	}
>  	if url != Path {
>  		admin.Domain = path.Base(url)
> ```
> 
> I'd also be ok with having it be configurable, but don't want to change the sso method signatures. The govc commands can just set the `ssoadmin.Client.Limit` as above. Would also prefer not to add a flag to the sso commands. This case could just be a set-and-forget env var, such as:
> 
> ```
> $ export GOVC_SSOADMIN_LIMIT=20000
> $ govc sso.group.ls
> $ ...
> ```
> 
> What do you think?

I prefer the path of least resistance and change. I will modify the PR to use `math.MaxInt32`.

dougm (approved):
Great, thanks @HakanSunay !

PR #3032, opened 2023-01-20
chore: bats related fixes

dougm:
## Description

- Commit 1991de51 recently changed the json tag for ArrayOf types

- Fix 'vcsim run container' test when using MacOS / Docker desktop

- Change vcsim/main.go to open the fifo after its running

The vcsim/main.go change may fix the random hangs we see in GH Actions. I was seeing random hangs locally when running all bats tests on MacOS.  Each time, the bats tests would be hung reading from the fifo here: https://github.com/vmware/govmomi/blob/99336079d7934f742f729c3d74ebdf8e4bdd6f4a/govc/test/test_helper.bash#L59

Not certain this will fix the random hangs in GH Actions, but there's no reason to open the fifo for writing until we have the data that'll be written to it anyhow.

### DIFF
--- a/govc/test/vcsim.bats
+++ b/govc/test/vcsim.bats
@@ -128,7 +128,7 @@ EOF
   run govc object.collect -s $vm summary.guest.ipAddress
   assert_success "10.0.0.1"
 
-  netip=$(govc object.collect -json -s $vm guest.net | jq -r .[].val.GuestNicInfo[].ipAddress[0])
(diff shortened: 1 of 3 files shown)


akutz (approved):
lgtm, thanks. Sorry for breaking this. We _really_ need to fix that weird, only-on-GH Actions issue with tests @dougm . It's way too easy to ignore those due to that.

Issue #3037, opened 2023-01-25
Example for SSO client based applications


jashanj0tsingh:
I want to create a rather lite dashboard that will leverage govmomi wrapper backend to interact with the vCenter server, the issue is that when I launch the URL for a VMs web console it asks me for credentials, I would like to reuse the credentials that client was created with and since the action is triggered by the client the authentication should be already taken care of in the first place.

This is what I have so far...

```
func NewSSOClient(ctx context.Context, client *vim25.Client) []byte {

	vSphereUri, err := GetConfiguration()
	LogError(err)

	sc := soap.NewClient(&vSphereUri, true)
	vc, err := vim25.NewClient(ctx, sc)
	LogError(err)

        stsc, err := sts.NewClient(ctx, vc)
	LogError(err)

	return []byte(stsc.URL().RawPath)
}
```

kmahyyg:
The SSO client related examples can be found on `govc/sso/` folder.

kmahyyg:
Comments edited, and updated. Please check on the web page.

Issue #3041, opened 2023-01-31
Add support for vCenter login via OAuth2 token


jhg03a:
**Is your feature request related to a problem? Please describe.**
We leverage SSO integration via OAuth2 ADFS OIDC.  I'm unable to leverage this authentication source with GOVC and service accounts that have a client id/secret.

**Describe the solution you'd like**
I'd like to see additional login options, similar to `-cert` and `-key`, to supply OAuth client_id/secret fields when creating a session login.

**Describe alternatives you've considered**
Instead create local vSphere default auth domain accounts.  This has the drawback that you're now managing users in multiple places and the accounting/auditing aspects now fall on vSphere instead of the SSO backend.  This also presents the challenges of manual/automated intervention for account creation/deletion/credential rotation.

**Additional context**
[vCenter Rest API](https://developer.vmware.com/apis/vsphere-automation/latest/vcenter/api/vcenter/authentication/token/post/)
[Powershell example](https://vdc-download.vmware.com/vmwb-repository/dcr-public/72479a44-60ed-4b00-839a-0e442d85a863/6a9f9b49-367a-4ece-b543-ac60cb45f9e9/doc/GUID-46FE9A79-1294-4480-A9D5-24CA35303329.html)

dougm:
We can look at adding this as an option to the [session.login](https://github.com/vmware/govmomi/blob/main/govc/USAGE.md#sessionlogin) command

I've not used the APIs for Oauth2, but might be able to get some help on this.

```console
% token=$(govc session.login -oauth2 -issue)
% govc session.login -u host -oauth2 -token $token
````

karaatanassov:
The flow of authentication with OAuth 2 is described here:

https://developer.vmware.com/docs/16116/vmware-vsphere-automation-rest-api-programming-guide-8-0-update-1/GUID-1A1D3802-2484-4C9C-B3A3-129C26E36E82.html

The gist is:

1. Obtain ID and access token from ADFS using some of the standard grant type - authorization, password, client credentials, refresh token.
2. Exchange the ID and access tokens for SAML token at: /api/vcenter/authentication/token
3.  Use the SAML token with SessionManager.LoginByToken() and /api/session APIs to obtain REST and SOAP sessions respectively (Note that the JSON protocol in 8.0u1 does not support SAML i.e. LoginByToken will only work with SOAP).

For govc user experience this probably requires 2 things:

[...]

PR #3047, opened 2023-02-09
examples: add session.Manager.LoginByToken

dougm:
Issue #3037

### DIFF
--- a/session/example_test.go
+++ b/session/example_test.go
@@ -1,0 +1,74 @@
+/*
+Copyright (c) 2023-2023 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package session_test
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	_ "github.com/vmware/govmomi/lookup/simulator"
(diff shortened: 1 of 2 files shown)


session/example_test.go
```diff
@@ -0,0 +1,74 @@
+/*
+Copyright (c) 2023 VMware, Inc. All Rights Reserved.
```
karaatanassov, line 2: The linter seems to want 2 year values like 

"Copyright (c) 2023-2023 VMware, Inc. All Rights Reserved."

karaatanassov (dismissed):
Thanks for the sample. It works! Nice that we have this.

Few questions and ideas not directly related to this review.

The REST API has some convoluted way of using tokens i.e. "Authorization: SIGN <saml_token>" header. Do you know if we have code for this?

Using external IDP to issue JWT, then token exchange to get SAML (vCenter scoped token) and last creating session seems to be the direction vCenter is going. 

Here is the token exchange REST API
https://developer.vmware.com/apis/vsphere-automation/latest/vcenter/api/vcenter/authentication/token/post/

We can discuss how to illustrate the 3 step process. One problem is that "external IdP" could be one of many - MS ADFS, Okta, CSP.

dougm:
> The REST API has some convoluted way of using tokens i.e. "Authorization: SIGN <saml_token>" header. Do you know if we have code for this?

We do and I've added an example to this PR. Client code is here: https://github.com/vmware/govmomi/blob/6687830863b6607134e359991c48fd758fc8f9ed/sts/signer.go#L241-L242

> Using external IDP to issue JWT, then token exchange to get SAML (vCenter scoped token) and last creating session seems to be the direction vCenter is going.
> 
> Here is the token exchange REST API https://developer.vmware.com/apis/vsphere-automation/latest/vcenter/api/vcenter/authentication/token/post/

Yes, this was requested in #3041 , I had started looking into it.. I believe WCP uses govmomi to do this. We should look at moving that code from WCP to upstream govmomi so it can be reused.  I recall speaking with @mikekim about this a while back, he might be able to help with the upstreaming effort.

Thanks Kiril!

karaatanassov (approved):

PR #3093, opened 2023-04-13
chore: Err on json.Decode for invalid type/vals

akutz:
## Description

This patch updates the behavior of the JSON codec package so that if the decode operation cannot convert a value into the destination type, the error is no longer ignored silently, but returned so the caller can handle it.

This patch also adds tests to validate these cases.

Please note this _is_ potentially a breaking change since previously, if `{"_typeName":"string", "_value":true}` was decoded, the result would have been an empty string. Now the result is an error returned from the decode operation that would read `json: cannot unmarshal bool into Go value of type string`.

Closes: `NA`

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

Several unit tests have been added to validate the test cases.

- [x] Invalid type/name value tests in `discriminator_test.go`
- [x] Invalid ExtraConfig type/name value in `json_test.go`

## Checklist:

  this project

### DIFF
--- a/vim25/json/discriminator.go
+++ b/vim25/json/discriminator.go
@@ -266,6 +266,19 @@ func (d *decodeState) discriminatorGetValue() (reflect.Value, error) {
 		return reflect.Value{}, err
 	}
 
+	// Check the saved error as well since the decoder.value function does not
+	// always return an error. If the reflected value is still zero, then it is
+	// likely the decoder was unable to decode the value.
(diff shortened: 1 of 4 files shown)


karaatanassov (approved):
Wow that is a good catch! This fix you make seems will handle all sort of such cases right? e.g. string instead of number and vice versa?

PR #3101, opened 2023-04-17
fix: include StoragePod Datastore children in ContainerView

### LINKED ISSUE #3094: [BUG] vcsim: viewmanager doesn't return all Datastore types when parent is StoragePod
**Describe the bug**
When using a ContainerView.Retrieve() method, vcsim doesn't return the Datastore managed entities when it's parent is of a StoragePod type.

**To Reproduce**

1. with golang: Create a ContainerView
1. with golang: Execute method ```Retrieve()```
1. When executing to the vCenter: returns all datastore entities
1. When executing to vcsim with a dump (govc save): doesn't return the datastore entities which are in a datastore cluster (storagepod)

**Expected behavior**
I expect all datastore managed entities, also the ones connected to a storagepod (datastore cluster)

**Affected version**
Please provide details on the version used, e.g. release tag, commit, module version, etc.
Using latest version:
```
Build Version: 0.30.4
Build Commit: 4561966a
Build Date: 2023-03-16T18:15:54Z
```

**Screenshots/Debug Output**
Expected (this output is from vCenter):
image
Getting (this output is from vcsim):
image

**Additional context**
Using get-datastore cmdlet does return the datastores, so the objects are loaded in vcsim, but the viewmanager doesn't return them like a vCenter system does. Also checked with 'govc find -l' doesn't return the datastores which parent is a storagepod.
Output from get-datastore:
image

dougm:
Closes #3094

### DIFF
--- a/govc/test/disk.bats
+++ b/govc/test/disk.bats
@@ -42,8 +42,16 @@ load test_helper
   run govc disk.create -datastore-cluster $pod "$id"
   assert_failure
 
+  run govc find $pod -type s
+  assert_success
+  [ ${#lines[@]} -eq 0 ]
+
   run govc object.mv /DC0/datastore/LocalDS_{1,2} $pod
   assert_success
+
+  run govc find $pod -type s
+  assert_success
+  [ ${#lines[@]} -eq 2 ]
 
   run govc disk.create -datastore-cluster $pod -size 10M "$id"
   assert_success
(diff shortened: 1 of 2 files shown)


hickeng (approved):
StoragePod is the only thing that inherits from Folder currently. If you can think of a neat way to add things-that-implement-folder-interface to the switch it may be worthwhile for future proofing, even if the action is to panic so we're aware of a Folder sub-type that's not handled.

Issue #3127, opened 2023-05-16
[BUG] vim25 JSON does not support dateTime and binary in OptionValue


karaatanassov:
**Describe the bug**
The json serialization code for vim25 does not support `binary` and `dateTime` where "any" is used. For example this is the `OptionValue` in VMs

**Expected behavior**
dateTime nad binary should be supported like following in JSON 

```
{
    "_typeName": "OptionValue",
    "key": "option1",
    "value": {
        "_typeName": "dateTime",
        "_value": "2022-12-12T11:48:35.473645Z"
    }
}
{
    "_typeName": "OptionValue",
    "key": "option1",
    "value": {
        "_typeName": "binary",
        "_value": "SGVsbG8="
    }
}
```
**Affected version**
v.0.30.4

**Screenshots/Debug Output**
These types are missing in code https://github.com/vmware/govmomi/blob/main/vim25/types/json_encoding.go#L32

PR #3128, opened 2023-05-16
fix: support `binary` and `dateTime` in JSON

karaatanassov:
This change adds support for `binary` and `dateTime` in JSON.

For `binary` golang `[]byte` JSON serialization matches the vCenter format. Thus adding the name ot the maps was enough.

For `dateTime` the format of `time.Time` `Marshaler` and `Unmarshaler` matches vCenter format. The discriminator code required support for custom marshaling logic. In addition the name had to be mapped.

## Description

Please include a summary of the change and which issue is fixed. Please also
include relevant motivation and context. List any dependencies that are required
for this change.

Closes: #3127

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [X] Unit tests

## Checklist:

- [X] My code follows the `CONTRIBUTION`
  this project

### DIFF
--- a/vim25/json/discriminator.go
+++ b/vim25/json/discriminator.go
@@ -184,14 +184,17 @@ func (d *decodeState) discriminatorGetValue() (reflect.Value, error) {
 			// Assign the type instance to the outer variable, t.
 			t = ti
 
-			switch t.Kind() {
-			case reflect.Map, reflect.Struct:
-				// If the type is a map or a struct then it is not necessary to
-				// continue walking over the current JSON object since it will be
-				// completely rescanned to decode its value into the discovered
-				// type.
+			// Primitive types and types with Unmarshaler are wrapped in a
+			// structure with type and value fields. Structures and Maps not
+			// implementing Unmarshaler use discriminator embedded within their
+			// content.
+			if useNestedDiscriminator(t) {
+				// If the type is a map or a struct not implementing Unmarshaler
+				// then it is not necessary to continue walking over the current
+				// JSON object since it will be completely re-scanned to decode
+				// its value into the discovered type.
 				dd.opcode = scanEndObject
-			default:
+			} else {
 				// Otherwise if the value offset has been discovered then it is
 				// safe to stop walking over the current JSON object as well.
 				if valueOff > -1 {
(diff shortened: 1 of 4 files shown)


dekp (commented):
Looks good. Thanks.

vim25/json/discriminator.go
```diff
@@ -184,19 +184,28 @@ func (d *decodeState) discriminatorGetValue() (reflect.Value, error) {
...
-				// continue walking over the current JSON object since it will be
-				// completely rescanned to decode its value into the discovered
-				// type.
-				dd.opcode = scanEndObject
-			default:
-				// Otherwise if the value offset has been discovered then it is
-				// safe to stop walking over the current JSON object as well.
+			// If type implements Unmarshaler we want to deserialize the
+			// value field into value of type t.
+			if supportsUnmarshaler(t) {
```
akutz, line 189: @karaatanassov ,

I am struggling to understand why `supportsUnmarshaler` is invoked twice in `discriminatorGetValue`. Can you please add some code comments to both instances? I _think_:

* The first place that `supportsUnmarshaler` is doing so when the op indicates the value has already been read, thus this is to unmarshal the value now that the key has been detected.
* The second place is when the value op is discovered _after_ the key.

Is that the way of it?

vim25/json/discriminator_test.go
```diff
@@ -59,6 +61,58 @@ type DS8 struct {
 	F1 DS3 `json:"f1"`
 }
 
+type DS9 struct {
+	F1 int64
+}
+
+func (d DS9) MarshalJSON() ([]byte, error) {
```
akutz, line 68: Can you please also write a test where the unmarshaler interface is not implemented by address? I ask because of:

```go
var unmarshalerType = reflect.TypeOf((*Unmarshaler)(nil)).Elem()

func supportsUnmarshaler(t reflect.Type) bool {
	return t.Implements(unmarshalerType) || reflect.PtrTo(t).Implements(unmarshalerType)
}
```

I am thinking something like the following:

```go
type DS10 struct {
	ds9 *DS9
}

func (d DS10) UnmarshalJSON(b []byte) error {
	if d.ds9 == nil {
		return nil
	}
	return d.ds9.UnmarshalJSON(b)
}
```

Since the helper func you wrote, `supportsUnmarshaler`, can act on both, there should be coverage for it.

vim25/json/discriminator.go
karaatanassov: Yes. I think you are right.

If a type supports Unmarshaler then it's schema is always:

```
{
     "type": "<discriminator>",
     "value": <value>
}
```
In other words we do not care if the reflect type is struct/map or not.  We switch to the mode that is used for primitive types (not map or struct). 

I do not like how the "if ... switch" thing looks like and specifically copy of the code. I had a "never nester" pattern there with complex if condition and it was harder to read for me. So I changed it to this nested variant.
```
if supportsUnmarshaler(t) || !isMapOrStruct(t) {
```
I can revise it that form if it makes more sense to you.

vim25/json/discriminator_test.go
karaatanassov: Will do.

akutz (dismissed):

akutz:
Wait on https://github.com/vmware/govmomi/pull/3136 before merging this

akutz (approved):

PR #3072, opened 2023-03-21
Detailed error message for execli errors

liron-l:
## Description

Include full fault details in error message, otherwise we get
```
EsxCLI.CLIFault.summary
```
And cannot diagnose the SDK errors

Closes: #3073

## Type of change

Please mark options that are relevant:

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [ ] This change requires a documentation update
- [ ] Build related change

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. If applicable, please also list any relevant
details for your test configuration.

- [x] Test Description 1
Verified fix against live environment that returns the error for objects that are not found
- [ ] Test Description 2

## Checklist:

- [x] My code follows the `CONTRIBUTION`
  [guidelines](https://github.com/vmware/govmomi/blob/main/CONTRIBUTING.md) of
  this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged

PR #3131, opened 2023-05-19
fix: add esxcli.Fault and revert Error() string

dougm:
PR #3072 added fault detail to the error message, breaking existing code at runtime.

### DIFF
--- a/govc/host/esxcli/executor.go
+++ b/govc/host/esxcli/executor.go
@@ -1,11 +1,11 @@
 /*
-Copyright (c) 2014-2015 VMware, Inc. All Rights Reserved.
+Copyright (c) 2014-2023 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


akutz (approved):
lgtm, thank you! I do wonder if we can look into the ESX sources to find out if there are standard error strings we can then write helper funcs for in this package. I'm thinking something similar to https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors, where you can do `errors.IsNotFound(err)` or `errors.IsInvalid(err)`, etc.

govc/host/esxcli/executor.go
```diff
@@ -27,6 +26,23 @@ import (
 	"github.com/vmware/govmomi/vim25/xml"
 )
 
+type Fault struct {
+	Message string
+	Detail  string
+}
+
+func (f *Fault) Error() string {
```
akutz, line 34: The receiver does not need to be by address, right? It's more efficient to treat this by-value.

akutz (approved):

PR #3145, opened 2023-06-01
vcsim: async guest shutdown and standby

dougm:
GuestShutdown and GuestStandby methods don't block when called in real vCenter, but the VM power state transition won't happen until somtime after the guest has actually shutdown. It is also possible for shutdown to hang indefinitely. To simulate such behavior, an internal task is now used to transition state, which can be optionally delayed.

### DIFF
--- a/govc/test/vcsim.bats
+++ b/govc/test/vcsim.bats
@@ -283,6 +283,10 @@ EOF
   run govc vm.power -s $vm
   assert_success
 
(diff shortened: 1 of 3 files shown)


akutz (approved):
Lgtm, thank you!

akutz:
@dougm, I did end up finding some issues with this. Let's chat tomorrow. Thanks!

PR #3177, opened 2023-07-13
Container backed hosts in simulator

hickeng:
## Description

Adds support for container backed hosts in the simulator, in a similar fashion to the current container backed VMs.
Uses a similar pattern of `RUN.container` but in `host.ConfigManager.AdvancedOption` as `ExtraConfig` doesn't exist for hosts.
See simulator/container_host_system_test.go for specific tests and usage.

In adding this support, this PR also:
* untangles container operations from the sim VM implementation
* adds per-host OptionManager

This results in the following hierarchy of files relating to containers:
1. container.go - basic container operations
2. container_host_system.go and container_virtual_machine.go - host and vm specifics when mocking using containers.
3. host_system.go and virtual_machine.go - sim objects but mostly without container specifics.

This is an initial PR with expected follow up:
* ESX mock container image providing mocks for esxcli, configstorecli, etc
* "hostd" in the sim host responds on host mgmt IP, backed by a single vcsim for all "hosts"

The goal of this work is to be able to easily test logic that depends hosts actually having a network presence, including ESX daemons, etc.

Known issues:
- containers/podman#19219 - causes deletion of all volumes when _any_ host is deleted
- cannot run container-backed-vms using podman if mounting DMI

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [ ] Regression tests in simulator/feature_test.go
- [ ] New tests in container_host_system_test.go

## Checklist:

- [ ] My code follows the `CONTRIBUTION`
  this project


hickeng:
Current failure in govc tests is because the test assumes (correctly I think) that you should be able to _change_ a VM to have `RUN.container` specified, then power it on and get a container backing.

However the current behaviour of this PR does not work for that path:
```bash
[root@ghicken-z01 test]# ./vendor/github.com/bats-core/bats-core/bin/bats vcsim.bats -f "run container"
vcsim.bats
 ✗ vcsim run container
   (from function `flunk' in file test_helper.bash, line 207,
    from function `assert_success' in file test_helper.bash, line 212,
    in test file vcsim.bats, line 258)
     `assert_success' failed
   Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
   Client:       Podman Engine
   Version:      4.4.1
   API Version:  4.4.1
   Go Version:   go1.18.10
   Built:        Fri Feb 17 02:31:22 2023
   OS/Arch:      linux/amd64
   Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
   []
   Error: inspecting object: no such object: "vcsim-DC0_H0_VM0-265104de-1472-547c-b873-6dc7883fb6cb"
   command failed with exit status 1: Powering on VirtualMachine:vm-55... govc: *types.MissingPowerOnConfiguration
   2023/07/14 07:22:22 vcsim-DC0_H0_VM0-265104de-1472-547c-b873-6dc7883fb6cb volumes: vcsim-DC0_H0_VM0-265104de-1472-547c-b873-6dc7883fb6cb--dmi
   2023/07/14 07:22:22 vcsim-DC0_H0_VM0-265104de-1472-547c-b873-6dc7883fb6cb [docker volume rm -f vcsim-DC0_H0_VM0-265104de-1472-547c-b873-6dc7883fb6cb--dmi ]: exit status 125, vcsim-DC0_H0_VM0-265104de-1472-547c-b873-6dc7883fb6cb--dmi

1 test, 1 failure

[root@ghicken-z01 test]#
```

I think the simplest path forwards is to create a go test unit test that follows this pattern as it's not straight forwards to use delve with the bats tests.

dougm:
> I think the simplest path forwards is to create a go test unit test that follows this pattern as it's not straight forwards to use delve with the bats tests.

There's a go example/test, creates the vm, but might be useful for this:
https://github.com/vmware/govmomi/blob/cd0cb63078984db7062d3fbe22ea24e44a457f42/simulator/feature_test.go#L89

We could also only support the container backing for new VMs, not sure adding a container backing to an existing vm is a common use-case?

hickeng:
> We could also only support the container backing for new VMs, not sure adding a container backing to an existing vm is a common use-case?

I've no idea if it's common, but I'm loath to change behaviour unless it's actively problematic to support. I don't see why it would be if I've structured things well... just not working that way atm. It does mean we should consider the inverse - removing the container backing from a VM without deleting it which I suspect doesn't exist atm.

I'll likely create a `container_virtual_machine_test.go` and move the container backed VM tests into that for consistency.
At some point I'm going to want to mock the VMCI channel which will require enhancing that surface and don't want to drop those tests into `feature-test.go`

hickeng:
Tidied up commits, updated messages to meet contributor guidelines, etc.

simulator/container_host_system.go
```diff
@@ -0,0 +1,362 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/vmware/govmomi/vim25/methods"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+const KiB = 1024
```
dougm, line 27: These are also in simulator/esx, could they be moved to the govmomi/units package or simulator/internal ?

simulator/host_system.go
```diff
@@ -106,6 +135,241 @@ func (h *HostSystem) configure(spec types.HostConnectSpec, connected bool) {
...
+//
+// The pnics will be named using standard pattern, ie. vmnic0, vmnic1, ...
+// This will sanity check the NetConfig for "management" nicType to ensure that it maps through PortGroup->vSwitch->pNIC to vmnic0.
+func (h *HostSystem) configureContainerBacking(ctx *Context, image string, mounts []types.HostFileSystemMountInfo, networks ...string) error {
+	option := &types.OptionValue{
+		Key:   advOptContainerBackingImage,
+		Value: image,
+	}
+
+	advOpts := Map.Get(h.ConfigManager.AdvancedOption.Reference()).(*OptionManager)
```
dougm, line 166: `ctx.Map`, same thing currently, but hopefully someday we'll get rid of the simulator.Map package var

simulator/container_virtual_machine_test.go
```diff
@@ -0,0 +1,249 @@
...
+// content - the contents of index.html
+// port - the port to forward to the container port 80
+func constructNginxBacking(t *testing.T, content string, port int) []types.BaseOptionValue {
+	dir := t.TempDir()
+	for dirpart := dir; dirpart != "/"; dirpart = filepath.Dir(dirpart) {
+		os.Chmod(dirpart, 0755)
+	}
+
+	fpath := filepath.Join(dir, "index.html")
+	os.WriteFile(fpath, []byte(content), 0644)
```
dougm, line 47: Check error return here?

dougm (approved):
Thank you @hickeng , this looks great. Just a couple of cosmetic suggestions for now, I might have more but can always follow up after merge

simulator/container_host_system.go
hickeng: @dougm I'm a little torn here - the units in units/size.go are not following the IEC naming pattern but are using KB instead of KiB.
My inclination is to move all references in this PR to reference the `units` package, then do a rename across the entire codebase as a dedicated commit - that seems to have wider implications. Not sure how constant names play into semver.

dougm: I don't think the govmomi/units package is used much outside of the repo, breaking compile time change would be ok in this case.

dougm (approved):
Ship it!

PR #3186, opened 2023-07-24
Bypass lookup svc for STS and SSOadmin clients

mayankbh:
## Description

- For local clients on vCenter, lookup service can be omitted and local Envoy processes can be used instead, along with 'system-' prefixed variants of SSOAdmin/STS. This avoids the need to discover the vCenter SSO domain and avoids an extra round trip.

- Also extends simulator to allow aliasing the same SDK to multiple paths as part of the RegisterSDK() method. Moves vimService, ssoadmin, STS simulators to this model.

- Using an env var with a fallback default to allow testing.

Closes: #(issue-number)

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- Existing unit tests
- On VC: created a client pointing at localhost:1080 and verified that lookup service was not used.
- Added unit tests that break lookup service URLs and ensure that ssoadmin/sts clients still work.

## Checklist:

- [X] My code follows the `CONTRIBUTION`
  this project

### DIFF
--- a/internal/helpers.go
+++ b/internal/helpers.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+Copyright (c) 2020-2023 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -18,8 +18,10 @@ package internal
 
 import (
 	"net"
+	"os"
 	"path"
 
+	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 )
@@ -60,4 +62,19 @@ func HostSystemManagementIPs(config []types.VirtualNicManagerNetConfig) []net.IP
 	}
 
 	return ips
+}
+
+// UsingEnvoySidecar determines if the given *vim25.Client is using vCenter's
+// local Envoy sidecar (as opposed to using the HTTPS port.)
+// Returns a boolean indicating whether to use the sidecar or not.
+func UsingEnvoySidecar(c *vim25.Client) bool {
+	envoySidecarPort := os.Getenv("GOVMOMI_ENVOY_SIDECAR_PORT")
+	if envoySidecarPort == "" {
+		envoySidecarPort = "1080"
+	}
+	envoySidecarHost := os.Getenv("GOVMOMI_ENVOY_SIDECAR_HOST")
+	if envoySidecarHost == "" {
+		envoySidecarHost = "localhost"
+	}
+	return c.URL().Hostname() == envoySidecarHost && c.URL().Scheme == "http" && c.URL().Port() == envoySidecarPort
 }
--- a/ssoadmin/client.go
+++ b/ssoadmin/client.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2018-2022 VMware, Inc. All Rights Reserved.
+Copyright (c) 2018-2023 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -18,11 +18,13 @@ package ssoadmin
 
 import (
 	"context"
+	"fmt"
 	"math"
 	"path"
 	"reflect"
 	"strings"
 
+	"github.com/vmware/govmomi/internal"
 	"github.com/vmware/govmomi/lookup"
 	ltypes "github.com/vmware/govmomi/lookup/types"
 	"github.com/vmware/govmomi/ssoadmin/methods"
@@ -33,9 +35,11 @@ import (
 )
 
 const (
-	Namespace = "sso"
-	Version   = "version2"
-	Path      = "/sso-adminserver" + vim25.Path
+	Namespace  = "sso"
+	Version    = "version2"
+	basePath   = "/sso-adminserver"
+	Path       = basePath + vim25.Path
+	SystemPath = basePath + "/system-sdk"
 )
 
 var (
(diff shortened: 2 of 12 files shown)


sts/client_test.go
```diff
@@ -380,3 +386,34 @@ func TestIssueActAs(t *testing.T) {
...
+		simulator.Test(func(ctx context.Context, client *vim25.Client) {
+			_, err := NewClient(ctx, client)
+			require.NoError(t, err)
+		})
+	})
+
+	t.Run("STS client should work with Envoy sidecar even when lookup service is down", func(t *testing.T) {
+		model := simulator.VPX()
+
+		model.Create()
```
dougm, line 401: Does this test need model.Create() to happen here? Suppose it doesn't hurt, but looks like default simulator.Test behavior would work the same here.

internal/helpers.go
```diff
@@ -61,3 +63,18 @@ func HostSystemManagementIPs(config []types.VirtualNicManagerNetConfig) []net.IP
 
 	return ips
 }
+
+// UsingVCEnvoySidecar determines if the given *vim25.Client is using vCenter's
+// local Envoy sidecar (as opposed to using the HTTPS port.)
+// Returns a boolean indicating whether to use the sidecar or not, and the sidecar port.
+func UsingVCEnvoySidecar(c *vim25.Client) (bool, string) {
```
dougm, line 70: Does port need to be returned, since it's the same as caller's c.URL().Port() when true?
Also minor, but maybe just URL as the input param?
And super minor, but env vars don't have 'VC' in the name, reads better w/o it in the function name too:
```go
func UsingEnvoySidecar(u net.URL) bool
```

sts/client.go
```diff
@@ -31,8 +33,10 @@ import (
 )
 
 const (
-	Namespace = "oasis:names:tc:SAML:2.0:assertion"
-	Path      = "/sts/STSService"
+	Namespace     = "oasis:names:tc:SAML:2.0:assertion"
+	BasePath      = "/sts"
+	Path          = BasePath + "/STSService"
+	SystemSTSPath = BasePath + "/system-STSService/sdk"
```
dougm, line 39: Already in `sts` package, so just `SystemPath` for the name is good. And if not used outside the package, `systemPath`.

ssoadmin/client.go
```diff
@@ -33,9 +35,11 @@ import (
...
 
 const (
-	Namespace = "sso"
-	Version   = "version2"
-	Path      = "/sso-adminserver" + vim25.Path
+	Namespace     = "sso"
+	Version       = "version2"
+	BasePath      = "/sso-adminserver"
+	Path          = BasePath + vim25.Path
+	SystemSDKPath = BasePath + "/system-sdk"
```
dougm, line 42: `SystemPath`

sts/client.go
```diff
@@ -31,8 +33,10 @@ import (
 )
 
 const (
-	Namespace = "oasis:names:tc:SAML:2.0:assertion"
-	Path      = "/sts/STSService"
+	Namespace     = "oasis:names:tc:SAML:2.0:assertion"
+	BasePath      = "/sts"
```
dougm, line 37: `basePath` if not used outside this package

sts/simulator/simulator.go
```diff
@@ -36,6 +36,7 @@ func init() {
 		if r.IsVPX() {
 			path, handler := New(s.Listen, r.OptionManager().Setting)
 			s.Handle(path, handler)
+			s.Handle("/sts/system-STSService/sdk", handler)
```
dougm, line 39: `sts.SystemPath` ?

dougm (commented):
Thanks @mayankbh , looks great. Just a few minor/cosmetic suggestions

sts/simulator/simulator.go
mayankbh: Hmm I think there was an import cycle. I think it's because sts/client_test.go is in the 'sts' package, not sts_test . I could look to move it to a separate package

sts/client_test.go
mayankbh: Hmm, yeah, good point. Might have been leftover from me trying to fiddle with the TLS config before I realized it was the `new(tls.Config)` thing tripping me up

sts/simulator/simulator.go
dougm: It's ok to leave this as-is or move `const SystemPath` to sts/internal.
Changing the client_test package is a-ok too, if not too much of a hassle.

sts/client_test.go
mayankbh: Ah, I remember. The existing Test/Run helpers, if not given a Model, will create a blank one with a Nil service. In that path, the TLS config is always set up (ie. simulator serves HTTPS) - which breaks the UT, as that requires HTTP

internal/helpers.go
mayankbh: Ah, good point. This also reminded me that it's preferable to use u.Host instead of Hostname():port, because the latter will break IPv6.

sts/client_test.go
mayankbh: I suppose I could in theory change the Test/Run helpers but that seemed like a larger blast radius.

sts/simulator/simulator.go
mayankbh: Fixed, was a clean change as it wasn't using anything internal to the sts package.

sts/client_test.go
dougm: Ah gotcha. Haven't tried, but seems like that'd need the patch below? 

```diff
modified   simulator/model.go
@@ -876,9 +876,10 @@ func (m *Model) Run(f func(context.Context, *vim25.Client) error) error {
 		if err != nil {
 			return err
 		}
+	} else {
+		m.Service.TLS = new(tls.Config)
 	}
 
-	m.Service.TLS = new(tls.Config)
 	m.Service.RegisterEndpoints = true
 
 	s := m.Service.NewServer()
```

dougm (approved):
Thanks @mayankbh , lgtm

PR #3188, opened 2023-07-25
vcsim: add lookup.EndpointURL tests

dougm:
- Add ssoadmin to lookup simulator

- Add tests validating sts/ssoadmin use of lookup.EndpointURL


mayankbh (approved):
LGTM. Thanks!

Issue #2521, opened 2021-07-19
[BUG] Fix status error returned from rest code


gnufied:
**Describe the bug**

When creating tags or cateogies, the sdk returns following error type:

https://github.com/vmware/govmomi/blob/master/vapi/rest/client.go#L178

```
return &statusError{res}
```

The problem is - users of library generally want to know the http error code at very minimum, so as they can make reasonable assumption about error. Such as - did getting a resource failed because of network error or was it 404?

embano1:
@dougm this is also something I think we should address mid-term. Not sure about the impact on the APIs, thoughts?

dougm:
Agreed, I had started to look at doing this a while back.. was thinking we could export the existing `soap.statusError` and also expose the `http.Response.StatusCode`.  Then use that error type in the rest client too.

https://github.com/vmware/govmomi/blob/4a7de71a1d515a62ecf5a62c043e7e3859359b13/vim25/soap/client.go#L560-L584

brakthehack:
Interested in seeing this come about. A few more options:

1. Make statusError public and have it carry core parts of the request such status code. If we expose it, then clients can at least cast to that error and grab the response information when they need to.
2. Optionally, update all signatures to return this error.
3. For operations that return an object with a pointer, make the pointer nil when 404 is encountered. Seems there are similar issues which report that this should be expected behavior: https://github.com/vmware/govmomi/issues/1639.

PR #3194, opened 2023-08-01
govc: Add -file flag to cluster.module.rm to delete a list of cluster modules read from a file

chrischdi:
## Description

This PR adds a `-file` flag to the `govc cluster.module.rm` command.

Deleting lots of cluster modules takes a long time.

E.g. deleting 10k cluster modules using bash and the current implementation would take ~5 mins. Example command:
```sh
while read -r ID; do govc cluster.module.rm $ID ; done < modules.txt
```

With this PR the following would do the same and only take ~30s:

```sh
govc cluster.module.rm - < modules.txt
```

Closes: #(issue-number)

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [x] Created 10k cluster modules and deleted using the new flag
  - extracted all existing modules via `govc cluster.module.ls | awk '{print $2}' > modules.txt`
  - ran `govc cluster.module.rm -file modules.txt`

## Checklist:

### DIFF
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -586,10 +586,14 @@ Usage: govc cluster.module.rm [OPTIONS] ID
 
 Delete cluster module ID.
 
+If ID is "-", read a list from stdin.
+
 Examples:
   govc cluster.module.rm module_id
+  govc cluster.module.rm - < input-file.txt
 
 Options:
+  -ignore-not-found=false  Treat "404 Not Found" as a successful delete.
 ```
 
 ## cluster.module.vm.add
--- a/govc/cluster/module/rm.go
+++ b/govc/cluster/module/rm.go
@@ -17,8 +17,11 @@ limitations under the License.
 package module
 
 import (
+	"bufio"
 	"context"
 	"flag"
+	"os"
+	"strings"
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
@@ -27,6 +30,7 @@ import (
 
 type rm struct {
 	*flags.ClientFlag
+	ignoreNotFound bool
 }
 
 func init() {
@@ -36,6 +40,8 @@ func init() {
 func (cmd *rm) Register(ctx context.Context, f *flag.FlagSet) {
 	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
 	cmd.ClientFlag.Register(ctx, f)
+
+	f.BoolVar(&cmd.ignoreNotFound, "ignore-not-found", false, "Treat \"404 Not Found\" as a successful delete.")
 }
 
 func (cmd *rm) Usage() string {
@@ -45,8 +51,11 @@ func (cmd *rm) Usage() string {
 func (cmd *rm) Description() string {
 	return `Delete cluster module ID.
 
+If ID is "-", read a list from stdin.
+
 Examples:
-  govc cluster.module.rm module_id`
+  govc cluster.module.rm module_id
+  govc cluster.module.rm - < input-file.txt`
 }
 
 func (cmd *rm) Run(ctx context.Context, f *flag.FlagSet) error {
@@ -60,6 +69,31 @@ func (cmd *rm) Run(ctx context.Context, f *flag.FlagSet) error {
 	if err != nil {
 		return err
 	}
+	m := cluster.NewManager(c)
 
-	return cluster.NewManager(c).DeleteModule(ctx, moduleID)
+	if moduleID == "-" {
+		scanner := bufio.NewScanner(os.Stdin)
+		for scanner.Scan() {
+			moduleID := scanner.Text()
+			if moduleID == "" {
+				continue
+			}
+			if err := cmd.deleteModule(ctx, m, moduleID); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+
+	return cmd.deleteModule(ctx, m, moduleID)
+}
+
+func (cmd *rm) deleteModule(ctx context.Context, m *cluster.Manager, moduleID string) error {
+	if err := m.DeleteModule(ctx, moduleID); err != nil {
+		if cmd.ignoreNotFound && strings.HasSuffix(err.Error(), "404 Not Found") {
+			return nil
+		}
+		return err
+	}
+	return nil
 }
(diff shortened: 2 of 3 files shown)


govc/cluster/module/rm.go
```diff
@@ -36,30 +40,53 @@ func init() {
...
+
 Examples:
-  govc cluster.module.rm module_id`
+  govc cluster.module.rm module_id
+  govc cluster.module.rm -file modules.txt`
 }
 
 func (cmd *rm) Run(ctx context.Context, f *flag.FlagSet) error {
-	if f.NArg() != 1 {
+	if f.NArg() != 1 && cmd.file == "" {
```
bryanv, line 62: This doesn't look quite right: does this return ErrHelp for eg `-file=foo bar1 bar2`?

govc/USAGE.md
```diff
@@ -36,338 +36,339 @@ but appear via `govc $cmd -h`:
...
- - [vm.upgrade](#vmupgrade)
- - [vm.vnc](#vmvnc)
- - [volume.ls](#volumels)
- - [volume.rm](#volumerm)
- - [volume.snapshot.create](#volumesnapshotcreate)
- - [volume.snapshot.ls](#volumesnapshotls)
- - [volume.snapshot.rm](#volumesnapshotrm)
- - [vsan.change](#vsanchange)
- - [vsan.info](#vsaninfo)
+- [govc usage](#govc-usage)
```
bryanv, line 39: Oh - did we miss a regen from a prior change? Maybe 48bdfada6aa. Looks like the only change here is the first `[govc usage](#govc-usage)` line, and then that caused an extra space ident to be added.

chrischdi: That was auto format of vscode adding a section from above (treating this toc as toc for the whole file but it is just for the below parts).

govc/cluster/module/rm.go
```diff
@@ -36,30 +40,53 @@ func init() {
...
+		}
+		defer modulesFile.Close()
+
+		scanner := bufio.NewScanner(modulesFile)
+		for scanner.Scan() {
+			moduleID := scanner.Text()
+			if moduleID == "" {
+				continue
+			}
+			m.DeleteModule(ctx, moduleID)
```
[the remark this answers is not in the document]
chrischdi: adjusted to return. But I'd be also okay with aggregating.

With aggregating it would process the next entries and maybe ddos a server which is in trouble even more?!

govc/cluster/module/rm.go
chrischdi: Ah yeah, I should not rush a PR shortly before calling it a day ... 👍

govc/cluster/module/rm.go
```diff
@@ -36,30 +40,56 @@ func init() {
...
+
 Examples:
-  govc cluster.module.rm module_id`
+  govc cluster.module.rm module_id
+  govc cluster.module.rm -file modules.txt`
 }
 
 func (cmd *rm) Run(ctx context.Context, f *flag.FlagSet) error {
-	if f.NArg() != 1 {
+	if !(f.NArg() == 1 && cmd.file == "") && !(f.NArg() == 0 && cmd.file != "") {
```
chrischdi, line 62: That should do the trick.

There are other valid variants but I think this one is the most readable.

E.g.:
```go
if f.NArg() > 1 || (f.NArg() == 1 && cmd.file != "") || (f.NArg() == 0 && cmd.file == "")
```

```go
if !((f.NArg() == 1 && cmd.file == "") || (f.NArg() == 0 && cmd.file != "")) {
```

```go
if (f.NArg() != 1 || cmd.file != "") && (f.NArg() != 0 || cmd.file == "") {
```

chrischdi:
Many thanks @bryanv for the review. I addressed the comments. I should next time not rush a PR before calling it a day 🤦

hickeng (commented):
Unless the "list in a file" is a pattern used elsewhere, consider instead the "read from stdin when `-` is specified" pattern.

That's a standard unix idiom and allows for use in scripting pipelines.
When used as an isolated command either a `<` redirection or a HERE document can be used to transfer a file/inline list to stdin.

dougm:
> Unless the "list in a file" is a pattern used elsewhere, consider instead the "read from stdin when `-` is specified" pattern.

We have a few cases where `-` == stdin, such as {guest,datastore}.upload commands. But we don't have a pattern where a `-file` flag is used to specify N arguments (such as IDs in this case)

We already have inconsistency where some commands take just the argument needed for a single API call (such as cluster.module.rm currently), but others take N arguments and call the method N within its own loop (such as vm.destroy).
So I'd prefer not to add a special case `-file` flag to this command, but would be ok with 
`govc cluster.module.rm -` reading a list of IDs from stdin.

We could later add helpers to apply this pattern and others such as parallel API calls, etc., to a broader set of commands.

Note that we do also have a way to hide commands and flags by default: https://github.com/vmware/govmomi/blob/ba843e53ef62703cf15f0f263c003d720b63cc1e/govc/cli/register.go#L25-L32

Used with a flag here for example: https://github.com/vmware/govmomi/blob/ba843e53ef62703cf15f0f263c003d720b63cc1e/govc/library/deploy.go#L66-L68

chrischdi:
Thanks for the great response. Using stdin also makes the implementation more clean and fulfils the same functionality 👍

I also added bats tests to cover that functionality :-)

chrischdi:
@dougm and @hickeng kindly asking for giving this another look :-)

govc/cluster/module/rm.go
```diff
@@ -60,6 +69,31 @@ func (cmd *rm) Run(ctx context.Context, f *flag.FlagSet) error {
...
+		return nil
+	}
 
-	return cluster.NewManager(c).DeleteModule(ctx, moduleID)
+	return cmd.deleteModule(ctx, m, moduleID)
+}
+
+func (cmd *rm) deleteModule(ctx context.Context, m *cluster.Manager, moduleID string) error {
+	if err := m.DeleteModule(ctx, moduleID); err != nil {
+		if cmd.ignoreNotFound && strings.HasSuffix(err.Error(), "404 Not Found") {
```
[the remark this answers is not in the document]
dougm: Meant to add, we have an issue open for this: https://github.com/vmware/govmomi/issues/2521

dougm (approved):
Thanks @chrischdi , lgtm

PR #3249, opened 2023-09-28
api: disable HTTP Keep-Alive for direct ESX connections

dougm:
guest file transfer and ticketed datastore file transfer use direct ESX connections. The connection should be only be used for a single request, ensure this by disabling client side.

### DIFF
--- a/guest/toolbox/client.go
+++ b/guest/toolbox/client.go
@@ -284,6 +284,7 @@ func (c *Client) Download(ctx context.Context, src string) (io.ReadCloser, int64
 	}
 
 	p := soap.DefaultDownload
+	p.Close = true // disable Keep-Alive connection to ESX
 
 	if internal.UsingEnvoySidecar(c.ProcessManager.Client()) {
 		vc = internal.ClientWithEnvoyHostGateway(vc)
@@ -345,6 +346,8 @@ func (c *Client) Upload(ctx context.Context, src io.Reader, dst string, p soap.U
 	if err != nil {
 		return err
 	}
+
+	p.Close = true // disable Keep-Alive connection to ESX
 
 	if internal.UsingEnvoySidecar(c.ProcessManager.Client()) {
 		vc = internal.ClientWithEnvoyHostGateway(vc)
--- a/object/datastore.go
+++ b/object/datastore.go
@@ -278,7 +278,10 @@ func (d Datastore) uploadTicket(ctx context.Context, path string, param *soap.Up
 		return nil, nil, err
 	}
 
-	p.Ticket = ticket
+	if ticket != nil {
+		p.Ticket = ticket
+		p.Close = true // disable Keep-Alive connection to ESX
+	}
 
 	return u, &p, nil
 }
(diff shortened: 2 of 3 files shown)


guest/toolbox/client.go
```diff
@@ -346,6 +347,8 @@ func (c *Client) Upload(ctx context.Context, src io.Reader, dst string, p soap.U
 		return err
 	}
 
+	p.Close = true // disable Keep-Alive connection to ESX
```
mayankbh, line 350: Hmm, in the Envoy sidecar scenario in https://github.com/vmware/govmomi/pull/3227/files#diff-85b38e4cec5f84a008e29777420bd336a111a4685b883f55c4a7de9e96572a64R113 , I used DisableKeepAlives, will this supersede that?

mayankbh:
```	// DisableKeepAlives, if true, disables HTTP keep-alives and
	// will only use the connection to the server for a single
	// HTTP request.
```

```
	// Close indicates whether to close the connection after
	// replying to this request (for servers) or after sending this
	// request and reading its response (for clients).
	//
	// For server requests, the HTTP server handles this automatically
	// and this field is not needed by Handlers.
	//
	// For client requests, setting this field prevents re-use of
	// TCP connections between requests to the same hosts, as if
	// Transport.DisableKeepAlives were set.
```

Not quite sure I see the distinction..

mayankbh: Ah, nvm. Scrolled up to see the actual structs, one's on the Request, one's on the Transport

dougm: `transport.DisableKeepAlives = true` applies to every request made by the http.Client(s) using that transport. `Request.Close` only applies to that specific request.

mayankbh (approved):

object/datastore.go
```diff
@@ -278,7 +278,10 @@ func (d Datastore) uploadTicket(ctx context.Context, path string, param *soap.Up
 		return nil, nil, err
 	}
 
-	p.Ticket = ticket
+	if ticket != nil {
+		p.Ticket = ticket
```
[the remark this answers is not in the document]
dougm: Yes, ticket is only needed when a specific host is used to transfer a datastore file. By default, govmomi connects to vCenter for datastore file transfer, which does not use a ticket.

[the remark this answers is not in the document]
dougm: Example where we would use a ticket + direct host connection: https://github.com/vmware/govmomi/blob/f859a9463cb6effe310b359085335236a3d20701/govc/datastore/download.go#L91-L94

Use case: downloading a file the host has locked, e.g. a VM's vmware.log file
In the default case, vCenter chooses a random host to download from, which might not be the host running the VM, in which case 500 server error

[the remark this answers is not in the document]
mayankbh: > Yes, ticket is only needed when a specific host is used to transfer a datastore file. By default, govmomi connects to vCenter for datastore file transfer, which does not use a ticket.TIL. I think you explained this to me before and never foudn the code path that'd trigger that.

PR #3256, opened 2023-10-10
vcsim: fix ModifyListView to return unresolved references

dougm:
api: add ListView.Modify method, returns unresolved references

### DIFF
--- a/simulator/view_manager.go
+++ b/simulator/view_manager.go
@@ -227,8 +227,8 @@ func (m *ViewManager) CreateListView(ctx *Context, req *types.CreateListView) so
 	body := new(methods.CreateListViewBody)
 	list := new(ListView)
 
-	if err := list.add(ctx, req.Obj); err != nil {
-		body.Fault_ = Fault("", err)
+	if refs := list.add(ctx, req.Obj); len(refs) != 0 {
+		body.Fault_ = Fault("", &types.ManagedObjectNotFound{Obj: refs[0]})
 		return body
 	}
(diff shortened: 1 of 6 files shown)


simulator/view_manager.go
```diff
@@ -249,15 +249,16 @@ func (v *ListView) update(ctx *Context) {
 	ctx.Map.Update(v, []types.PropertyChange{{Name: "view", Val: v.View}})
 }
 
-func (v *ListView) add(ctx *Context, refs []types.ManagedObjectReference) *types.ManagedObjectNotFound {
+func (v *ListView) add(ctx *Context, refs []types.ManagedObjectReference) []types.ManagedObjectReference {
+	var unresolved []types.ManagedObjectReference
 	for _, ref := range refs {
 		obj := ctx.Session.Get(ref)
 		if obj == nil {
-			return &types.ManagedObjectNotFound{Obj: ref}
```
adikul30, line 256: Does this mean that before this change, if I added 10 task MoRefs at once, and the first one happened to be expxired, `add()` would directly return a fault to the simulator side `ModifyListView()` without adding the rest 9 tasks?

adikul30 (approved):
LGTM

adikul30 (approved):

simulator/view_manager.go
dougm: That's right. And with the change, all valid refs are added to the view and all invalid refs are returned.

PR #3259, opened 2023-10-13
vapi: add support for OperationID header

dougm:
soap.Client sends OperationID as part of the soap body header.
vapi uses the `X-Request-ID` http header.

### DIFF
--- a/vapi/rest/client.go
+++ b/vapi/rest/client.go
@@ -31,6 +31,7 @@ import (
 	"github.com/vmware/govmomi/vapi/internal"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
 )
 
 // Client extends soap.Client to support JSON encoding, while inheriting security features, debug tracing and session persistence.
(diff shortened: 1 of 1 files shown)


vapi/rest/client.go
```diff
@@ -179,6 +180,11 @@ func (c *Client) Do(ctx context.Context, req *http.Request, resBody interface{})
 		}
 	}
 
+	// OperationID (see soap.Client.soapRoundTrip)
+	if id, ok := ctx.Value(types.ID{}).(string); ok {
```
fdawg4l, line 184: Had no idea you could use an empty var as a value key.  Makes perfect sense.  Very elegant.

fdawg4l (approved):

fdawg4l:
Are we generating operation IDs for every request?

dougm:
> Are we generating operation IDs for every request?

govmomi doesn't generate the operation IDs, that's up to the caller. But we've been passing it to [soap.Client calls](https://github.com/vmware/govmomi/blob/8726f228c54e3f25c9c1c7a4b58db0b45d30598e/vim25/soap/client.go#L640-L642) for years (since VIC days), this just does the same for vapi/rest calls.

PR #3261, opened 2023-10-16
vcsim: Handle prepare guest operations when svm is nil

mayankbh:
## Description

It appears that the simulated VM logic is supposed to handle cases when svm (the receiver) is nil (eg. when you spin up a VM without using a backing container. (I think this might have been introduced by fdb4d84763b8b28f3dc94b1a2a0b0f60aae2eb55 )

This would lead to a panic due to the conditional in question attempting to access svm.vm.Runtime.PowerState immediately after.

Inverting the check to return GuestOperationsUnavailable if `svm == nil` seems the simplest way to address this. To my knowledge, Go will short circuit evaluation.

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- Used a go workspace, updated a vcsim based unit test involving guest file transfer. Verified that test no longer panicked, but returned a GuestOperationsUnavailable soap fault.

(still trying to find a unit test in govmomi that would be impacted by this. I guess most tests use container backed VMs?) 

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works - will defer for now, this seems to be an untested path and my main priority was to avoid vcsim panics.

### DIFF
--- a/simulator/container_virtual_machine.go
+++ b/simulator/container_virtual_machine.go
@@ -131,7 +131,7 @@ func (svm *simVM) syncNetworkConfigToVMGuestProperties() error {
 }
 
 func (svm *simVM) prepareGuestOperation(auth types.BaseGuestAuthentication) types.BaseMethodFault {
-	if svm != nil && (svm.c == nil || svm.c.id == "") {
+	if svm == nil || svm.c == nil || svm.c.id == "" {
 		return new(types.GuestOperationsUnavailable)
 	}
 


dougm (approved):
Thanks @mayankbh , lgtm

hickeng:
I'm not sure this change actually gives you what you want. What's your desired behaviour when invoking any type of guest interaction without a sim backing? Is this actually a call to guest file manager against a VM without container backing, or just one that hasn't been powered on yet to create the container?

From slack, your issue started with a call to `InitiateFileTransferToGuest`. If there isn't a sim backing what is expected/desired to occur, success or `GuestOperationsUnavailable` regardless of VM power state? Is the PR's perm fail actually useful from a testing perspective?

I recommend adding a representative unit test as it's apparent we're not exercising that path currently, and also to formalize the expected behaviour without sim backing for your use case.

hickeng (approved):

mayankbh:
> Wanted to make sure you didn't have a need for non-error returns for GuestOps from simulated VM without backing

PR #3272, opened 2023-10-31
api: add Content Library update session file APIs

### LINKED ISSUE #3258: [BUG] Some operations are missing in the Content Library update session file API
**Describe the bug**
Some operations are missing in the Content Library update session file API.

**To Reproduce**

In this [library_item_updatesession_file.go](https://github.com/vmware/govmomi/blob/main/vapi/library/library_item_updatesession_file.go), a couple of operations that are available on [vSphere API](https://developer.vmware.com/apis/vsphere-automation/latest/content/library/item.updatesession.file/) are missing here:
- list
- remove
- validate
- probe (not released yet but available for a while now)

**Expected behavior**
Expect the above operations to be available in the [library_item_updatesession_file.go](https://github.com/vmware/govmomi/blob/main/vapi/library/library_item_updatesession_file.go) in order to leverage govmomi for content library integration work in Golang.

**Affected version**
Most of the above operations are available since vSphere 6.0, with the exception of probe API not released yet.

dougm:
Closes #3258

### DIFF
--- a/govc/library/probe.go
+++ b/govc/library/probe.go
@@ -1,0 +1,115 @@
+/*
+Copyright (c) 2023-2023 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package library
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"text/tabwriter"
+
+	"github.com/vmware/govmomi/govc/cli"
+	"github.com/vmware/govmomi/govc/flags"
+	"github.com/vmware/govmomi/vapi/library"
+)
+
+type probe struct {
+	*flags.ClientFlag
+	*flags.OutputFlag
+
+	fail bool
+}
+
+func init() {
+	cli.Register("library.probe", &probe{}, true)
+}
+
+func (cmd *probe) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.ClientFlag, ctx = flags.NewClientFlag(ctx)
+	cmd.ClientFlag.Register(ctx, f)
+
+	cmd.OutputFlag, ctx = flags.NewOutputFlag(ctx)
+	cmd.OutputFlag.Register(ctx, f)
+
+	f.BoolVar(&cmd.fail, "f", false, "Fail if probe status is not success")
+}
+
+func (cmd *probe) Usage() string {
+	return "URI"
+}
+
+func (cmd *probe) Description() string {
+	return `Probes the source endpoint URI with https or http schemes.
+
+Examples:
+  govc library.probe https://example.com/file.ova`
+}
+
+type probeResult struct {
+	*library.ProbeResult
+}
+
+func (r *probeResult) Write(w io.Writer) error {
+	tw := tabwriter.NewWriter(w, 2, 0, 2, ' ', 0)
+
+	fmt.Fprintf(tw, "Status:\t%s\n", r.Status)
+	thumbprint := r.SSLThumbprint
+	if thumbprint == "" {
+		thumbprint = "-"
+	}
+	fmt.Fprintf(tw, "Thumbprint:\t%s\n", thumbprint)
+	for _, e := range r.ErrorMessages {
+		fmt.Fprintf(tw, "%s:\t%s\n", e.ID, e.Error())
+	}
+
+	return tw.Flush()
+}
+
+func (cmd *probe) Process(ctx context.Context) error {
+	if err := cmd.ClientFlag.Process(ctx); err != nil {
+		return err
+	}
+	return cmd.OutputFlag.Process(ctx)
+}
+
+func (cmd *probe) Run(ctx context.Context, f *flag.FlagSet) error {
+	if f.NArg() != 1 {
+		return flag.ErrHelp
+	}
+
+	c, err := cmd.RestClient()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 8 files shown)


govc/library/probe.go
```diff
@@ -0,0 +1,115 @@
...
+
+	f.BoolVar(&cmd.fail, "f", false, "Fail if probe status is not success")
+}
+
+func (cmd *probe) Usage() string {
+	return "URI"
+}
+
+func (cmd *probe) Description() string {
+	return `Probe source endpoint.
```
ericvmw, line 58: nit: Probe remote HTTPS endpoint.

govc/library/probe.go
```diff
@@ -0,0 +1,115 @@
...
+
+Examples:
+  govc library.probe http://example.com/file.ova`
```
ericvmw, line 61: nit: use https:// instead of http://

govc/library/probe.go
```diff
@@ -0,0 +1,115 @@
...
+
+type probeResult struct {
+	*library.ProbeResult
+}
+
+func (r *probeResult) Write(w io.Writer) error {
+	tw := tabwriter.NewWriter(w, 2, 0, 2, ' ', 0)
+
+	fmt.Fprintf(tw, "Status:\t%s\n", r.Status)
+	thumbprint := r.SSLThumbprint
```
ericvmw, line 72: Just a note: a new field for the full SSL certificate is being added to the ProbeResult, and will be released soon.

govc/library/session/ls.go
```diff
@@ -102,26 +107,45 @@ func (cmd *ls) Run(ctx context.Context, f *flag.FlagSet) error {
...
 			item, err := m.GetLibraryItem(ctx, session.LibraryItemID)
-			if err != nil {
-				return err
+			if err == nil {
+				// can only show library path if item exists
+				lib, err := m.GetLibraryByID(ctx, item.LibraryID)
+				if err != nil {
+					return err
+				}
+				session.LibraryItemID = fmt.Sprintf("/%s/%s", lib.Name, item.Name)
```
ericvmw, line 127: Curious:
why use lib name and item name as the libraryItemID instead of the real UUID of the item? I'm not familiar with the API code here, I guess this is following some existing pattern in govmomi?

vapi/library/library_item_updatesession_file.go
```diff
@@ -66,24 +73,31 @@ func (c *Manager) AddLibraryItemFile(ctx context.Context, sessionID string, upda
 }
 
 // AddLibraryItemFileFromURI adds a file from a remote URI.
```
ericvmw, line 75: nit: adds a file from a remote URI to the update session and completes the update session.

govc/library/session/ls.go
dougm: In general govc uses human-readable names rather than IDs where possible. Such as using name or inventory path rather than MOID. And in the case of content library, we also support a [named path](https://github.com/vmware/govmomi/blob/main/vapi/library/finder/README.md)
But it is also possible to use the ID instead if needed.

govc/library/probe.go
dougm: ack. `ProbeResult` includes an `SSLCertificate` field that'll be included with the `-json` output.

govc/library/probe.go
dougm: http is also supported, but will qualify that with the 2 supported schemes. Also note this library.probe command is currently disabled by default, enabled with env `GOVC_SHOW_UNRELEASED=true`.

vapi/library/library_item_updatesession_file.go
dougm: Added. I don't remember why we're calling `CompleteLibraryItemUpdateSession` within this method - we could remove that call. I checked GitHub (and bora), don't see any references to the `AddLibraryItemFileFromURI` method outside of govc. What do you think?

govc/library/session/ls.go
ericvmw: I see. It's just a little confusing when I see the field name indicate an ID, I guess it's fine as long as users can navigate objects following this pattern, i.e. able to find the library item given an update session ID.

vapi/library/library_item_updatesession_file.go
ericvmw: Thanks! I guess it tries to provide a convenient wrap interface that can finish the file import within one call, though the name may not indicate that clearly, I think it's better to provide consistent interfaces, if this is not a pattern we typically expose APIs in govmomi, it makes sense to decouple completeUpdateSession.

vapi/simulator/simulator.go
```diff
@@ -1443,6 +1524,10 @@ func (s *handler) libraryItemUpdateSessionFileID(w http.ResponseWriter, r *http.
 		}
 		OK(w, ids)
 	case "remove":
+		if up.State != "ACTIVE" {
+			s.error(w, fmt.Errorf("removeFile not allowed in state %s", up.State))
+			return
+		}
 		delete(s.Update, id)
 		OK(w)
 	case "validate":
```
ericvmw, line 1533: `validate` impl would be also needed for complete match of the [vSphere API](https://developer.vmware.com/apis/vsphere-automation/latest/content/api/content/library/item/update-session/update_session_id/fileactionvalidate/post/).

govc/library/session/ls.go
dougm: That was just re-using an existing type + field. Updated to use a helper type, so `LibraryItemID` is left as-is as you can see with the `-json` output.

vapi/library/library_item_updatesession_file.go
dougm: Agree it's better to decouple here, done.

vapi/simulator/simulator.go
dougm: Ah I missed validate listed in the issue. I will add that, may not be till tomorrow though.

dougm: Added client side API for validate. vcsim side still doesn't do any validation of files, but at least returns a response body now (`{has_errors=false}`). I can still add that, but client-side is ready to use

ericvmw: Cool, thanks!

ericvmw (approved):
Thanks for adding the support!

dougm:
Thanks @ericvmw !

PR #3277, opened 2023-11-03
api: add object.Datastore.FindInventoryPath

### LINKED ISSUE #3267: [BUG] `object.Datastore.NewURL()` uses the wrong field for dsName query
**Describe the bug**
The `Datastore.NewURL()` function composes a `url.URL` to get a file via the `/folder` endpoint. This endpoint requires the `dsName` query parameter to be set to the name of the datastore. In the current implementation the function `d.Name()` is used to fill this parameter. However, this function is embedded from `Common.Name()` and does not return the real name of the datastore. Instead it returns its inventory path or "" most of the time.

**To Reproduce**
Use the `Datastore.NewURL()` function to generate a download URL

**Expected behavior**
Instead of `d.Name()` the function should use `d.ObjectName(context.Context)` which receives the real name of the datastore from the VSX/ESX itself.

**Affected version**
<= v.0.32

**Additional context**
The solution poses a problem, since the  `d.ObjectName(context.Context)` call does require a context for the request scope. This is very unfortunate, but the current implementation is pretty useless anyways (and might not be used, otherwise this must have come up earlier.)

I would prepare a PR to fix this, if my proposed fix is okay for you (or another solution of course)
Cheers

dougm:
Helper to set Datastore InventoryPath and DatacenterPath, needed by NewURL() to compose an upload/download endpoint URL

Fixes #3267

### DIFF
--- a/object/datastore.go
+++ b/object/datastore.go
@@ -68,6 +68,31 @@ func NewDatastore(c *vim25.Client, ref types.ManagedObjectReference) *Datastore 
 	}
 }
 
+// FindInventoryPath sets InventoryPath and DatacenterPath,
+// needed by NewURL() to compose an upload/download endpoint URL
+func (d *Datastore) FindInventoryPath(ctx context.Context) error {
+	entities, err := mo.Ancestors(ctx, d.c, d.c.ServiceContent.PropertyCollector, d.r)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


aruneshpa (approved):
LGTM

PR #3292, opened 2023-11-21
vcsim: handle HostNotConnected when saving inventory

dougm:


dbhat-arkin:
Thanks @dougm for the changes.
The changes looks good to me.

PR #3331, opened 2024-01-08
api: WaitForUpdatesEx & DestroyPropertyFilter

akutz:
## Description

This patch introduces Ex variants for many of the property collector types/methods to take advantage of the remote API DestroyPropertyFilter. Now it is possible to use a single property collector for waiting for updates and remove previously added filters.

```
BREAKING: The semantics around the helper functions in the
          property package have changed. Please review any
          code that calls this package to ensure it is
          compatible with the new behaviors.
```

Closes: `NA`

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [x] New unit test for `property.WaitForUpdatesEx`
- [x] New unit test for `property.Collector`'s `WaitForUpdatesEx` function
- [x] New unit test for `DestroyPropertyFilter` 

## Checklist:

### DIFF
--- a/property/collector.go
+++ b/property/collector.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2015-2023 VMware, Inc. All Rights Reserved.
+Copyright (c) 2015-2024 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -19,6 +19,8 @@ package property
 import (
 	"context"
 	"errors"
+	"fmt"
+	"sync"
 
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/methods"
@@ -27,11 +29,19 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
+// ErrConcurrentCollector is returned from WaitForUpdates, WaitForUpdatesEx,
+// or CheckForUpdates if any of those calls are unable to obtain an exclusive
+// lock for the property collector.
+var ErrConcurrentCollector = fmt.Errorf(
+	"only one goroutine may invoke WaitForUpdates, WaitForUpdatesEx, " +
+		"or CheckForUpdates on a given PropertyCollector")
+
 // Collector models the PropertyCollector managed object.
 //
 // For more information, see:
 // http://pubs.vmware.com/vsphere-60/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc%2Fvmodl.query.PropertyCollector.html
 type Collector struct {
+	mu           sync.Mutex
 	roundTripper soap.RoundTripper
 	reference    types.ManagedObjectReference
 }
@@ -46,7 +56,7 @@ func DefaultCollector(c *vim25.Client) *Collector {
 	return &p
 }
 
-func (p Collector) Reference() types.ManagedObjectReference {
+func (p *Collector) Reference() types.ManagedObjectReference {
 	return p.reference
 }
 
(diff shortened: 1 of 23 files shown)


property/collector_test.go
```diff
@@ -0,0 +1,316 @@
...
+			t.Fatalf("error while powering on vm: %s", err)
+		}
+
+		select {
+		case <-time.After(3 * time.Second):
+			t.Fatalf("timed out while waiting for property update")
+		case result := <-chanResult:
+			switch tResult := result.(type) {
+			case types.VirtualMachinePowerState:
+				if tResult != types.VirtualMachinePowerStatePoweredOn {
```
akutz, line 102: @dougm I was stuck on this for several hours yesterday. Well, an earlier version of it. It was when I realized that the value of the change was _not_ a string, but a `VirtualMachinePowerState` :)

dougm (commented):
Thanks @akutz 
The new property.WaitForUpdatesEx lgtm.
Keep in mind `ServiceContent.PropertyCollector` (aka property.DefaultCollector()) is a session singleton, not safe for multiple threads to WaitForUpdatesEx on.
object.Task.{Wait,WaitForResult} and the new Ex versions both use DefaultCollector.Create(), so the Ex version(s) are not thread safe.
Thinking:
- leave the existing `object.Task` methods as-is
- change `task.Wait` to always use property.WaitForUpdatesEx, rather than add task.WaitEx (don't think anyone currently calls task.Wait directly?)
- then apps would call `task.Wait` to re-use a PC instance

property.Filter - tempted to do a breaking rename of the existing type, to property.Match or ?
Then property.Filter can be used to wrap the MO instead of FilterEx

akutz:
Hi @dougm ,

Thanks for taking a look. Yeah, the concurrency thing is an issue. What are your thoughts on starting `WaitForUpdatesEx` in a Goroutine the moment a session is established, using the default PC, and then we change the way `Wait` occurs for everybody else to:

* Register on update callback 
* Add prop filter
* Receive update via callback
* Remove prop filter

That way there's always a single, long-running WaitForUpdates goroutine that can be used to listen for all updates, whenever there's a prop filter added?

akutz:
Hey @dougm,

Please see the refactored PR. I added a lock to the `PropertyCollector` such that any calls to `WaitForUpdates`, `WaitForUpdatesEx`, or `CheckForUpdates` must obtain an exclusive lock on the property collector. The attempt to obtain the lock will timeout if the context is cancelled.

property/collector.go
```diff
@@ -96,7 +124,27 @@ func (p *Collector) CreateFilter(ctx context.Context, req types.CreateFilter) er
 	return nil
 }
 
-func (p *Collector) WaitForUpdates(ctx context.Context, version string, opts ...*types.WaitOptions) (*types.UpdateSet, error) {
+func (p *Collector) CreateFilterEx(ctx context.Context, req types.CreateFilter) (*Filter, error) {
```
dougm, line 127: couple of breaking changes already, how about changing the existing `CreateFilter` to return `(*Filter, error)` instead of adding `CreateFilterEx` ?

property/collector.go
```diff
@@ -96,7 +124,27 @@ func (p *Collector) CreateFilter(ctx context.Context, req types.CreateFilter) er
...
+
+	return &Filter{roundTripper: p.roundTripper, reference: resp.Returnval}, nil
+}
+
+func (p *Collector) WaitForUpdates(
+	ctx context.Context,
+	version string,
+	opts ...*types.WaitOptions) (*types.UpdateSet, error) {
+
+	if err := p.lock(ctx); err != nil {
```
dougm, line 143: Rather than block, should we return an error or even panic if a Collector instance is already in WaitForUpdates? And document a Collector instance's WaitForUpdates can only be used by a single thread?

dougm (commented):
Thanks @akutz , can you also add a `BREAKING: ` section to the commit? Suggesting one more for the list (CreateFilter)

property/collector.go
akutz: Will do. I'll switch to using `TryLock` from a mutex and if it fails, either panic or return an error.

dougm (approved):
lgtm, thanks @akutz !

Issue #3339, opened 2024-01-17
Provide applicaiton self identification in the default user-agent value


karaatanassov:
**Is your feature request related to a problem? Please describe.**
To debug issues in vCenter it is hard to attribute API call to specific application just by knowing that govmomi is used.

govmomi is used widely by kubernates operators, terraform and other software packages. It is thus becoming harder to identify the source off API calls by user agent. 

**Describe the solution you'd like**
Add the name of the executable in the default user agent of govmomi.

**Describe alternatives you've considered**
Work with govmomi consumers to provide customised user agents. There are more and more govmomi consumers and working with each consumers, making pull requests etc is difficult and unnecessary chore This has the drawback that the other meaningful metadata i.e. go version, Os may get lost when people only identify their product and version.

PR #3340, opened 2024-01-17
Self identify the executable making calls to vCenter

karaatanassov:
When reviewing data from vCenter for troubleshooting and analysis it is hard to distinguish where the request comes from. Having the govmomi and golang version is helpful. As we now have increasing number of go clients we need more data.

The go debug value produce din tests is like
"__debug_bin2861234748 govmomi/0.31.0 (go1.20.5;darwin;arm64)"

## Description

Add the executable name to the default user agent 

Closes: #3339 

## Type of change

- [ X] New feature (non-breaking change which adds functionality)
  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [X] Unit tests for the user-agent

## Checklist:

### DIFF
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -59,9 +59,10 @@ const (
 )
 
 // defaultUserAgent is the default user agent string, e.g.
-// "govmomi/0.28.0 (go1.18.3;linux;amd64)"
+// "govc govmomi/0.28.0 (go1.18.3;linux;amd64)"
 var defaultUserAgent = fmt.Sprintf(
-	"%s/%s (%s)",
+	"%s %s/%s (%s)",
+	execName(),
 	version.ClientName,
 	version.ClientVersion,
 	strings.Join([]string{runtime.Version(), runtime.GOOS, runtime.GOARCH}, ";"),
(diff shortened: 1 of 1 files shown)


karaatanassov:
I considered adding OS version as reported by `uname -a` e.g. `5.10.175-6.ph4`. Unfortunately this requires complex golang code and is not supported by the 1.21 runtime. Some example of complexity include using unsafe POSIX API uname on Linux, using command line on Windows etc. These seem overkill for the concern we have and also may be issue when auditing/scanning the security of govmomi code.

vim25/soap/client.go
```diff
@@ -927,3 +930,16 @@ func (c *Client) DownloadFile(ctx context.Context, file string, u *url.URL, para
...
 	return c.WriteFile(ctx, file, rc, contentLength, param.Progress, param.Writer)
 }
+
+// execName gets the name of the executable for the current process
+func execName() string {
+	name, err := os.Executable()
+	if err != nil {
+		return "N/A"
+	}
+	_, file := filepath.Split(name)
```
dougm, line 940: could simplify this with:
```go
return strings.TrimSuffix(filepath.Base(name), ".exe")
```
And inline string (rather than `const`) is fine here too.

dougm (commented):
Thanks @karaatanassov , lgtm, just a minor suggestion to simplify

vim25/soap/client.go
karaatanassov: Thanks Doug! Yes code is much easier to read. Thank you.

dougm (approved):

PR #3383, opened 2024-03-08
fix: lookup client to use envoy sidecar url if enabled

ramineni:
currently, lookup client by default fetches the uri from "config.vpxd.sso.sts.uri" value if service content set . With envoy sidecar enabled, this MR is to first check for envoy sidecar url if present and use the same.

### DIFF
--- a/lookup/client.go
+++ b/lookup/client.go
@@ -23,6 +23,7 @@ import (
 	"log"
 	"net/url"
 
+	"github.com/vmware/govmomi/internal"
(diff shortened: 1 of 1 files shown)


ramineni:
cc @mayankbh

ramineni:
@dougm Thanks for the review. you are right. Fixed it. Thanks.

dougm (approved):
Thanks @ramineni

PR #3388, opened 2024-03-18
vcsim: add ExtensionManager and TaskManager.CreateTask support

dougm:
- needed an easier way to control Task state when testing, so added TaskManager.CreateTask and Task methods for updating state (wcpsvc also uses these methods)
- in real VC, CreateTask requires a taskID from an Extension, so went ahead and added basic ExtensionManager support too (and with this enabled govc extension bats)

### DIFF
--- a/govc/task/create.go
+++ b/govc/task/create.go
@@ -1,0 +1,100 @@
+/*
+Copyright (c) 2024-2024 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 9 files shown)


akutz (approved):
This lgtm, thanks! I do wonder if the commits should have something more in their bodies to indicate the use cases for these? I suppose we expect folks to know how to find the VMware docs for these APIs, and that is okay.

dougm:
> This lgtm, thanks! I do wonder if the commits should have something more in their bodies to indicate the use cases for these? I suppose we expect folks to know how to find the VMware docs for these APIs, and that is okay.

Thanks for reviewing @akutz . I added some use-case context in the description, will include more in the commit message next time.

PR #3389, opened 2024-03-18
vcsim: add TaskHistoryCollector support

### LINKED ISSUE #2567: Add task manager to `vcsim`
`vcsim` currently does not implement task manager funtionality.

Similar to the event manager capabilities in `vcsim`, add task manager.

n/a

**Additional context**
Useful for testing, e.g. govmomi/task

dougm:
refactor portions of 'EventHistoryCollector' to 'HistoryCollector' for use with 'TaskHistoryCollector'

TaskHistoryCollector provides a way to collect Task history beyond the 10 minute max of the 'RecentTask' property.
And unlike 'RecentTask', can filter based on Task state and collect tasks of child entities.

govc: add history flags to 'tasks' command

Closes #2567

### DIFF
--- a/govc/task/recent.go
+++ b/govc/task/recent.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2017-2023 VMware, Inc. All Rights Reserved.
+Copyright (c) 2017-2024 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -22,11 +22,15 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"time"
 
 	"github.com/vmware/govmomi/govc/cli"
 	"github.com/vmware/govmomi/govc/flags"
 	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/task"
 	"github.com/vmware/govmomi/view"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/methods"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 )
@@ -37,6 +41,13 @@ type recent struct {
 	max    int
 	follow bool
 	long   bool
+
+	state flags.StringList
+	begin time.Duration
+	end   time.Duration
+	r     bool
+
+	plain bool
 }
 
 func init() {
@@ -50,6 +61,10 @@ func (cmd *recent) Register(ctx context.Context, f *flag.FlagSet) {
 	f.IntVar(&cmd.max, "n", 25, "Output the last N tasks")
 	f.BoolVar(&cmd.follow, "f", false, "Follow recent task updates")
 	f.BoolVar(&cmd.long, "l", false, "Use long task description")
+	f.Var(&cmd.state, "s", "Task states")
+	f.DurationVar(&cmd.begin, "b", 0, "Begin time of task history")
+	f.DurationVar(&cmd.end, "e", 0, "End time of task history")
+	f.BoolVar(&cmd.r, "r", false, "Include child entities when PATH is specified")
 }
 
 func (cmd *recent) Description() string {
@@ -64,7 +79,11 @@ By default, all recent tasks are included (via TaskManager), but can be limited 
 to a specific inventory object.
 
 Examples:
-  govc tasks
+  govc tasks        # tasks completed within the past 10 minutes
+  govc tasks -b 24h # tasks completed within the past 24 hours
+  govc tasks -s queued -s running # incomplete tasks
+  govc tasks -s error -s success  # completed tasks
+  govc tasks -r /dc1/vm/Namespaces # tasks for VMs in this Folder only
   govc tasks -f
   govc tasks -f /dc1/host/cluster1`
 }
@@ -80,12 +99,16 @@ func (cmd *recent) Process(ctx context.Context) error {
 	return nil
 }
 
+// chop middle of s if len(s) > n
 func chop(s string, n int) string {
-	if len(s) < n {
+	diff := len(s) - n
+	if diff <= 0 {
 		return s
 	}
+	diff /= 2
+	m := len(s) / 2
 
-	return s[:n-1] + "*"
+	return s[:m-diff] + "*" + s[1+m+diff:]
 }
 
 // taskName describes the tasks similar to the ESX ui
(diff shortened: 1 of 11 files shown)


govc/task/recent.go
```diff
@@ -101,7 +124,112 @@ func taskName(info *types.TaskInfo) string {
...
+}
+
+// useRecent returns true if any options are specified that require use of TaskHistoryCollector
+func (cmd *recent) useRecent() bool {
+	return cmd.begin == 0 && cmd.end == 0 && !cmd.r && len(cmd.state) == 0
+}
+
+func (cmd *recent) newCollector(ctx context.Context, c *vim25.Client, ref *types.ManagedObjectReference) (collector, error) {
+	if cmd.useRecent() {
+		// original flavor of this command that uses `RecentTask` instead of `TaskHistoryCollector`
```
akutz, line 167: Should we document somewhere that the default behavior could fail if the number of allowed task collectors is exceeded? See https://vdc-download.vmware.com/vmwb-repository/dcr-public/184bb3ba-6fa8-4574-a767-d0c96e2a38f4/ba9422ef-405c-47dd-8553-e11b619185b2/SDK/vsphere-ws/docs/ReferenceGuide/vim.TaskManager.html#createCollector

govc/task/recent.go
```diff
@@ -101,7 +124,112 @@ func taskName(info *types.TaskInfo) string {
...
+	}
+
+	filter := types.TaskFilterSpec{
+		Entity: &types.TaskFilterSpecByEntity{
+			Entity:    *ref,
+			Recursion: r,
+		},
+		Time: &types.TaskFilterSpecByTime{
+			TimeType:  types.TaskFilterSpecTimeOptionStartedTime,
+			BeginTime: types.NewTime(now.Add(cmd.begin * -1)),
```
akutz, line 208: What does multiplying this by `-1` do? Same question for the end time below.

simulator/task_manager.go
```diff
@@ -43,29 +46,96 @@ func (m *TaskManager) init(r *Registry) {
 			m.Description = esx.Description
 		}
 	}
+
+	if m.MaxCollector == 0 {
+		m.MaxCollector = 1000
```
akutz, line 51: We should probably make this a well-documented constant.

simulator/task_manager.go
```diff
@@ -109,3 +179,220 @@ func (m *TaskManager) CreateTask(ctx *Context, req *types.CreateTask) soap.HasFa
...
+	}
+
+	walk(ctx.Map.Get(root), match)
+
+	return seen
+}
+
+// entityMatches returns true if the spec Entity filter matches the task entity.
+func (c *TaskHistoryCollector) entityMatches(ctx *Context, task *types.TaskInfo, spec *types.TaskFilterSpec) bool {
+	e := spec.Entity
```
akutz, line 224: If `spec` is `nil`, this will NPE. Can `spec` be `nil`?

simulator/task_manager.go
```diff
@@ -109,3 +179,220 @@ func (m *TaskManager) CreateTask(ctx *Context, req *types.CreateTask) soap.HasFa
...
+		return taskFilterChildren(ctx, e.Entity, task)
+	case types.TaskFilterSpecRecursionOptionAll:
+		return isSelf || taskFilterChildren(ctx, e.Entity, task)
+	}
+
+	return false
+}
+
+func (c *TaskHistoryCollector) stateMatches(_ *Context, task *types.TaskInfo, spec *types.TaskFilterSpec) bool {
+	if len(spec.State) == 0 {
```
akutz, line 244: Is this desired? If there are no specified specs, then match everything?

simulator/task_manager.go
```diff
@@ -109,3 +179,220 @@ func (m *TaskManager) CreateTask(ctx *Context, req *types.CreateTask) soap.HasFa
...
+		return true
+	}
+
+	for _, state := range spec.State {
```
akutz, line 248: Should the name of this function be `anyStateMatches` since any `state` from the filter that matches the task's state will cause `true` to be returned?

simulator/task_manager.go
```diff
@@ -109,3 +179,220 @@ func (m *TaskManager) CreateTask(ctx *Context, req *types.CreateTask) soap.HasFa
...
+
+// taskMatches returns true one of the filters matches the task.
+func (c *TaskHistoryCollector) taskMatches(ctx *Context, task *types.TaskInfo) bool {
+	spec := c.Filter.(types.TaskFilterSpec)
+
+	matchers := []func(*Context, *types.TaskInfo, *types.TaskFilterSpec) bool{
+		c.stateMatches,
+		c.timeMatches,
+		c.entityMatches,
+		// TODO: spec.UserName, etc
```
akutz, line 288: Should any of the fields that we do not yet consider be allowed when creating the filter? Should an InvalidState or ServerError error be returned from the simulator to prevent folks from expecting this filter configs to work?

simulator/task_manager_test.go
```diff
@@ -14,34 +14,167 @@ See the License for the specific language governing permissions and
...
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/task"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
 )
 
 func TestTaskManagerRecent(t *testing.T) {
```
akutz, line 32: Convert to an `Example` test pattern?

simulator/history_collector.go
```diff
@@ -0,0 +1,178 @@
...
+			return c.pos.Prev()
+		}
+		return c.page.Back()
+	}
+
+	c.read(max, next, take)
+}
+
+func pushHistory(l *list.List, item types.AnyType) {
+	if l.Len() > maxPageSize*5 {
```
akutz, line 162: Why `5`?

simulator/history_collector.go
```diff
@@ -0,0 +1,178 @@
...
+	l.PushBack(item)
+}
+
+func validatePageSize(count int32) (int, *soap.Fault) {
+	size := int(count)
+
+	if size == 0 {
+		size = defaultPageSize
+	} else if size < 0 || size > maxPageSize {
+		return -1, Fault("", &types.InvalidArgument{InvalidProperty: "maxCount"})
```
akutz, line 174: `&types.InvalidArgument{InvalidProperty: "maxCount"}` is used three times. Maybe `var errInvalidArgMaxCount` as a package level error?

akutz (commented):
Thanks @dougm !

govc/task/recent.go
dougm: Example, given `-b 1h`, we want to show tasks from the past hour.
https://pkg.go.dev/time#Time.Sub
> To compute t-d for a duration d, use t.Add(-d)

Tho reading that changed to `-cmd.begin` instead of `* -1`

govc/task/recent.go
dougm: We `Destroy()` the collector when the command is done. The default VC max is 32, same as EventHistoryCollector - used by the govc `events` command, have not had any reports of hitting the max.

simulator/history_collector.go
dougm: Good question. The `5` was added ~4 years ago in 5fd7e264 . Looking at real VC, seems it doesn't save more than maxPageSize (1000), so I changed to that.

simulator/history_collector.go
dougm: Yes, done.

simulator/task_manager.go
dougm: Added a `var maxCollectors`. Looks like VC's default is 32 and can be changed w/ OptionManager, but don't have a use-case to support that yet.

simulator/task_manager.go
dougm: The `spec` can't be nil, changed the param

simulator/task_manager.go
dougm: Yes, right. When `spec.State` _is_ specified, you can limit to tasks of state `failed` for example.

simulator/task_manager.go
dougm: I just chose `${fieldName}Matches`, same pattern as EventManager. If `any` makes it more clear, we should add the prefix to all of the *Matches functions, and for EventManager too. Lemme know what you think.

simulator/task_manager.go
dougm: Have the same TODO in EventManager.. it probably isn't much more work to implement the matchers, than to figure out which spec fields to error on. Just kept it to the common use-cases for now.

simulator/task_manager_test.go
dougm: There is a simpler example here, which now runs against vcsim with this PR: https://github.com/vmware/govmomi/blob/main/examples/tasks/main.go

simulator/task_manager.go
akutz: Should it be a constant that we can adjust as needed? Is there any reason it _should_ be a variable?

simulator/task_manager.go
akutz: I do not feel strongly enough to want to refactor existing names for another manager. Thanks!

simulator/task_manager.go
dougm: If we end up with a use-case to support changing the max via OptionManager, then we'd need a variable. But I just clumped it into the existing set of non-exported `var (...)`, do we gain anything by making it a constant?

dougm:
Thanks @akutz !

PR #3391, opened 2024-03-19
fix: endpoint url to bypass lookup service if envoy sidecar enabled

ramineni:
EndpointURL returns FQDN:443 as endpoint url even if envoy sidecar enabled. This MR updates to bypass lookup service if envoy sidecar enabled.

## Description

Please include a summary of the change and which issue is fixed. Please also
include relevant motivation and context. List any dependencies that are required
for this change.

Closes: #(issue-number)

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

- [ ] Test Description 1
- [ ] Test Description 2

## Checklist:

- [ ] My code follows the `CONTRIBUTION` [guidelines] of this project

### DIFF
--- a/lookup/client.go
+++ b/lookup/client.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/base64"
+	"fmt"
 	"log"
 	"net/url"
 
@@ -119,6 +120,10 @@ func (c *Client) SiteID(ctx context.Context) (string, error) {
 // If the endpoint is found, its TLS certificate is also added to the vim25.Client's trusted host thumbprints.
 // If the Lookup Service is not available, the given path is returned as the default.
 func EndpointURL(ctx context.Context, c *vim25.Client, path string, filter *types.LookupServiceRegistrationFilter) string {
+	// Services running on vCenter can bypass lookup service.
+	if useSidecar := internal.UsingEnvoySidecar(c); useSidecar {
+		return fmt.Sprintf("http://%s%s", c.URL().Host, path)
+	}
 	if lu, err := NewClient(ctx, c); err == nil {
 		info, _ := lu.List(ctx, filter)
 		if len(info) != 0 && len(info[0].ServiceEndpoints) != 0 {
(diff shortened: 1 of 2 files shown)


ramineni:
cc @dougm @mayankbh

lookup/client.go
```diff
@@ -119,6 +120,11 @@ func (c *Client) SiteID(ctx context.Context) (string, error) {
 // If the endpoint is found, its TLS certificate is also added to the vim25.Client's trusted host thumbprints.
 // If the Lookup Service is not available, the given path is returned as the default.
 func EndpointURL(ctx context.Context, c *vim25.Client, path string, filter *types.LookupServiceRegistrationFilter) string {
+	// Services running on vCenter can bypass lookup service.
+	if useSidecar := internal.UsingEnvoySidecar(c); useSidecar {
+		return fmt.Sprintf("http://%s%s", c.URL().Host, path)
```
mayankbh, line 125: Not sure I follow. Wouldn't this mean we always return the 'base' address of the Envoy sidecar rather than the service specific route? (which is only known around line 131 I think)

(It seems like it would be easier to verify with a test)

ramineni: sorry, didnt understand. 
path passed above to the function, for suppose for licensing service is /ls/sdk (https://gitlab.eng.vmware.com/core-build/govmomi-internal/-/blob/master/ls/client.go#L37) , we are appending the same to 'base' address .

L131, in this case returns , https://sc1-10-78-227-166.eng.vmware.com:443/ls/sdk (from the logs)

> (It seems like it would be easier to verify with a test)
you meant to add unit test or any other way?

mayankbh: Ah, misunderstood. So callers can pass in the 'default' path and have it returned if the Envoy sidecar is in use _or_ if lookup service is broken after this change.

ramineni: yes, if lookup service is broken and not using envoy sidecar, it still follows the original flow and just return the path

dougm (approved):
lgtm, thanks @ramineni

lookup/client.go
mayankbh: Okay, makes sense, thanks for clarifying.

PR #3410, opened 2024-04-12
govc: add default output formatting for namespace commands

dougm:
Including a few other minor changes for the new set of namespace commands.

### DIFF
--- a/govc/namespace/info.go
+++ b/govc/namespace/info.go
@@ -19,7 +19,10 @@ package namespace
 import (
 	"context"
 	"flag"
+	"fmt"
 	"io"
+	"strings"
+	"text/tabwriter"
 
 	"github.com/vmware/govmomi/vapi/namespace"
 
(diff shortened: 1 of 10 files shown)


vapi/namespace/simulator/simulator.go
```diff
@@ -281,7 +281,6 @@ func (h *Handler) namespaces(w http.ResponseWriter, r *http.Request) {
 		if len(subpath) > 0 {
 			if entry, contains := namespacesMap[subpath]; contains {
 				var spec namespace.NamespacesInstanceUpdateSpec
-				// If vapi.Decode fails it sets the status to bad request
```
dougm, line 284: note: removed these comments since it's included in the function doc (and we use vapi.Decode this way all over)

stoyanzhelyazkov (approved):

govc/namespace/info.go
```diff
@@ -65,27 +74,28 @@ func (cmd *info) Usage() string {
...
 
 	nm := namespace.NewManager(rc)
 
 	d, err := nm.GetNamespace(ctx, f.Arg(0))
-
 	if err != nil {
 		return err
 	}
 
-	cmd.JSON = !cmd.All()
```
stoyanzhelyazkov, line 89: Just FYI - I did the same thing on another PR over a month ago https://github.com/vmware/govmomi/pull/3368
I'm saying this in case you're working on eliminating the "only-JSON-output-supported" pattern :)

dougm: I remember, but I've yet to use vLCM (directly at least). I use vSphere Namespaces everyday and am happy to replace stuff like below with the new govc commands, thanks!

```bash
# now govc namespace.create
govc session.login -r -X POST /api/vcenter/namespaces/instances \
  <<<'{"cluster": "domain-c9","namespace": "velero"}'

# now govc namespace.ls
govc session.login -r -X GET /api/vcenter/namespaces/instances | jq -r .[].namespace
```

Issue #3417, opened 2024-04-18
[BUG] WaitEx omits updates when the server GUID is not set


stoyanzhelyazkov:
**Describe the bug**

It is possible for Managed Object References to be returned with an empty string for the `ServerGUID` field.
I am not sure whether this is a problem with the SOAP response serialization or if its just they way some of the VC APIs work.

In any case the new `WaitEx` method filters incoming updates by matching the entire Managed Object Reference.
This means that if the `ServerGUID` field is an empty string the entire update is filtered out.

This is the exact line where the the filtering occurs
https://github.com/vmware/govmomi/blob/main/task/wait.go#L144

`WaitEx` was introduced in https://github.com/vmware/govmomi/pull/3331
The original Wait method did not have the MOR comparison

**To Reproduce**

The bug is easily reproducibly by using govc

1. Create a virtual disk on a datastore 
2. Execute `datastore.disk.info`
3. Observe that the execution hangs up

Reverting https://github.com/vmware/govmomi/pull/3331 or modifying the MOR comparison to just check the Value field
resolves the problem.

**Expected behavior**

WaitEx should not drop valid updates and should not cause the execution to hang for the duration of the default timeout

**Affected version**

Latest code from the main branch

**Screenshots/Debug Output**
N/A

**Additional context**

This issue is the root cause for https://github.com/hashicorp/terraform-provider-vsphere/issues/2162

stoyanzhelyazkov:
@akutz I'd appreciate your advice on how to fix this.
https://github.com/vmware/govmomi/pull/3331 is quite complex

PR #3418, opened 2024-04-18
bugfix: Exclude ServerGUID from MOR comparison in WaitEx

stoyanzhelyazkov:
## Description

In rare cases it is possible for the ServerGUID field to be empty on managed objects.

One way to observe this is to run `queryVirtualDiskInfo` on the VC. The equivalent `govc` command is `datastore.disk.info`

WaitEx compares the entire ManagedObjectReference structure when filtering updates and if the ServerGUID is unset on the update object the data is dropped.

Since the property collector is scoped to a single VC the server GUID shouldn't even matte so we can safely ignore when filtering.

Closes: #3417

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

I set up a VMDK as described in https://github.com/hashicorp/terraform-provider-vsphere/issues/2162
on a 7.0.3 vCenter

I observed that the ServerGUID is indeed empty and that I am unable to retrieve the data for the virtual disk.

After applying the code changes I was able to query the virtual disk.

Also ran `make test` and `make fix`

## Checklist:

### DIFF
--- a/task/wait.go
+++ b/task/wait.go
@@ -141,7 +141,7 @@ func WaitEx(
 		func(updates []types.ObjectUpdate) bool {
 			for _, update := range updates {
 				// Only look at updates for the expected task object.
-				if update.Obj == ref {
+				if update.Obj.Value == ref.Value && update.Obj.Type == ref.Type {
 					if cb.fn(update.ChangeSet) {
 						return true
 					}


akutz (approved):
Lgtm, thanks!

PR #3426, opened 2024-04-24
api: Generate enum consts with Values funcs

akutz:
## Description

This patch ensures enum types are generated with Values functions that make it trivial to discover all values for a given enum.

@dougm, I am not sure if this is something we want or not. I propose it to make it simpler for projects like VM Operator to generate a list of all constant values for enums like GuestOSID. I checked, and while it is possible to use unsafe code to list all types in a given package, Go's compiler makes it impossible to list all exported constants at the package level. Other than AST with source analysis, this is the only way for a downstream project to get a list of all an "enum"'s values. The question is -- is it worth the increase in memory?

cc @dilyar85 as I was doing this so you could create a `pkg/gen/guestosids/guestosids.go` file in VM Operator that is effectively a `main.go` and could be used with `//go:generate` in the `./api/v1alpha3/virtualmachine_types.go` file to automatically generate a type with the correct kubebuilder comments for an enum. 

Closes: `NA`

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

`NA`

## Checklist:


dougm (approved):
Thanks @akutz , very nice. This would also be quite useful within govmomi, we can replace this: https://github.com/vmware/govmomi/blob/main/simulator/guest_id.sh#L29

And lists of enum consts are sprinkled all over govc, e.g. https://github.com/vmware/govmomi/blob/ae5fd928c9958968b1b4055edbd09e904f76860c/govc/cluster/change.go#L30-L34

akutz:
> Thanks @akutz , very nice. This would also be quite useful within govmomi, we can replace this: https://github.com/vmware/govmomi/blob/main/simulator/guest_id.sh#L29
> 
> And lists of enum consts are sprinkled all over govc, e.g.
> 
> https://github.com/vmware/govmomi/blob/ae5fd928c9958968b1b4055edbd09e904f76860c/govc/cluster/change.go#L30-L34

I have replaced the aforementioned `guest_id.sh` with the new work.

dougm (approved):
Awesome, thanks @akutz

Issue #3430, opened 2024-05-02
Ability to create file-share volumes using vcsim


vdkotkar:
**Is your feature request related to a problem? Please describe.**
We are using vcsim in vSphere CSI plugin for writing unit tests. We are able to create block volumes using vcsim. I wanted to check if we can create file-share (RWX) volumes using vcsim.
For creating file share volumes, we require compatible datastores to create such volume. And it checks for following things while getting compatible datastores:
1. It checks for ClusterComputeResources where user has “Host.Config.Storage” privilege on the cluster.
2. Cluster should be “vsan” cluster and “FileServiceConfig” should be enabled on the cluster.
3. It then gets all datastores belonging to clusters satisfying above conditions. And if datastore type is “vsan”, then it is considered eligible for creating a file-share volume.

Is it possible to simulate this case using VC simulator? If currently it is not possible, then can we enhance simulator to support creating file-share volumes?

**Describe the solution you'd like**
I want to write unit tests in vSphere CSI plugin which should be able to create file-share volumes using VC simulator in govmomi.

**Describe alternatives you've considered**

**Additional context**

dougm:
#3432 has part of what you'd need. vcsim doesn't have built-in support for vsan Datastore type, but can also look at adding that.. I actually have an old incomplete git stash, but haven't had a change to see where I left off.

PR #3432, opened 2024-05-05
vcsim: add vSAN FileServiceConfig support

dougm:
govc: add vsan.info -file-service-enabled flag

Issue #3430


akutz (approved):
Nice, thanks Doug!

PR #3435, opened 2024-05-08
feat: Bindings for vSphere Zone APIs

spacegospod:
## Description

Adding bindings for the [consumption domains API](https://developer.broadcom.com/xapis/vsphere-automation-api/latest/vcenter/consumption_domains/) (a.k.a. vSphere Zones)

Related: https://github.com/hashicorp/terraform-provider-vsphere/issues/2185

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

Tested against a live VC.
Created a simulator for the new endpoints and added a unit test that verifies the contract.

## Checklist:

### DIFF
--- a/vapi/vcenter/consumptiondomains/associations/associations.go
+++ b/vapi/vcenter/consumptiondomains/associations/associations.go
@@ -1,0 +1,86 @@
+/*
+Copyright (c) 2024-2024 VMware, Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package associations
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/vmware/govmomi/vapi/rest"
+)
+
+const (
+	basePath         = "/api/vcenter/consumption-domains/zones/cluster"
+	associationsPath = basePath + "/%s/associations"
+)
+
+// Manager extends rest.Client, adding vSphere Zone association related methods.
+type Manager struct {
+	*rest.Client
+}
+
+// NewManager creates a new Manager instance with the given client.
+func NewManager(client *rest.Client) *Manager {
+	return &Manager{
+		Client: client,
+	}
+}
+
+// Status
+// https://developer.broadcom.com/xapis/vsphere-automation-api/latest/vcenter/data-structures/ConsumptionDomains_Zones_Cluster_Associations_Status
+type Status struct {
+	Success bool `json:"success"`
+}
+
+// AddAssociations
+// https://developer.broadcom.com/xapis/vsphere-automation-api/latest/vcenter/api/vcenter/consumption-domains/zones/cluster/zone/associations__action=add/post
+func (c *Manager) AddAssociations(zone string, cluster string) error {
+	path := c.Resource(fmt.Sprintf(associationsPath, zone)).WithParam("action", "add")
+	req := path.Request(http.MethodPost, []string{cluster})
+	var res Status
+	if err := c.Do(context.Background(), req, &res); err != nil {
+		return err
+	}
+
+	// This endpoint does not always return and error upon failure.
+	// In such cases we need to parse the response to figure out the actual status.
+
+	if !res.Success {
+		return errors.New("unable to add associations")
+	}
+
+	return nil
+}
+
+// RemoveAssociations
+// https://developer.broadcom.com/xapis/vsphere-automation-api/latest/vcenter/api/vcenter/consumption-domains/zones/cluster/zone/associations__action=remove/post/
+func (c *Manager) RemoveAssociations(zone string, cluster string) error {
+	path := c.Resource(fmt.Sprintf(associationsPath, zone)).WithParam("action", "remove")
+	req := path.Request(http.MethodPost, []string{cluster})
+	return c.Do(context.Background(), req, nil)
+}
+
+// GetAssociations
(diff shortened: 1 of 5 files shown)


vapi/vcenter/consumptiondomains/simulator/simulator.go
```diff
@@ -0,0 +1,197 @@
...
+		vapi.ApiErrorNotFound(w)
+	}
+}
+
+func (h *Handler) associations(w http.ResponseWriter, r *http.Request) {
+	subpath := r.URL.Path[len(associationsPath)+1:]
+	pathParts := strings.Split(subpath, "/")
+
+	if len(pathParts) != 2 || pathParts[1] != "associations" {
+		vapi.ApiErrorNotFound(w)
```
dougm, line 132: Looks like you need a `return` here too?

vapi/vcenter/consumptiondomains/simulator/simulator.go
```diff
@@ -0,0 +1,197 @@
...
+
+		action := queryParts[1]
+
+		var clusterIds []string
+		if !vapi.Decode(r, w, &clusterIds) {
+			vapi.ApiErrorGeneral(w)
+			return
+		}
+
+		if action == "add" {
```
dougm, line 159: Minor, but find switch easier to read + extend than if/else/return
```go
	switch r.URL.Query().Get("action") {
	case "add":
	case "remove":
	default:
	}
```

vapi/vcenter/consumptiondomains/associations/associations.go
```diff
@@ -0,0 +1,82 @@
...
+		Client: client,
+	}
+}
+
+// AddAssociations
+// https://developer.broadcom.com/xapis/vsphere-automation-api/latest/vcenter/api/vcenter/consumption-domains/zones/cluster/zone/associations__action=add/post
+func (c *Manager) AddAssociations(zone string, clusters []string) error {
+	path := c.Resource(fmt.Sprintf(associationsPath, zone)).WithParam("action", "add")
+	req := path.Request(http.MethodPost, clusters)
+	var res map[string]interface{}
```
dougm, line 50: The response [here](https://developer.broadcom.com/xapis/vsphere-automation-api/latest/vcenter/data-structures/ConsumptionDomains_Zones_Cluster_Associations_Status/) has 2 fields. You may want to consider a Status type instead:
```go
// Exception includes error detail for Status responses.
type Exception struct {
	ErrorType string                    `json:"error_type"`
	Messages  []rest.LocalizableMessage `json:"messages,omitempty"`
}

// Status contains information about the outcome of the operation to add a set of
// vSphere clusters to a zone or the operation to remove a set of vSphere clusters
// from a zone.
type Status struct {
	// This is true if the complete operation succeeded without any errors.
	Success bool `json:"success"`
	// FailedClusters represents the associations between the identifiers of the clusters
	// for which the operation failed and the reason of the failure.
	FailedClusters map[string]Exception `json:"failed_clusters,omitempty"`
}
```

vapi/vcenter/consumptiondomains/associations/associations.go
```diff
@@ -0,0 +1,82 @@
...
+	if err := c.Do(context.Background(), req, &res); err != nil {
+		return err
+	}
+
+	// This endpoint does not always return and error upon failure.
+	// In such cases we need to parse the response to figure out the actual status.
+
+	success := res["success"].(bool)
```
dougm, line 58: This would panic if `success` isn't in the map, can prevent that:
```go
	success, ok := res["success"].(bool)
	if !(ok && success) {
		return errors.New(...)
	}
```

dougm (commented):
Thanks @spacegospod , a few suggestions

vapi/vcenter/consumptiondomains/associations/associations.go
spacegospod: I know about the second field but I thought I'd keep it simpler and just report the overall status.
I'll follow your advice and add them to the return type since I'll have to revise the diff anyway

spacegospod: I re-tested this API

First of all it does not allow for a zone to be associated with more than one cluster. Even the UI uses a radio button that allows single selection.
The bigger problem is that if you provide an incorrect cluster identifier it responds with a message that says the user is not authorized. This is wrong and I will probably log a bug internally.

What I'm going to do is keep the reeturn type and I will change the method signature to accept a single cluster ID instead of an array

vapi/vcenter/consumptiondomains/simulator/simulator.go
spacegospod: agreed

vapi/vcenter/consumptiondomains/associations/associations.go
spacegospod: thanks, fixed

vapi/vcenter/consumptiondomains/associations/associations.go
dougm: What about the failed_clusters field, does that not capture some detail when Status == false ?

Multiple clusters might be part of a future enhancement. In which case you could do:
```patch
modified   vapi/vcenter/consumptiondomains/associations/associations.go
@@ -50,9 +50,9 @@ type Status struct {
 
 // AddAssociations
 // https://developer.broadcom.com/xapis/vsphere-automation-api/latest/vcenter/api/vcenter/consumption-domains/zones/cluster/zone/associations__action=add/post
-func (c *Manager) AddAssociations(zone string, cluster string) error {
+func (c *Manager) AddAssociations(zone string, cluster ...string) error {
 	path := c.Resource(fmt.Sprintf(associationsPath, zone)).WithParam("action", "add")
-	req := path.Request(http.MethodPost, []string{cluster})
+	req := path.Request(http.MethodPost, cluster)
 	var res Status
 	if err := c.Do(context.Background(), req, &res); err != nil {
 		return err
```

dougm (approved):
lgtm, thanks @spacegospod !

vapi/vcenter/consumptiondomains/associations/associations.go
spacegospod: I'd be more than happy to add support for multiple clusters when that is an official feature.
IMO it's best if we provide the same interface as the VC UI which also allows for single-cluster assignment.

As for failed clusters - I mentioned that the API is broken and I don't feel comfortable discussing internal bugs on github. The structure for "failed_clusters" is also undocumented, I can only base my bindings on the buggy response that I get today

PR #3451, opened 2024-05-28
vcsim: add PropertyCollector index support

dougm:
PropertyCollector supports use of index (unique key) to collect a single array element. Further, a property path may also be applied to the given element.

### DIFF
--- a/govc/test/object.bats
+++ b/govc/test/object.bats
@@ -411,6 +411,66 @@ EOF
   govc object.collect -O -json | jq .
 }
 
+@test "object.collect index" {
+  vcsim_env
+
+  export GOVC_VM=/DC0/vm/DC0_H0_VM0
+
+  # NOTE: '-o' flag uses RetrievePropertiesEx() and mo.ObjectContentToType()
+  # By default, WaitForUpdatesEx() is used with raw types.ObjectContent
+
+  run govc object.collect -o $GOVC_VM 'config.hardware[4000]'
+  assert_failure
+
+  run govc object.collect -o $GOVC_VM 'config.hardware.device[4000'
+  assert_failure
+
+  run govc object.collect -o $GOVC_VM 'config.hardware.device["4000"]'
+  assert_failure # Key is int, not string
+
+  run govc object.collect -o -json $GOVC_VM 'config.hardware.device[4000]'
+  assert_success
+
+  run jq -r .config.hardware.device[].deviceInfo.label <<<"$output"
+  assert_success ethernet-0
+
+  run govc object.collect -o $GOVC_VM 'config.hardware.device[4000].enoent'
+  assert_failure # InvalidProperty
+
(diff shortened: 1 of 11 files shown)


govc/test/object.bats
```diff
@@ -411,6 +411,66 @@ EOF
...
+  run govc object.collect -o $GOVC_VM 'config["guestinfo.a"]'
+  assert_failure
+
+  run govc object.collect -o $GOVC_VM 'config.extraConfig["guestinfo.a"]'
+  assert_success # Key does not exist, not an error
+
+  run govc vm.change -e "guestinfo.a=1" -e "guestinfo.b=2"
+  assert_success
+
+  run govc object.collect -json $GOVC_VM 'config.extraConfig["guestinfo.b"]'
```
hickeng, line 458: If you have `guestinfo.abc=3` does that match (ie. substring match) or is it doing an exact token match between the separators?

dougm: It's exact afaik, equal to the `Key` field of the array type.

simulator/property_collector.go
```diff
@@ -214,6 +218,64 @@ func fieldValue(rval reflect.Value, p string) (interface{}, error) {
...
+			item = item.Elem()
+		}
+		if item.Kind() == reflect.Ptr {
+			item = item.Elem()
+		}
+		if item.Kind() != reflect.Struct {
+			return reflect.Value{}, errInvalidField
+		}
+
+		field := item.FieldByName("Key")
```
akutz, line 240: I am having trouble tracing how we get here, so maybe this is a non-issue, but what if the property path is for an array of non-struct types, such as an array of integers or strings? I believe a numerical index should work in that case.

dougm: afaik, property path array index can only be used when struct has a `key` field, see for example:

```console
% govc object.collect $GOVC_VM disabledMethod
disabledMethod  []string  MakePrimaryVM_Task,TerminateFaultTolerantVM_Task,...

% govc object.collect $GOVC_VM 'disabledMethod[1]'
govc: ServerFaultCode: InvalidProperty
```

Another case, AuthorizationRole has no `key` field. And while it does have a unique `roleId` field, same `InvalidProperty` result when using a valid `roleId`
```console
% id=$(govc object.collect -o -json AuthorizationManager:AuthorizationManager roleList | jq .roleList[].roleId | head -1)

% govc object.collect AuthorizationManager:AuthorizationManager "roleList[$id]"
govc: ServerFaultCode: InvalidProperty
```

This [doc](https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-web-services-programming-guide/GUID-DB836CA6-8316-4B2F-8CE5-8AD933F8DE36.html) also seems to indicate must be key-based:

> Nested properties can also refer to entries in a key-based array

aruneshpa (approved):

PR #3461, opened 2024-06-03
nfc: Record the SSL thumbprint for the host the (device) URL refers to

HakanSunay:
## Description

With this change we are recording the SSL thumbprint for the host the device URL refers to. This is done so that clients of the nfc wrapper can perform their own SSL verification using the recorded thumbprint if necessary. The thumbprint is stored in the FileItem struct, but the New constructor is not modified so as not to introduce any breaking changes.

Please note that the thumbprint is recorded only for the import case (when deploying an OVF in my case) since that is the only necessary path for our internal usage.

Closes: `NA`

## Type of change

  not work as expected)
- [ ] Build related change

## How Has This Been Tested?

I have performed isolated testing against both a vCenter and a ESXi host. I have observed that the thumbprint is recorded and accessible through the `FileItem` struct as expected.

## Checklist:

### DIFF
--- a/nfc/lease.go
+++ b/nfc/lease.go
@@ -145,7 +145,11 @@ func (l *Lease) newLeaseInfo(li *types.HttpNfcLeaseInfo, items []types.OvfFileIt
 		// this is an import
 		for _, item := range items {
 			if device.ImportKey == item.DeviceId {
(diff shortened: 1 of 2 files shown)


dougm (approved):
lgtm

PR #3462, opened 2024-06-03
vcsim: add library item storage API support

### LINKED ISSUE #3213: Allow the usage of .iso files from a Content Library with device.cdrom.insert

In a previous issue found [here](https://github.com/vmware/govmomi/issues/1697) the desired outcome was discussed and proposed, but a [hacky solution](https://github.com/vmware/govmomi/blob/main/govc/USAGE.md#libraryinfo) was chosen (`govc device.cdrom.insert -vm $vm -device cdrom-3000 $(govc library.info -L /lib1/item1/file1)`) and the issue was closed. The solution previously proposed, which was to use library.info -L to get the path does not work for me (which is also the example used with the --help of the command) to mount a .iso from a content library through device.cdrom.insert. This method results in a no media found screen, the file referenced exists, and can be queried through a library.info call with the full path.

I would like a solution that can reference the .iso file directly within the device.cdrom.insert command, without making a separate call to the datastore to get its path (this call doesn't work in my case). There were excellent proposed ideas in the above linked issue, but in the end I would like it to use the Content Library ISO File native feature.

I am not entirely sure what alternatives there are, I have tried the officially stated method to get a iso mounted through vm.create and then device.cdrom.insert which just ends in a no media found screen on boot. And I am not able to figure out what was meant in the last comment on the issue, so his [workaround](https://github.com/vmware/govmomi/issues/1697#issuecomment-716825220) is unknown, I assume he somehow found the absolute path to the datastore itself, and not the datastore path of the .iso file... which I have not been able to find through govc commands.

**Additional context**

N/A

dougm:
govc: support Content Library iso files for cdrom backing

The 'device.cdrom.insert' and 'vm.create -iso' commands can now a URL with 'library' scheme to back a cdrom device from a Content Library.

api: refactor 'library.info -L' functionality into library PathFinder helper.

Fixes #3213

### DIFF
--- a/govc/library/info.go
+++ b/govc/library/info.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2018-2023 VMware, Inc. All Rights Reserved.
+Copyright (c) 2018-2024 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -22,8 +22,6 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"path"
-	"strings"
 	"text/tabwriter"
 	"time"
 
@@ -33,10 +31,6 @@ import (
 	"github.com/vmware/govmomi/units"
 	"github.com/vmware/govmomi/vapi/library"
 	"github.com/vmware/govmomi/vapi/library/finder"
-
-	"github.com/vmware/govmomi/property"
-	"github.com/vmware/govmomi/vim25/mo"
-	"github.com/vmware/govmomi/vim25/types"
 )
 
 type info struct {
@@ -47,8 +41,9 @@ type info struct {
 	long bool
 	link bool
 	url  bool
+	stor bool
 
-	names map[string]string
+	pathFinder *finder.PathFinder
 }
 
 func init() {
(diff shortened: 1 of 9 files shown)


govc/library/info.go
```diff
@@ -66,8 +61,7 @@ func (cmd *info) Register(ctx context.Context, f *flag.FlagSet) {
 	f.BoolVar(&cmd.long, "l", false, "Long listing format")
 	f.BoolVar(&cmd.link, "L", false, "List Datastore path only")
 	f.BoolVar(&cmd.url, "U", false, "List pub/sub URL(s) only")
-
-	cmd.names = make(map[string]string)
+	f.BoolVar(&cmd.stor, "s", false, "Include file item storage details")
```
dilyar85, line 64: It appears that this arg only takes effect when the cmd is called with a specific library item file (e.g. `library.info -s <library-name>` or `library.info -s <library-name>/<item-name>` would not contain any storage details). Should we clarify this in the help message to avoid confusion?

vapi/library/library_item_storage.go
```diff
@@ -43,11 +43,11 @@ func (c *Manager) ListLibraryItemStorage(ctx context.Context, id string) ([]Stor
 }
 
 // GetLibraryItemStorage returns the storage for a specific file in a library item.
-func (c *Manager) GetLibraryItemStorage(ctx context.Context, id, fileName string) (*Storage, error) {
+func (c *Manager) GetLibraryItemStorage(ctx context.Context, id, fileName string) ([]Storage, error) {
 	url := c.Resource(internal.LibraryItemStoragePath).WithID(id).WithAction("get")
 	spec := struct {
-		Name string `json:"name"`
+		Name string `json:"file_name"`
```
dilyar85, line 49: Huh, I didn't notice that the GET API requires `file_name` here. I guess this didn't get caught since the image-registry operator always uses the LIST API call. Thanks for fixing this!

dilyar85 (approved):
LGTM!

govc/library/info.go
dougm: Right, yes, updated the doc

PR #3424, opened 2024-04-23
vcsim: Fix:#3423 - mandated to pass the new disk size in Bytes and KB…

Davinder2me:
…. If either of value is 0, ReconfigVM_Task is throwing *types.InvalidDeviceOperation

Closes: #3423

## Description

Fix the new disk size check. 

Closes: #(issue-number)

## Type of change

Please mark options that are relevant:

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [ ] This change requires a documentation update
- [ ] Build related change

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. If applicable, please also list any relevant
details for your test configuration.

- [X] Deployed VM using vRA using vCSIM.  And Vm provisioning was successful. ReconfigureVM_task successful.

## Checklist:

- [X] My code follows the `CONTRIBUTION` [guidelines] of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged

[guidelines]: https://github.com/vmware/govmomi/blob/main/CONTRIBUTING.md

PR #3465, opened 2024-06-07
fix: govc vm.disk.change should only modify CapacityInBytes

### LINKED ISSUE #3464: [BUG] Can't resize a VM's disk
**Describe the bug**

Since version v0.37.2, when trying to change a disk's size, I get the error :
```
$ ./govc vm.disk.change -vm VMNAME -disk.filePath '[DS] PATH/VM.vmdk' -size 29G
./govc: error resizing main disk
Logged Item:  Invalid operation for device '0'.
```

**To Reproduce**

1. Create a VM with a disk attached
2. Try to expand the disk using `govc vm.disk.change -vm VMNAME -disk.filePath '[DS] PATH/VM.vmdk' -size XXG`

**Expected behavior**
The disk should be updated with the new size without any error message, same as in v0.37.1

**Affected version**
Works in v0.37.1. It broke in v0.37.2

dougm:
PR #3424 switched from the deprecated CapacityInKB field to CapacityInBytes. That PR focus was to fix an issue with the simulator, but broke against real vCenter. Just leaving CapacityInKB as-is makes vCenter happy again.

Fixes #3464

### DIFF
--- a/govc/vm/disk/change.go
+++ b/govc/vm/disk/change.go
@@ -154,7 +154,6 @@ func (cmd *change) Run(ctx context.Context, f *flag.FlagSet) error {
 
 	if int64(cmd.bytes) != 0 {
 		editdisk.CapacityInBytes = int64(cmd.bytes)
-		editdisk.CapacityInKB = int64(0) // zero deprecated field
 	}
 
 	if editdisk.StorageIOAllocation == nil {


davinderkvnera (approved):
Thanks.

PR #3466, opened 2024-06-11
Add storage profile related api and govc helpers

dougm:
api: add pbm.DatastoreMap for finding compatible Datastores
- add pbm.ProfileMap for finding storage profiles by name

govc: add vm.create '-profile' and '-place' flags

api: add storage_specs field to Namespace related structures

govc: add '-storage' flag to namespace.create and namespace.update commands
- The '-cluster', '-library' and '-storage' flags can be specified by name, rather than ID
- The namespace.info command outputs cluster, library and storage by name, rather than ID

### DIFF
--- a/pbm/pbm_util.go
+++ b/pbm/pbm_util.go
@@ -1,11 +1,11 @@
 /*
-Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+Copyright (c) 2017-2024 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


dougm:
Example of using `-place` to see which Datastore is recommended:
```console
% govc vm.create -place -net "VM Network" -cluster $cluster -iso library:/govc/ubuntu-14.04.6-server-amd64/*.iso ubuntu
Target:       /test-vpx-1717673857-26202-wcp.wcp-sanity/host/test-vpx-1717673857-26202-wcp.wcp-sanity-cluster
  Datastore:  /test-vpx-1717673857-26202-wcp.wcp-sanity/datastore/sharedVmfs-0
  Pool:       /test-vpx-1717673857-26202-wcp.wcp-sanity/host/test-vpx-1717673857-26202-wcp.wcp-sanity-cluster/Resources
  Host:       /test-vpx-1717673857-26202-wcp.wcp-sanity/host/test-vpx-1717673857-26202-wcp.wcp-sanity-cluster/10.182.172.85
```
Note that adding `-xml` or `-dump` flags will output the complete spec input and recommendations response.

pbm/pbm_util.go
```diff
@@ -146,3 +152,99 @@ func verifyPropertyValueIsBoolean(propertyValue string, dataType string) (bool,
...
+func (c *Client) ProfileMap(ctx context.Context, uid ...string) (*ProfileMap, error) {
+	m := &ProfileMap{Name: make(map[string]types.BasePbmProfile)}
+
+	rtype := types.PbmProfileResourceType{
+		ResourceType: string(types.PbmProfileResourceTypeEnumSTORAGE),
+	}
+
+	category := types.PbmProfileCategoryEnumREQUIREMENT
+
+	ids := make([]types.PbmProfileId, len(uid))
```
dilyar85, line 172: Nit: Use `var ids []types.PbmProfileId` here and init the slice in the "else" statement to avoid unnecessary allocation when `len(uid) == 0`.

dilyar85 (approved):
LGTM with a minor comment left. Thanks again for adding these supports in both api and govc!

pbm/pbm_util.go
dougm: I considered that and opted to have 1 less line of code.. but I agree this pattern/practice is better in general, updated.

dilyar85 (approved):
LGTM!

PR #3476, opened 2024-06-24
fix: xml marshal byte array fields as vCenter does

### LINKED ISSUE #1977: QueryTpmAttestationReport() does not get the proper PCR digest
I am providing a sample code that I have written, 

```
query := types.QueryTpmAttestationReport{This: host.Reference()}
attestationReport, err := methods.QueryTpmAttestationReport(ctx, vmwareClient.RoundTripper, &query)
tpmPCRValues := tpmAttestationReport.Returnval.TpmPcrValues[0]
fmt.Printf("VMware host attestation report : %v\n", tpmPCRValues)
```

Although I see that it is a SHA 256 digest, I am getting just 3 bytes as part of the digest value

Sample Output,

`{{{} SHA256 [45 57 48] } 0}`

I am expecting a 32 byte digest value in case the digest type is SHA 256. Am I missing something ?

### LINKED ISSUE #3469: [BUG] PropertyCollector not populating ScsiLun.AlternateName data correctly/fully
**Describe the bug**

When using mo.HostSystem.Properties and specifying nil for ps, ScsiLun.AlternateName.Data is only populated with the first (least most significant) item in the json array that is stored in the vmware database. For each entry in AlternateName in the GENERIC_VPD namespace, the Data field should contain the inquiry data for the VPDs as reported available by the standard inquiry and read by vmware. Using govmomi, we can only get the least most significant entry in the array (internally represented by vmware and visible using the vm-support command to see the vmware-vimdump file.) This value will show up as 1-3 bytes, depending on the number (i.e.. 125 appears as '313235') - see output below.

**To Reproduce**

This is the code I am using to see the issue

```


    hosts, err := finder.HostSystemList(ctx, "*")
	if err != nil {
		return nil, err
	}

	var targetHost *object.HostSystem
	for _, h := range hosts {

		if h.Name() == hostname {
			targetHost = h
			break
		}
	}
	
	var populatedHost mo.HostSystem
	hostSystemErr := targetHost.Properties(ctx, targetHost.Reference(), nil, &populatedHost)

	if hostSystemErr != nil {
		log.Debugf("Host System Query bombed: %s", hostSystemErr.Error())
		return lunNames, hostSystemErr
	}

	hostLuns := populatedHost.Config.StorageDevice.ScsiLun

	for _, lun := range hostLuns {

		log.Debugf("** Lun: %s **", lun.GetScsiLun().CanonicalName)

		altNames := lun.GetScsiLun().AlternateName

		for _, alt := range altNames {
			log.Debugf("NS: %s, ID: %x Data: %x", alt.Namespace, alt.NamespaceId, alt.Data)
		}
	}
	
```

**Expected behavior**
for GENERIC_VPD namespace, the Data field should contain the full amount of data stored in the json array in the vmware internal database. It currently only gets the single digit, as explained above. It also appears other namespaces have the same issue. 

**Affected version**
	github.com/vmware/govmomi v0.36.2

**Screenshots/Debug Output**

 the output appears like this when the above code is run.

```

time="2024-06-11T18:30:50Z" level=debug msg="** Lun: <redacted> **"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 2d3438"
time="2024-06-11T18:30:50Z" level=debug msg="NS: SERIALNUM, ID: 4 Data: 30"
time="2024-06-11T18:30:50Z" level=debug msg="NS: NAA, ID: 3 Data: 30"
time="2024-06-11T18:30:50Z" level=debug msg="NS: UNKNOWN, ID: 5 Data: 34"
time="2024-06-11T18:30:50Z" level=debug msg="NS: UNKNOWN, ID: 5 Data: 31"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 30"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 30"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 313235"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 313235"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 30"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 313039"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 313033"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 30"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 313235"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 313235"
time="2024-06-11T18:30:50Z" level=debug msg="NS: GENERIC_VPD, ID: 5 Data: 30"


```

**Additional context**
Using pyvmomi, we are able to see this data without an issue. Both libraries should provide the same functionality.

dougm:
Go's encoding/xml package and vCenter marshal '[]byte' differently:
Go encodes the entire array in a single xml element, for example:
```xml
  <foo>
    <bar>hello</bar>
  </foo>
```
vCenter encodes each byte of the array in its own xml element, example with same data as above:
```xml
  <foo>
    <bar>104</bar>
    <bar>101</bar>
    <bar>108</bar>
    <bar>108</bar>
    <bar>111</bar>
  </foo>
```
This behavior is hardwired, see the xml/encoding source for the handful of []byte special cases along the lines of:
```go
  if reflect.Slice && slice element type == reflect.Uint8
```
While govmomi maintains a fork of Go's xml/encoding package, we prefer to further diverge.
Proposed solution is to use a 'ByteSlice' type that implements xml marshaling as vCenter does,
but otherwise behaves just as `[]byte` does.

Commits that follow enhance vcsim and govc support around various `[]byte` fields.

Fixes #1977
Fixes #3469

### DIFF
--- a/vim25/types/helpers.go
+++ b/vim25/types/helpers.go
@@ -1,11 +1,11 @@
 /*
-Copyright (c) 2015-2022 VMware, Inc. All Rights Reserved.
+Copyright (c) 2015-2024 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
--- a/gen/gen_from_vmodl.rb
+++ b/gen/gen_from_vmodl.rb
@@ -1,4 +1,4 @@
-# Copyright (c) 2014 VMware, Inc. All Rights Reserved.
+# Copyright (c) 2014-2024 VMware, Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -86,6 +86,9 @@ class Prop
       when "dateTime"
         type ="time.Time"
       when "byte"
+        if slice?
+          return "types.ByteSlice"
+        end
       when "double"
         type ="float64"
       when "float"
--- a/gen/vim_wsdl.rb
+++ b/gen/vim_wsdl.rb
@@ -1,4 +1,4 @@
-# Copyright (c) 2014-2021 VMware, Inc. All Rights Reserved.
+# Copyright (c) 2014-2024 VMware, Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -321,6 +321,10 @@ class Simple
           self.need_omitempty = false
         end
       when "byte"
+        if slice?
+          prefix = ""
+          t = "#{pkg}ByteSlice"
+        end
       when "double"
         t = "float64"
       when "float"
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -200,6 +200,8 @@ but appear via `govc $cmd -h`:
  - [host.storage.info](#hoststorageinfo)
  - [host.storage.mark](#hoststoragemark)
  - [host.storage.partition](#hoststoragepartition)
+ - [host.tpm.info](#hosttpminfo)
+ - [host.tpm.report](#hosttpmreport)
  - [host.vnic.change](#hostvnicchange)
  - [host.vnic.hint](#hostvnichint)
  - [host.vnic.info](#hostvnicinfo)
@@ -2976,6 +2978,7 @@ Display SSL certificate info for HOST.
 
 Options:
   -host=                 Host system [GOVC_HOST]
+  -show=false            Show PEM encoded server certificate only
 ```
 
 ## host.date.change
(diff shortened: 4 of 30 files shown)


vim25/types/helpers.go
```diff
@@ -316,6 +322,42 @@ func (ci VirtualMachineConfigInfo) ToConfigSpec() VirtualMachineConfigSpec {
 	return cs
 }
 
+type ByteArray []byte
```
akutz, line 325: A comment on the exported symbol describing its purpose and the behavior of the XML encoding/decoding logic would be nice.

akutz, line 325: I also wonder if we should not just create a file `byte_array.go` in case we have more of these in the future? It would make finding them much easier...

akutz (changes requested):
Hey @dougm,

The `ByteArray` change looks good with a few nits. It did take me a moment to realize why you defined `ByteArray` instead of extending `ArrayOfBytes`. I wonder if it makes sense to instead name `type ByteArray` as `type ByteSlice` to further distinguish the two types. Because `ByteArray` really is a `[]byte` (slice of bytes), so I think I prefer `type ByteSlice`. Thoughts?

vim25/types/helpers.go
dougm: Agreed, meant to, done.

dougm: Done, to byte_slice.go

akutz (approved):
Lgtm, thanks Doug!

PR #3481, opened 2024-07-02
api: change CoresPerNumaNode to a pointer type

dougm:
This is needed for the use case of setting CoresPerNumaNode to '0'.


sreyasn (approved):

sreyasn:
LGTM. Thanks @dougm

PR #3492, opened 2024-07-08
chore: fix codeql

tenthirtyam:
## Description

CodeQL is reporting errors. I have simplified and modified the workflow configuration much like I recently did for the `terraform-provider-vsphere`.

Note that the actions now use the `@<commit_hash>` vs arbitrary major version (e.g. @v3).

## Type of Change

- [x] Chore


dougm (approved):
Awesome, thanks @tenthirtyam

PR #3502, opened 2024-07-19
api: use CheckTestType in compat and provisioning checkers

bryanv:
Use the CheckTestType enum as the testType arg type instead of a string in the VmCompatibilityChecker and VmProvisioningChecker functions.  The req task structs end up getting auto generated using the string type so we need to do a little unfortunate conversion.

Add check test type param to the govc commands


dougm (approved):
lgtm, thanks @bryanv

PR #3510, opened 2024-08-02
api: convert vSAN directory name to uuid in ResolveLibraryItemStorage

dougm:
govc: add library.info -S flag
vcsim: add DatastoreNamespaceManager


dilyar85 (approved):
LGTM! Thanks for addressing this promptly.

PR #3511, opened 2024-08-02
fix: restore govc import.spec DiskProvisioning default value

### LINKED ISSUE #3483: govc import ova template failing with version v0.38.0
Following command to upload vmtemplate fails with the latest released version of govc - v0.38.0, while it works with previous version and it successfully uploads the vm template to vCenter.
`govc import.ova -ds=${DATA_STORE} -host=${ESXI_HOST} -options=${TEMPLATE}.json ${TEMPLATE}.ova`

Following error observed 
`govc: The device or operation specified at index '1' is not supported for the existing virtual machine platform.`

Could it be related to any configuration issue or incompatibility with the VM template and the vCenter ?

dougm:
PR #3427 changed to using generated enum lists in various govc commands.
This also changed the default DiskProvisioning value from "flat" to "monolithicSparse",
which results in a DeviceUnsupportedForVmPlatform fault when importing.

This change restores the default to "flat".
The default values for IPProtocol and IPAllocationPolicy had not changed,
but this commit explicitly sets their values rather than using first element of the generated list.

Fixes #3483

vcsim: add ovf DiskProvisioning validation

### DIFF
--- a/govc/test/import.bats
+++ b/govc/test/import.bats
@@ -212,6 +212,25 @@ load test_helper
   grep "invalid NetworkMapping.Name" <<<"$output"
 }
 
+@test "import invalid disk provisioning" {
+  vcsim_env
+
+  ovf="$GOVC_IMAGES/$TTYLINUX_NAME.ovf"
+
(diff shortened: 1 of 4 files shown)


luisdavim (approved):
LGTM

PR #3529, opened 2024-08-23
api: ToString for vim types

akutz:
## Description

This patch adds a ToString helper function for invoking on any vim type. If the specified value is a primitive type, the returned string is generated using Sprintf. Otherwise the returned string is created first by attempting to marshal the data using the vimtype JSON encoder, and if that fails, the stdlib JSON encoder. If that fails, then `fmt.Sprintf("%v")` is used.

Closes: `NA`

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build related change

## How Has This Been Tested?

```shell
go test -v -count 1 -run '^TestToString$' ./vim25/types
```

The following screenshot also demonstrates there is 100% code coverage for this new function:

image

## Checklist:

### DIFF
--- a/vim25/types/helpers.go
+++ b/vim25/types/helpers.go
@@ -17,6 +17,9 @@ limitations under the License.
 package types
 
 import (
+	"bytes"
+	"encoding/json"
+	"fmt"
 	"net/url"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


vim25/types/helpers.go
```diff
@@ -316,6 +319,73 @@ func (ci VirtualMachineConfigInfo) ToConfigSpec() VirtualMachineConfigSpec {
...
+		return marshalWithStdlibJSONEncoder()
+	}
+
+	s = w.String()
+	if len(s) == 0 {
+		return s
+	}
+
+	// Do not include the newline character added by the vimtype JSON encoder.
+	return s[:len(s)-1]
```
dougm, line 386: Would that chop a non-newline char if the stdlib JSON encoder is used?
Should be ok to use `strings.TrimSuffix` instead either way.

dougm (approved):
Thanks @akutz , lgtm

vim25/types/helpers.go
akutz: json.Marshal doesn't add the newline, only encode. Since we use marshal for stdlib, this is why it's not needed for that.

dougm: got it, I just lost track of the `defer` part, this line doesn't apply to the recover / stdlib case.

dougm (approved):
Thanks @akutz !

PR #3531, opened 2024-08-23
govc: add namespace.registervm command

dougm:
api: add namespace.RegisterVM binding
vcsim: add namespace/registervm implementation


aruneshpa (approved):
LGTM.

PR #3539, opened 2024-09-04
Add support for using storage profiles when creating and attaching VirtualDisks

dougm:
This includes support both First Class Disks and "classic" VM Disks.

govc: add disk.create '-profile' flag
govc: add vm.disk.create '-profile' flag
govc: add vm.disk.attach '-profile' flag
chore: govc: add common StorageProfileFlag
api: add VirtualMachine.AddDeviceWithProfile method
vcsim: add VirtualMachine AttachDisk and DetachDisk methods
govc add disk.attach and disk.detach commands
api: VirtualMachine.AttachDisk unitNumber param is optional


dilyar85 (approved):
LGTM, thanks @dougm!

PR #3553, opened 2024-09-19
govc: add kms commands

dougm:


simulator/crypto_manager_kmip.go
```diff
@@ -257,15 +257,28 @@ func (m *CryptoManagerKmip) SetDefaultKmsCluster(
 	return &body
 }
 
+// real vCenter only allows TrustAuthority, but we allow more to simplify test setup
```
akutz, line 260: is the vCenter type not trusted?

akutz (approved):
Lgtm, barring one _possible_ change? Not sure if something was omitted or not. Thanks @dougm !

simulator/crypto_manager_kmip.go
dougm: They are different [KmsManagementTypes](https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U3/html/ReferenceGuides/vim.encryption.KmipClusterInfo.KmsManagementType.html). "vCenter" == "Standard" in the UI.
I haven't looked to see how-to use "trustAuthority" type, only noticed it when I tried using `RegisterKmsCluster` w/ "vCenter" (which is invalid).

PR #3559, opened 2024-09-20
govc: add native kms provider support

dougm:
api: add vapi/crypto package with support for native KMS
api: add rest.Client.DownloadAttachment method

### DIFF
--- a/crypto/doc.go
+++ b/crypto/doc.go
@@ -1,0 +1,21 @@
+/*
+Copyright (c) 2024-2024 VMware, Inc. All Rights Reserved.
+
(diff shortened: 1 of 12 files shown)


akutz (approved):
My _only_ feedback is that it stinks we cannot align the `crypto` and `vapi/crypto` packages a little better. But I guess in the end it makes more sense to keep them aligned with the actual APIs. I wonder if we should add a `README.md` to the `crypto` package (and vice versa) that indicates to users where to look for the different types of crypto logic.

akutz (approved):
Thanks again @dougm !

PR #3563, opened 2024-09-25
api: extract NewDatastoreURL from Datastore.NewURL method

dougm:
This allows the NewDatastoreURL to be used without an object.Datastore instance.

### DIFF
--- a/object/datastore.go
+++ b/object/datastore.go
@@ -105,25 +105,33 @@ func (d Datastore) Path(path string) string {
 	}).String()
 }
 
-// NewURL constructs a url.URL with the given file path for datastore access over HTTP.
-func (d Datastore) NewURL(path string) *url.URL {
-	u := d.c.URL()
-
-	scheme := u.Scheme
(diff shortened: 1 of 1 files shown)


object/datastore.go
```diff
@@ -105,25 +105,29 @@ func (d Datastore) Path(path string) string {
...
+	base.Path = fmt.Sprintf("/folder/%s", path)
+	base.RawQuery = url.Values{
+		"dcPath": []string{dcPath},
+		"dsName": []string{dsName},
+	}.Encode()
+
+	return &base
+}
+
+// NewURL constructs a url.URL with the given file path for datastore access over HTTP.
```
aruneshpa, line 126: Maybe add a comment here that the caller needs to call "FindInventoryPath" (or some other method) that sets the datacenterPath in the DS object?

aruneshpa (approved):

object/datastore.go
dougm: Thanks @aruneshpa , agree, added.

aruneshpa (approved):

PR #3566, opened 2024-09-30
api: disable http.Transport.ForceAttemptHTTP2 by default

### LINKED ISSUE #3564: [BUG] datastore.upload slow performance since v0.31.0 due to the http Client configuration changes
**Describe the bug**
Changes in govmomi v0.31.0 result in poor HTTP upload performance for large files (such as ISOs) over high latency connections, e.g. to servers in a different location.

The performance degradation is extreme, in my scenario uploads with **v0.30.7** run at **30+MB/sec** whereas from **v0.31.0** onwards runs at **0.5 MB/sec**.

This seems to be a knock on from this change https://github.com/vmware/govmomi/pull/3161 which in turn resulted in the client getting a transport configuration of `ForceAttemptHTTP2= true`.
- I fear there are significant performance problems with golang http2 of large files and distant connections. Ref: https://www.google.com/search?q=golang+http2+performance

**To Reproduce**
1. A distant (suggest 100ms latency) vCenter/ESXi is instance is required (could be simulated latency).
2. Observe the performance uploading large file using versions **0.30.7** vs **0.31.0**  
    e.g. `govc datastore.upload some-large-file /test-deleteme` 

**Expected behavior**
Performance should be similar to what it was back on <= 0.30.7

**Affected version**
Degradation introduced in v0.31.0 as part of https://github.com/vmware/govmomi/pull/3161 

**Potential fix**
Please consider adjusting the client transport settings using in govmomi, potentially disabling HTTP2 effectively restoring the settings as they were before.

I tried this via a hack (below) and it sped things up massively!
```go
	if d, ok := http.DefaultTransport.(*http.Transport); ok {
		t = d.Clone()
		t.ForceAttemptHTTP2 = false
		t.TLSClientConfig = nil
	} else {
```
... though this probably requires more understanding.

+ @dougm who may have better understanding.

dougm:
Revert soap.Client http.Transport back to manual construction, rather than Clone() (see 313aa85)

Disable ForceAttemptHTTP2 by default, as we currently see degraded transfer rates with large file uploads.

Closes #3564

### DIFF
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -135,20 +135,31 @@ func ParseURL(s string) (*url.URL, error) {
 	return u, nil
 }
 
+// Go's ForceAttemptHTTP2 default is true, we disable by default.
+// This undocumented env var can be used to enable.
+var http2 = os.Getenv("GOVMOMI_HTTP2") == "true"
+
 func NewClient(u *url.URL, insecure bool) *Client {
 	var t *http.Transport
 
 	if d, ok := http.DefaultTransport.(*http.Transport); ok {
-		t = d.Clone()
+		// Inherit the same defaults explicitly set in http.DefaultTransport,
+		// unless otherwise noted.
+		t = &http.Transport{
+			Proxy:                 d.Proxy,
+			DialContext:           d.DialContext,
+			ForceAttemptHTTP2:     http2, // false by default in govmomi
+			MaxIdleConns:          d.MaxIdleConns,
+			IdleConnTimeout:       d.IdleConnTimeout,
+			TLSHandshakeTimeout:   d.TLSHandshakeTimeout,
+			ExpectContinueTimeout: d.ExpectContinueTimeout,
+		}
 	} else {
 		t = new(http.Transport)
 	}
 
-	if insecure {
-		if t.TLSClientConfig == nil {
-			t.TLSClientConfig = new(tls.Config)
-		}
-		t.TLSClientConfig.InsecureSkipVerify = insecure
+	t.TLSClientConfig = &tls.Config{
+		InsecureSkipVerify: insecure,
 	}
 
 	c := newClientWithTransport(u, insecure, t)


RPitt (approved):
That looks good to me!
Not sure if I can approve but pressing the button anyway :)

vim25/soap/client.go
```diff
@@ -135,19 +135,34 @@ func ParseURL(s string) (*url.URL, error) {
...
 
+	if t.TLSClientConfig == nil {
+		t.TLSClientConfig = new(tls.Config)
+	}
+
 	if insecure {
-		if t.TLSClientConfig == nil {
-			t.TLSClientConfig = new(tls.Config)
-		}
 		t.TLSClientConfig.InsecureSkipVerify = insecure
```
mayankbh, line 166: nit: is this `if` redundant?

vim25/soap/client.go
```diff
@@ -135,19 +135,34 @@ func ParseURL(s string) (*url.URL, error) {
...
+// Go's ForceAttemptHTTP2 default is true, we disable by default.
+// This undocumented env var can be used to enable.
+var http2 = os.Getenv("GOVMOMI_HTTP2") == "true"
+
 func NewClient(u *url.URL, insecure bool) *Client {
 	var t *http.Transport
 
 	if d, ok := http.DefaultTransport.(*http.Transport); ok {
-		t = d.Clone()
+		// Inherit the same defaults explicitly set in http.DefaultTransport,
```
mayankbh, line 146: skimming https://go.dev/src/net/http/transport.go's Clone() implementation it seems like we'd lose a few other defaults, right?

(would it make sense to clone and then override the ForceAttemptHTTP2 field only?)

dougm: We're using the same defaults as http.DefaultTransport: https://github.com/golang/go/blob/8194d735cff90871b1ea5c92e83ddd50abdd4185/src/net/http/transport.go#L45-L56

And changing `ForceAttemptHTTP2=false`. See patch in the original issue, Clone() + ForceAttemptHTTP2=false breaks, unless `TLSClientConfig` set to nil. Rather than dig into this and what else could go wrong with Clone(), chose to revert to our original behavior of copying just the initial fields set by http.DefaultTransport.

vim25/soap/client.go
dougm: Yes, I'll fix that

mayankbh (approved):

PR #3570, opened 2024-10-03
govc: add storage.policy.info flag to query IO filters

dougm:
- Use internalpbm namespace for PbmQueryIOFiltersFromProfileId
- vcsim: override response namespace from struct tag


akutz (approved):
Thanks @dougm !

PR #3574, opened 2024-10-04
vcsim: handle traversal spec object updates

dougm:
Handle the case when an object used for the traversal itself is updated (e.g. ListView).


akutz (approved):
Lgtm, thanks for this @dougm !

PR #3576, opened 2024-10-05
vcsim: support clearing ManagedBy field

dougm:
govc: add vm.change -managed-by flag


prziborowski (approved):
That works for me! Thank you for the fast turnaround.

PR #3584, opened 2024-10-07
vcsim: lock ViewManager in RegisterObject callbacks

dougm:
ViewManager updates itself when objects are added to and removed from inventory. These callbacks need to lock the ViewManager when making these changes.


akutz (approved):
Thanks @dougm !

PR #3585, opened 2024-10-07
api: tags error handling

tenthirtyam:
## Description

Checks if `err` is not `nil` after calling `c.Do`. If an error occurs, it returns `nil` and the error. If no error occurs, it returns the `nil` or the result and `nil` depending on the function.

This ensures that the functions handle errors gracefully and avoids unspecified behavior during the return process.

Closes: ##3112


vapi/tags/categories.go
```diff
@@ -108,13 +111,20 @@ func (c *Manager) UpdateCategory(ctx context.Context, category *Category) error
 		},
 	}
 	url := c.Resource(internal.CategoryPath).WithID(category.ID)
-	return c.Do(ctx, url.Request(http.MethodPatch, spec), nil)
```
dougm, line 111: This is just returning an error, doesn't look like this needs to change?

vapi/tags/categories.go
```diff
@@ -108,13 +111,20 @@ func (c *Manager) UpdateCategory(ctx context.Context, category *Category) error
...
+	if err := c.Do(ctx, req, nil); err != nil {
+		return err
+	}
+	return nil
 }
 
 // DeleteCategory deletes a category.
 func (c *Manager) DeleteCategory(ctx context.Context, category *Category) error {
 	url := c.Resource(internal.CategoryPath).WithID(category.ID)
-	return c.Do(ctx, url.Request(http.MethodDelete), nil)
```
dougm, line 117: Same, just returning an error

vapi/tags/categories.go
```diff
@@ -91,7 +91,10 @@ func (c *Manager) CreateCategory(ctx context.Context, category *Category) (strin
 	}
 	url := c.Resource(internal.CategoryPath)
 	var res string
-	return res, c.Do(ctx, url.Request(http.MethodPost, spec), &res)
```
dougm, line 94: +1 for changing the cases where we return a response value (`var res string` here). But don't think we need to change anything when just return an error (see comments below)

vapi/tags/tag_association.go
```diff
@@ -45,7 +45,10 @@ func (c *Manager) AttachTag(ctx context.Context, tagID string, ref mo.Reference)
 	}
 	spec := internal.NewAssociation(ref)
 	url := c.Resource(internal.AssociationPath).WithID(id).WithAction("attach")
-	return c.Do(ctx, url.Request(http.MethodPost, spec), nil)
```
dougm, line 48: just returning an error, should be fine as-is

vapi/tags/tag_association.go
```diff
@@ -93,7 +99,10 @@ func (c *Manager) AttachTagToMultipleObjects(ctx context.Context, tagID string,
 	}{ids}
 
 	url := c.Resource(internal.AssociationPath).WithID(id).WithAction("attach-tag-to-multiple-objects")
-	return c.Do(ctx, url.Request(http.MethodPost, spec), nil)
```
dougm, line 96: just an error

vapi/tags/tags.go
```diff
@@ -106,13 +109,19 @@ func (c *Manager) UpdateTag(ctx context.Context, tag *Tag) error {
 		},
 	}
 	url := c.Resource(internal.TagPath).WithID(tag.ID)
-	return c.Do(ctx, url.Request(http.MethodPatch, spec), nil)
```
dougm, line 109: just an error

vapi/tags/tags.go
```diff
@@ -106,13 +109,19 @@ func (c *Manager) UpdateTag(ctx context.Context, tag *Tag) error {
...
+	if err := c.Do(ctx, url.Request(http.MethodPatch, spec), nil); err != nil {
+		return err
+	}
+	return nil
 }
 
 // DeleteTag deletes an existing tag.
 func (c *Manager) DeleteTag(ctx context.Context, tag *Tag) error {
 	url := c.Resource(internal.TagPath).WithID(tag.ID)
-	return c.Do(ctx, url.Request(http.MethodDelete), nil)
```
dougm, line 115: just an error

dougm (commented):
Thanks @tenthirtyam , looks like we can leave some of the code as-is, in the cases where we just return an `error`?

tenthirtyam:
Sure thing. I'll amend this in the morning.

vapi/tags/tags.go
tenthirtyam: Resolved in https://github.com/vmware/govmomi/commit/d14cbf39ee092c1a12388ab5b8009bc3c443c898.

vapi/tags/tags.go
tenthirtyam: Resolved in https://github.com/vmware/govmomi/commit/d14cbf39ee092c1a12388ab5b8009bc3c443c898.

vapi/tags/tag_association.go
tenthirtyam: Resolved in https://github.com/vmware/govmomi/commit/d14cbf39ee092c1a12388ab5b8009bc3c443c898.

vapi/tags/tag_association.go
tenthirtyam: Resolved in https://github.com/vmware/govmomi/commit/d14cbf39ee092c1a12388ab5b8009bc3c443c898.

vapi/tags/categories.go
tenthirtyam: Resolved in https://github.com/vmware/govmomi/commit/d14cbf39ee092c1a12388ab5b8009bc3c443c898.

vapi/tags/categories.go
tenthirtyam: Resolved in https://github.com/vmware/govmomi/commit/d14cbf39ee092c1a12388ab5b8009bc3c443c898.

dougm (approved):
Thanks @tenthirtyam

PR #3587, opened 2024-10-09
govc: support use of service version via GOVC_VIM_VERSION env var

dougm:
When the '-vim-version' flag is set to an empty value "", service version is used. Doing the same with a value of "-" allows service version to be used via the env var.


derekbeard (commented):
Looks good to me.  Thanks for adding this.

derekbeard (approved):

Issue #3555, opened 2024-09-20
[BUG] OVF file uploads not using absolute path


Xiol:
**Describe the bug**
Similar to #3550, the actual file uploads that `Importer.Upload()` attempts are done via relative paths. This results in the file not being found when the upload is attempted from a different working directory when using `importer.Importer`  from another application.

**To Reproduce**

1. Attempt to upload an OVF using `importer.Importer` while having your working directory different to the OVF path.
1. Notice that the upload fails to find some of the files.

**Expected behavior**
The upload should happen successfully regardless of the current working directory.

**Affected version**
v0.43.0

**Screenshots/Debug Output**
N/A

**Additional context**
N/A

Xiol:
After further investigation this appears to have been caused by the FileArchive.Open() function not handling paths in a cross-platform manner (our code unfortunately targets Windows machines). I've opened another PR to resolve this.

PR #3596, opened 2024-10-17
fix: Cross-platform path handling for OVF FileArchive uploads

Xiol:
## Description

Path handling for OVF uploads was using hardcoded path separators that did not support usage on Windows. This change handles paths in a cross-platform manner and ensures the absolute path is being used at all times.

Closes: #3555 

## How Has This Been Tested?

We have tested this in our own environment and confirmed it resolves the issues we were seeing with relative paths having paths prefixed on incorrectly.


dougm (approved):
Thanks @Xiol

PR #3603, opened 2024-10-25
fix: vcsim: avoid race in DestroyPropertyFilter

dougm:
Need to lock the filter's PropertyCollector when removing reference to self.


akutz (approved):
Thank you @dougm !

bryanv (approved):

PR #3609, opened 2024-11-03
api: support MOID conversion in Finder methods

dougm:
With this change, finder methods can convert ManagedObjectReference or MOID to an object instance. For example:

```console
# `-i` outputs MOR
% govc find -l -i -type m / -name nginx
VirtualMachine:vm-1100  /test-vpx-1730432486-649974-wcp.wcp-sanity/vm/nginx

% govc vm.info VirtualMachine:vm-1100
Name:           nginx
  Path:         /test-vpx-1730432486-649974-wcp.wcp-sanity/vm/nginx
  UUID:         421b8fd2-3006-0a69-1f16-7d088f90145f
  Guest name:   Other (32-bit)
   ...

# `-I` outputs MOID
% govc find -l -I -type m / -name nginx
vm-1100  /test-vpx-1730432486-649974-wcp.wcp-sanity/vm/nginx

% govc vm.info vm-1100
Name:           nginx
  Path:         /test-vpx-1730432486-649974-wcp.wcp-sanity/vm/nginx
  UUID:         421b8fd2-3006-0a69-1f16-7d088f90145f
  Guest name:   Other (32-bit)
...

% govc collect -s vm-1100 runtime.host | cut -d: -f2
host-14

% govc host.info host-14
Name:              10.192.85.11
  Path:            /test-vpx-1730432486-649974-wcp.wcp-sanity/host/test-vpx-1730432486-649974-wcp.wcp-sanity-cluster/10.192.85.11
  Manufacturer:    VMware, Inc.
  Logical CPUs:    4 CPUs @ 2100MHz
 ...
```

### DIFF
--- a/find/finder.go
+++ b/find/finder.go
@@ -1,11 +1,11 @@
 /*
-Copyright (c) 2014-2023 VMware, Inc. All Rights Reserved.
+Copyright (c) 2014-2024 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
@@ -22,6 +22,7 @@ import (
 	"path"
 	"strings"
 
+	"github.com/vmware/govmomi/fault"
 	"github.com/vmware/govmomi/internal"
 	"github.com/vmware/govmomi/list"
 	"github.com/vmware/govmomi/object"
@@ -115,6 +116,19 @@ func (f *Finder) findRoot(ctx context.Context, root *list.Element, parts []strin
 
 func (f *Finder) find(ctx context.Context, arg string, s *spec) ([]list.Element, error) {
 	isPath := strings.Contains(arg, "/")
+
+	if !isPath {
+		if ref := object.ReferenceFromString(arg); ref != nil {
+			p, err := InventoryPath(ctx, f.client, *ref)
+			if err == nil {
+				if t, ok := mo.Value(*ref); ok {
+					return []list.Element{{Object: t, Path: p}}, nil
+				}
+			} else if !fault.Is(err, &types.ManagedObjectNotFound{}) {
+				return nil, err
+			} // else fall through to name based lookup
+		}
+	}
 
 	root := list.Element{
 		Object: object.NewRootFolder(f.client),
@@ -821,11 +835,6 @@ func (f *Finder) Network(ctx context.Context, path string) (object.NetworkRefere
 }
 
 func (f *Finder) networkByID(ctx context.Context, path string) (object.NetworkReference, error) {
-	if ref := object.ReferenceFromString(path); ref != nil {
-		// This is a MOID
-		return object.NewReference(f.client, *ref).(object.NetworkReference), nil
-	}
-
 	kind := []string{"DistributedVirtualPortgroup"}
 
 	m := view.NewManager(f.client)
--- a/find/finder_test.go
+++ b/find/finder_test.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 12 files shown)


find/finder.go
```diff
@@ -113,9 +113,68 @@ func (f *Finder) findRoot(ctx context.Context, root *list.Element, parts []strin
...
+	return nil
+}
+
 func (f *Finder) find(ctx context.Context, arg string, s *spec) ([]list.Element, error) {
 	isPath := strings.Contains(arg, "/")
 
+	if !isPath {
+		if ref := referenceFromString(arg); ref != nil {
+			p, err := InventoryPath(ctx, f.client, *ref)
+			if err == nil {
```
rikatz, line 170: In case err != nil here, should we care? eg.: maybe the client has some issue? context is timed out? (I don't know which cases can happen here, just wondering)

find/finder_test.go
```diff
@@ -105,3 +105,42 @@ func TestFindNetwork(t *testing.T) {
...
+		for _, vm := range vms {
+			ref := vm.Reference()
+			byRef, err := find.VirtualMachine(ctx, ref.String())
+			if err != nil {
+				t.Fatal(err)
+			}
+			if byRef.Reference() != ref {
+				t.Error(byRef.Reference())
+			}
+			_, err = find.VirtualMachine(ctx, ref.String()+"invalid")
```
rikatz, line 127: a suggestion is to maybe also check if the inventorypath is really populated here

find/finder.go
```diff
@@ -113,9 +113,68 @@ func (f *Finder) findRoot(ctx context.Context, root *list.Element, parts []strin
...
+	if ref := object.ReferenceFromString(arg); ref != nil {
+		return ref
+	}
+
+	id := strings.SplitN(arg, "-", 2)
+	if len(id) != 2 {
+		return nil
+	}
+
+	if kind, ok := refTypeMap[id[0]]; ok {
```
rikatz, line 145: Just for my understanding, this is invalid when someone passes as an argument something like:

"datacenter-3" right?

So which specific case we fall into this? I am not down deep on vSphere API, I remember this may be a supervisor domain, just trying to understand what "argument" here is valid.

find/finder.go
```diff
@@ -113,9 +113,68 @@ func (f *Finder) findRoot(ctx context.Context, root *list.Element, parts []strin
...
+		return &types.ManagedObjectReference{
+			Type:  kind,
+			Value: arg,
+		}
+	}
+
+	return nil
+}
+
 func (f *Finder) find(ctx context.Context, arg string, s *spec) ([]list.Element, error) {
```
rikatz, line 164: a question here is: do we want to make it a standard behavior on finder, or should we extend its configuration to at least make the consumer explicitly state that searching by MOID is also desired?

An example: I was testing on vsphere that one can call the datacenter as "Datacenter:datacenter-3" and this is a valid name (tho I don't know which kind of person would call a datacenter on a moid format...). 

In this case, I may say I want to search my datacenter named "Datacenter:datacenter-3" but because we assume this is a moid it will fallback into a bad path.

find/finder.go
dougm: No, "datacenter-3" is valid. 

`if object.ReferenceFromString` is true when value is a managed object reference (MOR). That is, it includes the `Type:` `Datacenter:datacenter-3`

This section parses managed object ID (MOID). That is, without the `Type:` - the type is derived from the ID value.

find/finder.go
dougm: The call to `InventoryPath` will fail with `ManagedObjectNotFound` fault in that case, and we fall through to the name based lookup.

find/finder.go
dougm: If there was a generate connection/auth/etc issue, we'd most likely hit the same on the next API call. But I'll add an explicit check for `ManagedObjectNotFound` which is when we want to fall through to name based lookup.

find/finder_test.go
dougm: Good idea, done.

find/finder.go
dougm: Moved all this into `object.ReferenceFromString` as it isn't Finder specific, and included these comments there. `TestReferenceFromString` should also make this more clear.

rikatz (approved):

find/finder.go
lubronzhan: I just saw `datacenter-3` is also a valid datacenter-3, `Datacenter:datacenter-3` is less likely to be used since I don't see anyone use `:` inside their name so far. But if x-x is more likely to be a often used valid name? it could be a case that one datacenter has name `datacenter-3`, and another datacenter `datacenter-4` has moid `datacenter-3`, Then the client that use this func might find the datacenter that it wants to use is unexpected.

dougm: @lubronzhan in that case the Finder call to `InventoryPath` results in `ManagedObjectNotFound` and we fall through to name based lookup. Clients shouldn't depend on the `object.ReferenceFromString` result alone, it needs to be validated as Finder does.

dougm: Here's an example:
```console
% govc datacenter.create datacenter-123

% govc find -l -i / -type d
Datacenter:datacenter-2   /DC0
Datacenter:datacenter-74  /datacenter-123

% govc datacenter.info datacenter-123
Name:                datacenter-123
  Path:              /datacenter-123
  Hosts:             0
  ...
  
% govc datacenter.info Datacenter:datacenter-123
govc: datacenter 'Datacenter:datacenter-123' not found

% govc datacenter.info datacenter-74
Name:                datacenter-123
  Path:              /datacenter-123
  Hosts:             0
  ...
```

PR #3611, opened 2024-11-04
chore: skip tests unless running docker on linux

### LINKED ISSUE #3604: [BUG] some tests fail without docker on linux
**Describe the bug**
go test is failing on FreeBSD 13 ports repository. Execution of 'make test' is failing. The same results for Go 1.21 & 1.22 from Ports.

**To Reproduce**

1. cd sysutils/govmomi
2. make test

**Expected behavior**
test is PASSed. If platform is unsupported, skip test as unsupported.

**Affected version**
Tested on 0.45.1, 0.44.1, 0.33.0

**Screenshots/Debug Output**
Log is attached.
[govmimo-test.txt](https://github.com/user-attachments/files/17532658/govmimo-test.txt)

**Additional context**
Highly possible, the FAIL result is because Docker isn't well supported. Is any way to skip tests for unsupported platforms?

dougm:
Fixes #3604

### DIFF
--- a/guest/toolbox/example_test.go
+++ b/guest/toolbox/example_test.go
@@ -18,18 +18,24 @@ package toolbox_test
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/exec"
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/guest/toolbox"
 	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/test"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
 func ExampleClient_Run() {
 	simulator.Run(func(ctx context.Context, c *vim25.Client) error {
+		if !test.HasDocker() {
+			fmt.Println("Linux")
+			return nil
+		}
 		vm, err := find.NewFinder(c).VirtualMachine(ctx, "DC0_H0_VM0")
 		if err != nil {
 			return err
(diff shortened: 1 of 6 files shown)


guest/toolbox/example_test.go
```diff
@@ -18,18 +18,24 @@ package toolbox_test
...
 	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/test"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
 func ExampleClient_Run() {
 	simulator.Run(func(ctx context.Context, c *vim25.Client) error {
+		if !test.HasDocker() {
+			fmt.Println("Linux")
```
rikatz, line 36: was this print left intentionally here?

test/helper.go
```diff
@@ -28,6 +30,16 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
+func HasDocker() bool {
+	if runtime.GOOS != "linux" {
```
rikatz, line 34: I am assuming this is also true for WSL/Windows :)

rikatz (approved):
leaving approval, besides the fmt.Println everything seems fine to me

guest/toolbox/example_test.go
dougm: Yes, it was already the expected output of this Example test, see `// Output:` below.
But in Example tests, we can't use `testing.T.Skip()`, so just print the expected output and return.

test/helper.go
dougm: I had to google 'WSL', Windows Subsystem for Linux.. yeah not worried about that :)

guest/toolbox/example_test.go
rikatz: aaaaah indeed sorry! missed the output assertion

PR #3617, opened 2024-11-08
Minimum version selection

akutz:
## Description

This patch updates this project's root `go.mod` file to specify the minimum required version of Go and minimum required versions for dependencies.

This change ensures consumers of the GoVmomi module are not forced to use a version of Golang or dependency just because this project's root `go.mod` file specifies them.

The release of Go 1.21 changed the meaning of the Go version specified in  `go.mod` (https://go.dev/doc/modules/gomod-ref#go-notes):

> The go directive sets the minimum version of Go required to use this module. Before Go 1.21, the directive was advisory only; now it is a mandatory requirement: Go toolchains refuse to use modules declaring newer Go versions.

This means if GoVmomi specified Go 1.23.2 in its `go.mod` file, any project that imported GoVmomi would also be required to use Go 1.23.2.

Of course the `govc` and `vcsim` binaries still specify the latest Go version and newest dependencies to mitigate CVEs. Except for these two packages, the rest of GoVmomi is never built as a binary artifact, ex. executable / plug-in. Thus it is fine for the rest of the project to use an older Go version and older dependencies.

The only downside to this change is that `govc` and `vcsim` may no longer be installed by running:

* `go install github.com/vmware/govmomi/govc@latest`
* `go install github.com/vmware/govmomi/vcsim@latest`

This is because both packages are now distinct sub-modules that still depend on the other packages in this project. Therefore these sub-mods use the following replace directive:

```
replace github.com/vmware/govmomi -> ../
```

This teaches govc and vcsim where to look for the imported GoVmomi packages. This ensures that building these binaries will pick up any local changes to the rest of the project.

However, the `go install` command does not support installing modules from a remote location if the module contains a `replace` directive. This means users will need to either clone all of github.com/vmware/govmomi before running `go -C govc install` or `go -C vcsim install`, or they can download the released binaries directly.

Closes: `NA`

## How Has This Been Tested?

Ran the following commands locally:

* `make install`
* `gorelease release --snapshot --clean` 
* `make test`

## Guidelines


go.mod
```diff
@@ -1,24 +1,21 @@
 module github.com/vmware/govmomi
 
-go 1.22
+go 1.21 // required for slices/maps package and built-in clear function
```
dougm, line 3: I had bumped to 1.22 here just because we support the latest 2 Go minor versions and only test w/ those versions.

dougm (approved):
lgtm, thanks @akutz !

go.mod
akutz: Hmm, how do you want to handle this? Should we consider changing this behavior to "leave version alone unless the library needs a newer version?"

PR #3619, opened 2024-11-08
govc: Move govc commands into cli pkg

akutz:
## Description

This patch refactors the `govc` command so all of its actual logic and commands are located in the root GoVmomi module under the `cli` package. The `govc` package may still be used to build, install, and test `govc`, but this change allows others to import the pieces of `govc` without being beholden to the `govc` module's Golang version.

All packages `./govc/*` (except `./govc/test`) have been relocated under the `./cli` package.

Closes: `NA`

## How Has This Been Tested?

* `make install`
* `make test`
* `make govc-test`

## Guidelines


dougm (commented):
Thanks @akutz

Also need to update: https://github.com/vmware/govmomi/blob/main/.goreleaser.yml#L20

Can we leave emacs in the govc directory? That'll avoid breaking this: https://github.com/melpa/melpa/blob/master/recipes/govc

Maybe keep examples/lib/ssh.sh under govc too. Or we could just remove it, the embedded examples we have now are more useful.

akutz:
> Thanks @akutz
> 
> Also need to update: https://github.com/vmware/govmomi/blob/main/.goreleaser.yml#L20
> 
> Can we leave emacs in the govc directory? That'll avoid breaking this: https://github.com/melpa/melpa/blob/master/recipes/govc
> 
> Maybe keep examples/lib/ssh.sh under govc too. Or we could just remove it, the embedded examples we have now are more useful.

Done, done, and done!

dougm (approved):
lgtm, thanks @akutz !

PR #3622, opened 2024-11-12
fix: vcsim: avoid possible panic when applying PropertyFilter

dougm:


akutz (approved):
Lgtm, thanks @dougm !

PR #3627, opened 2024-11-18
fix: govc: wire up flags for namespace service create

rosenhouse:
## Description

`govc namespace.service.create` had flags defined in code that were never usable.

## How Has This Been Tested?

```
govc namespace.service.create -h
```

Now shows the `spec-type` `trusted` and `accept-eula` flags.  Before this commit, those flags weren't shown in help and caused an error if provided.

## Other notes
I looked for other incorrect usages like this, by doing `rg 'flag.*Var' ./govc`.  This was the only one.

### DIFF
--- a/govc/USAGE.md
+++ b/govc/USAGE.md
@@ -4750,6 +4750,9 @@ Examples:
   govc namespace.service.create manifest.yaml
 
 Options:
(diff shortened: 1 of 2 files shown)


dougm (approved):
Thanks @rosenhouse

PR #3633, opened 2024-11-23
vcsim: add esxcli support

dougm:
api: move cli/host/esxcli.Command to cli/esx package

BREAKING: Package "github.com/vmware/govmomi/govc/host/esxcli" is now "github.com/vmware/govmomi/cli/esx"
A 'context.Context' param has also been added to the package's methods.
See cli/esx/example_test.go for usage.


cli/esx/command.go
```diff
@@ -1,5 +1,5 @@
 /*
```
tenthirtyam, line 1: Suggestion to update the headers to the following but can be left to a seperate PR.

THis has been done on several other projects based on discussion with Chip C.

```go
// © Broadcom. All Rights Reserved.
// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
// SPDX-License-Identifier: Apache-2.0
```

dougm:
Thanks @akutz !

cli/esx/command.go
dougm: Thanks @tenthirtyam

tenthirtyam: If you'd like help making this change, feel free to DM me and I'll help out with that.

PR #3636, opened 2024-12-05
api: add override mechanism for SOAP Header.Cookie

dougm:
Sessions are consumed by API endpoints via one of:
- HTTP Cookie (vim25 for example)
- HTTP Header (vmodl2 /rest and /api)
- SOAP Header (pbm, vslm, sms for example)

The soap.Client had always set the SOAP Header cookie regardless if consumed by an API endpoint.
Clients must now opt-in to this behavior if they need it, such as pbm and vslm.

This also allows clients to change the name of soap.Header.Cookie (default is still "vcSessionCookie")

vcsim: add override mechanism for session mapping

On the simulator side, API endpoints can specify where a session is sourced from. Default is still the "vmware_soap_session" HTTP Cookie. The pbm simulator now specifies the "vcSessionCookie" SOAP Header, behaving as real pbm/VC does.

### DIFF
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2014-2023 VMware, Inc. All Rights Reserved.
+Copyright (c) 2014-2024 VMware, Inc. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


vim25/soap/client.go
```diff
@@ -86,7 +86,7 @@ type Client struct {
 	Types     types.Func `json:"types"`
 	UserAgent string     `json:"userAgent"`
 
-	cookie          string
+	Cookie          func() any
```
akutz, line 89: Should we document this and indicate that someone can set it to `nil` to avoid participating in the cookie?

akutz (approved):
Lgtm. Left one comment.

akutz (approved):
Thanks Doug!

PR #3637, opened 2024-12-05
vcsim: add UpdateAssignedLicense method

dougm:
- govc object.save includes licenseAssignmentManager property
- fix: assert Entity interface in mo.Value()


dbhat-arkin (approved):
LGTM.
Thanks

dbhat-arkin (approved):
LGTM .. Thanks

PR #3641, opened 2024-12-05
fix: vcsim: swap order of xsi type attribute

### LINKED ISSUE #2114: Folder type is returned as ManagedObjectReference
Hi Doug,

image
It is for com.vmware.vim25.ws.VimStub#retrieveProperties, where managed object reference can be HostSystem, DataStore or VirtualMachine. And the property is parent.

Is this info sufficient ?

dougm:
PropertyCollector responses always included the 'xsi:type' attribute. When a property type is ManagedObjectReference, there are 2 'type' attributes in the response, for example:
```xml
 <val xsi:type="ManagedObjectReference" type="LicenseAssignmentManager">LicenseAssignmentManager</val>
```
Some clients (Java) have problems with this. Changing the order to be the same as real vCenter to workaround:
```xml
 <val type="LicenseAssignmentManager" xsi:type="ManagedObjectReference">LicenseAssignmentManager</val>
```

Fixes #2114

### DIFF
--- a/vim25/xml/marshal.go
+++ b/vim25/xml/marshal.go
@@ -497,11 +497,6 @@ func (p *printer) marshalValue(val reflect.Value, finfo *fieldInfo, startTemplat
 		start.Name.Local = name
 	}
 
-	// Add type attribute if necessary
-	if finfo != nil && finfo.flags&fTypeAttr != 0 {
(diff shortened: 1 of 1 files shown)


dbhat-arkin (approved):
LGTM.
Thanks

PR #3651, opened 2024-12-12
vcsim: fix object.save when run directly against ESX

dougm:
QueryTpmAttestationReport is NotImplemented, vCenter only method.


dbhat-arkin (approved):
LGTM

PR #3653, opened 2024-12-15
fix: check `cerr != nil` but return a nil value error `err`

alingse:
## Description

the if branch just check `cerr !=nil` and return a nil value error `err`

so, it's a harmless bug.

I have an idea to detect the code that return a no-realte nilness error bug, and check the top 1000 github go repos, and found this. see https://github.com/alingse/sundrylint/issues/4

Closes: #(issue-number)

## How Has This Been Tested?

just use the cli can reproduce this

## Guidelines


dougm (approved):
Thanks @alingse !

alingse:
@dougm ok,no problem, I will fix it after work.

dougm (approved):
Thanks @alingse , I also squashed the commits and update the commit message

PR #3657, opened 2024-12-19
govc: add datastore.ls -H option

dougm:
Use human friendly name by default, rather than vSAN ID for top-level dirs.

Note that the datastore.ls command was originally written against vSphere 5.5, the 'FileInfo.FriendlyName' field was not added until 6.5


akutz (approved):
Lgtm, thanks @dougm !

PR #3664, opened 2024-12-31
vcsim: create cls vmdk descriptor and backing files

dougm:
Certain vcsim VirtualDiskManager methods (Copy,Move,Delete) expect a .vmdk descriptor and backing -flat.vmdk, create both when creating and syncing vmdk library items.

The new vmdk.Descriptor has more use cases in vcsim and client side, those will come in future PRs.


akutz (approved):
Thanks @dougm !

PR #3666, opened 2025-01-02
govc: add datastore.download -json support

dougm:
Only applies to ".ovf" and ".vmdk" files currently.


akutz (approved):
Nice one, thanks @dougm !

PR #3668, opened 2025-01-03
goreleaser: remove brew tap setup

chenrui333:
## Description

since the formula is maintained in the core tap side, remove the brew tap setup in the goreleaser config

relates to https://github.com/Homebrew/homebrew-core/pull/203058


dougm (approved):
Thanks @chenrui333 !

PR #3669, opened 2025-01-06
fix: govc: import.ovf remote support regression

dougm:
- Fix regression introduced in 36ee9a1e / PR #3596
- Add test for remote import.ovf


luisdavim (approved):
LGTM

PR #3670, opened 2025-01-06
vcsim: add DatastoreNamespaceManager Create/Delete directory methods

dougm:


luisdavim (approved):
LGTM

dougm:
> Lgtm @dougm , thanks, but should it not simulate the UUID as the underlying dir name and use the Convert call to get back the UUID from a friendly name later?

Will follow up on this in a separate PR

PR #3676, opened 2025-01-13
vcsim: add vslm VStorageObjectManager (aka FCD Global Catalog) support

### LINKED ISSUE #2542: vcsim does not handle VSLM GlobalObjectManager methods
`vcsim` does not currently handle calls from the VSLM `GlobalObjectManager`, which means it's currently impossible to use vcsim to test those code that uses them.

It should be fairly straightforward to add a simulator implementation for `VslmVStorageObjectManager` and register it as a handler for the `/vslm/sdk` path. Most of the endpoints are largely the same as the existing `VcenterVStorageObjectManager`, although there are additional endpoints that the local version does not support.

The only problem I can think of offhand with that approach is that a struct that handles `VslmVStorageObjectManager` needs to share the same set of objects as the existing `VcenterVStorageObjectManager`. This makes it difficult to use the simulator's `RegisterSDK` method to add the handler. There is also an ordering dependency which means that the setup might have to happen as part of initially configuring the `VStorageObjectManager`.

dougm:
govc: make datastore (-ds flag) optional in (most) disk commands

Fixes #2542


cli/disk/manager.go
```diff
@@ -0,0 +1,309 @@
...
+		if err := m.initGlobalManager(ctx); err != nil {
+			return nil, err
+		}
+	}
+
+	return m, nil
+}
+
+func (m *Manager) initGlobalManager(ctx context.Context) error {
+	if m.GlobalObjectManager == nil {
```
akutz, line 78: Should this use `sync.Once` anywhere?

akutz (approved):
Lgtm @dougm , thank you!

cli/disk/manager.go
dougm: Might just get rid of this deferred initGlobalManager in the follow up to add metadata and query support. Instead will always create `vslm.NewClient` in `NewManager`, unless we're connected directly to ESX.
Also, only consumer of this disk.Manager is govc. I think most applications should directly just use vslm.Client against vCenter. Maybe we do the same in govc at some point and this Manager goes away completely.

PR #3677, opened 2025-01-14
vcsim: fix lookup service deferred registration

dougm:
With this change, List() can be called withing having called RetrieveServiceContent()


shihhaoli (approved):
Tested manually.

PR #3678, opened 2025-01-16
vcsim: add ssoadmin GetTrustedCertificates method

dougm:


shihhaoli (approved):
Manually tested and works well. Thanks!

PR #3681, opened 2025-01-17
vcsim: fix save+load of Alarm objects

dougm:
govc object.save includes Alarm objects, but vcsim -load did not wrap the Alarm objects.
This causes panic when trying to call CreateAlarm() which expects wrapped Alarm objects.


shihhaoli (approved):
tested and crashed no longer happened, thanks

PR #3682, opened 2025-01-22
vcsim: add disk query and metadata support

dougm:
- govc: add disk.metadata ls and update commands

- api: add vlsm.GlobalObjectManager.List method


akutz (approved):
Lgtm, thanks @dougm !

PR #3686, opened 2025-01-27
Add ResourcePool.resourceConfigSpecDetailed to missingSet filter

dougm:
New property (resourceConfigSpecDetailed) is only valid when collected against vCenter. Against ESX is NotSupported and returned by RetrievePropertiesEx in the missingSet when all properties are requested, which by default is propagated as an error.


divyenpatel (approved):
/approve
/lgtm

PR #3694, opened 2025-02-10
vcsim: add query and metadata support for CNS Volume API

dougm:
vcsim: add disk backing for cns volumes

govc: add volume.create command

govc: add volume.ls query filter flags

### DIFF
--- a/cli/volume/ls.go
+++ b/cli/volume/ls.go
@@ -1,18 +1,6 @@
-/*
-Copyright (c) 2020-2023 VMware, Inc. All Rights Reserved.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
 
 package volume
 
@@ -35,6 +23,7 @@ import (
 type ls struct {
 	*flags.ClientFlag
 	*flags.DatastoreFlag
+	*flags.StorageProfileFlag
 	*flags.OutputFlag
 
 	types.CnsQueryFilter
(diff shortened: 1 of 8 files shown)


cli/volume/ls.go
```diff
@@ -45,6 +34,21 @@ type ls struct {
 	back bool
 }
 
+type keyValue []vim.KeyValue
+
+func (e *keyValue) String() string {
```
akutz, line 39: This function receiver doesn't need to be by address, and changing it could prevent a possible NPE.

cns/simulator/simulator.go
```diff
@@ -65,93 +56,153 @@ type CnsVolumeManager struct {
...
 
-		operationResult := []cnstypes.BaseCnsVolumeOperationResult{}
+		var operationResult []cnstypes.BaseCnsVolumeOperationResult
+
 		for _, createSpec := range req.CreateSpecs {
-			staticProvisionedSpec, ok := interface{}(createSpec.BackingObjectDetails).(*cnstypes.CnsBlockBackingDetails)
-			if ok && staticProvisionedSpec.BackingDiskId != "" {
-				datastore := simulator.Map.Any("Datastore").(*simulator.Datastore)
+			details, ok := createSpec.BackingObjectDetails.(*cnstypes.CnsBlockBackingDetails)
+			if ok && details.BackingDiskId != "" {
```
akutz, line 71: If you invert the logic here so `if !ok || details.BackingDiskId == "" { continue; }` then you could dedent the large block below.

cns/simulator/simulator.go
```diff
@@ -65,93 +56,153 @@ type CnsVolumeManager struct {
...
+				datastore := func() *simulator.Datastore {
```
akutz, line 72: Why is this a function? Also, to be safe, you could pass in the `id` value instead of grabbing it from the outer loop. That would prevent against any accident weak references.

akutz (changes requested):
Howdy @dougm , 

Thank you so much for doing this, I really appreciate it. I left a few nits we should consider. I also wonder if we should ask someone like Somenath or Zhibin to review this and provide their thoughts. I'll start a thread internally for us.

cli/volume/ls.go
```diff
@@ -56,13 +60,21 @@ func (cmd *ls) Register(ctx context.Context, f *flag.FlagSet) {
...
+
 	cmd.OutputFlag, ctx = flags.NewOutputFlag(ctx)
 	cmd.OutputFlag.Register(ctx, f)
 
 	f.BoolVar(&cmd.long, "l", false, "Long listing format")
 	f.BoolVar(&cmd.id, "i", false, "List volume ID only")
 	f.BoolVar(&cmd.disk, "L", false, "List volume disk or file backing ID only")
 	f.BoolVar(&cmd.back, "b", false, "List file backing path")
+
+	f.Var((*flags.StringList)(&cmd.Names), "n", "List volumes with names")
```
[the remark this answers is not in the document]
dougm: The volume.ls generated usage/help (`-h`) includes these new flags.
In general, govc commands "Examples" section doesn't need to include examples for all flags.

fatsbean (approved):
LGTM.

akutz (approved):
Lgtm, thanks @dougm !

PR #3701, opened 2025-03-03
vcsim: remove simulator.Map package variable

dougm:
This package var has prevented multiple instances of vcsim running in parallel the same process.
Each simulator Model/Service instance now has its own Registry (Map) instance and Context.
The simulator.{Test,Run} functions are now passed the simulator.Context instance as the
context.Context callback param. 'simulator.Map' is now a helper function / shortcut to
reference the Map field of the Context instance.

BREAKING: simulator.Map package variable has been removed

BREAKING: simulator.SpoofContext function has been renamed to NewContext

### DIFF
--- a/cns/simulator/simulator_test.go
+++ b/cns/simulator/simulator_test.go
@@ -58,7 +58,7 @@ func TestSimulator(t *testing.T) {
 	existingNumDisks := len(queryResult.Volumes)
 
 	// Get a simulator DS
-	datastore := simulator.Map.Any("Datastore").(*simulator.Datastore)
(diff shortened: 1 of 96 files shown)


dougm:
vm-operator update against this branch: https://github.com/vmware-tanzu/vm-operator/compare/main...dougm:vcsim-refactor

dilyar85 (approved):
LGTM! I didn't review every line due to the size of the change, but I did some sanity checks and confirmed there's no usage of the `simulator.Map` package var. Thanks @dougm!

PR #3702, opened 2025-03-05
govc: add volume.extend command

dougm:
govc: add volume.rm -keep flag to retain backing disk

vcsim: don't delete backing disk in CnsDeleteVolume when DeleteDisk==false

chore: Use cns Task helpers in govc volume commands

### DIFF
--- a/cns/cns_util.go
+++ b/cns/cns_util.go
@@ -1,18 +1,6 @@
-/*
-Copyright (c) 2019 VMware, Inc. All Rights Reserved.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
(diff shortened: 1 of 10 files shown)


cns/cns_util.go
```diff
@@ -49,9 +37,6 @@ func GetQuerySnapshotsTaskResult(ctx context.Context, taskInfo *vim25types.TaskI
 	}
 	if taskInfo.Result != nil {
 		snapshotQueryResult := taskInfo.Result.(cnstypes.CnsSnapshotQueryResult)
-		if &snapshotQueryResult == nil {
```
aruneshpa, line 52: I don't understand this.  Was this a pointer to a pointer?  If so, do we no longer want to check this?

aruneshpa (approved):
LGTM.  Thanks @dougm !

PR #3704, opened 2025-03-06
Use the same session of SOAP client when VC is bigger than 8.0U3

rikatz:
## Description

On vCenter versions bigger than 8.0U3 there is no need anymore to do a new Rest Login. Instead, the same token/cookie from the Soap client can be reused.

The current change makes the rest.NewClient verify if the parent soap client was created against a vCenter bigger than 8U3 and reuse the session, and also turns Login() a noop in the same case.

## How Has This Been Tested?

Did some internal tests with the code calling library and tags API. More tests are pending

## Guidelines

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/vmware/govmomi
 go 1.21 // required for slices/maps package and built-in clear function
 
 require (
+	github.com/Masterminds/semver/v3 v3.3.1
 	github.com/a8m/tree v0.0.0-20230208161321-36ae24ddad15
 	github.com/dougm/pretty v0.0.0-20160325215624-add1dbc86daf
 	github.com/google/go-cmp v0.7.0
--- a/vapi/rest/client.go
+++ b/vapi/rest/client.go
@@ -30,10 +30,16 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
+
 	"github.com/vmware/govmomi/vapi/internal"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
+)
+
+var (
+	vc8U3Ver = semver.MustParse("8.0.3-0")
 )
 
 // Client extends soap.Client to support JSON encoding, while inheriting security features, debug tracing and session persistence.
@@ -42,6 +48,8 @@ type Client struct {
 
 	*soap.Client
 	sessionID string
+
+	u3OrGreater bool
 }
 
 // Session information
@@ -66,7 +74,20 @@ func (m *LocalizableMessage) Error() string {
 func NewClient(c *vim25.Client) *Client {
 	sc := c.Client.NewServiceClient(Path, "")
 
-	return &Client{Client: sc}
+	restc := &Client{Client: sc}
+
+	// On 8.0U3 and greater, we don't need to Login to get a SessionID.
+	// Instead we can copy the SOAP session cookie
+	clientVer, err := semver.NewVersion(c.ServiceContent.About.Version)
+	if err == nil && clientVer.GreaterThanEqual(vc8U3Ver) {
+		restc.u3OrGreater = true
+
+		if c.SessionCookie() != nil && c.SessionCookie().Value != "" {
+			restc.SessionID(c.SessionCookie().Value)
+		}
+	}
+
+	return restc
 }
 
 // SessionID is set by calling Login() or optionally with the given id param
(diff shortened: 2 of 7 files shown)


vapi/rest/client.go
```diff
@@ -66,7 +74,20 @@ func (m *LocalizableMessage) Error() string {
...
+	restc := &Client{Client: sc}
+
+	// On 8.0U3 and greater, we don't need to Login to get a SessionID.
+	// Instead we can copy the SOAP session cookie
+	clientVer, err := semver.NewVersion(c.ServiceContent.About.Version)
+	if err == nil && clientVer.GreaterThanEqual(vc8U3Ver) {
+		restc.u3OrGreater = true
+
+		if c.SessionCookie() != nil && c.SessionCookie().Value != "" {
+			restc.SessionID(c.SessionCookie().Value)
```
dougm, line 86: Not sure about changing the behavior of `NewClient` and `Login`. Thinking about the various ways existing applications handle the need to re-authenticate and not wanting to break them. Using `SessionID()` here does not account for that session becoming invalid. Untested, but I'm thinking like the patch below. So if an app/caller is aware of this U3 feature, they simply don't call the `Login/Logout` methods and every rest API call will pull the current soap session from the `soap.Client.Jar`. What do you think?

```patch
modified   vapi/rest/client.go
@@ -76,7 +76,13 @@ func (c *Client) SessionID(id ...string) string {
 	if len(id) != 0 {
 		c.sessionID = id[0]
 	}
-	return c.sessionID
+	if c.sessionID != "" {
+		return c.sessionID
+	}
+	if c := c.SessionCookie(); c != nil {
+		return c.Value // use the current soap session
+	}
+	return ""
 }
```

dougm: Another version of the patch (also untested), but we should just share the CookieJar, can't think of why we didn't do that originally. Then you can use the already embedded soap.Client.SessionCookie:

```patch
diff --git a/vapi/rest/client.go b/vapi/rest/client.go
index 048b65f3..4d59048c 100644
--- a/vapi/rest/client.go
+++ b/vapi/rest/client.go
@@ -76,7 +76,13 @@ func (c *Client) SessionID(id ...string) string {
 	if len(id) != 0 {
 		c.sessionID = id[0]
 	}
-	return c.sessionID
+	if c.sessionID != "" {
+		return c.sessionID
+	}
+	if c := c.SessionCookie(); c != nil {
+		return c.Value
+	}
+	return ""
 }
 
 type marshaledClient struct {
diff --git a/vim25/soap/client.go b/vim25/soap/client.go
index b7ff99f5..d2f12b09 100644
--- a/vim25/soap/client.go
+++ b/vim25/soap/client.go
@@ -260,8 +260,8 @@ func (c *Client) newServiceClientWithTransport(path string, namespace string, t
 	}
 	c.hostsMu.Unlock()
 
-	// Copy the cookies
-	client.Client.Jar.SetCookies(u, c.Client.Jar.Cookies(u))
+	// Share the cookie Jar, it is thread safe
+	client.Client.Jar = c.Client.Jar
 
 	// Copy any query params (e.g. GOVMOMI_TUNNEL_PROXY_PORT used in testing)
 	client.u.RawQuery = vc.RawQuery
```

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/vmware/govmomi
 go 1.21 // required for slices/maps package and built-in clear function
 
 require (
+	github.com/Masterminds/semver/v3 v3.3.1
```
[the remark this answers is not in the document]
tenthirtyam: `hashicorp/go-version` is also one I recommend in the event you need and external package.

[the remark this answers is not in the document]
tenthirtyam: We should consider - future of course - if we need our own external library based on our versioning scheme that can be used by our projects. Happy to hit this one up later. LMK.

vapi/rest/client.go
rikatz: I was thinking about it. My bigger concern is about situations like a CSI/CPI case. They may connect to multiple vCenters, so in that case still for those programs to be aware or not, they would need to check the VC version before doing Login/Logout, right? I can test if just changing the approach as suggested would work.

dougm: vim25.Client and vCenter instance are 1:1, same with rest.Client, don't see multiple VCs being a problem here.
CSI/CPI/etc apps would only need be aware of VC version if they choose to opt-into this session sharing.
Otherwise the existing pattern of Login+Logout to the rest endpoint will continue to work the same way.

rikatz:
folks, I've been on vacation and am coming back kind of now. will be back to this PR soon, sorry for the delay

PR #3707, opened 2025-03-13
api: Add optional CategoryID,TagID fields for category and tag creation

### LINKED ISSUE #3706: Add categoryID option when creating a category, tagID when creating tag.

I would like to replicate tagging across multiple vCenter Server instances.
If I am unable to create a category with the same ID as the source, then the linkage of the tags to categories needs to be manually adjusted.

https://developer.broadcom.com/xapis/vsphere-automation-api/latest/cis/api/cis/tagging/category/post/
This mentions an optional category_id parameter.
I would like to be able to provide that to the CreateCategory request.
I believe currently it excludes that value in the create struct.

Alternatively I just manually make the REST calls.

dougm:
- govc: Add '-id' option for tags.category and tags create commands
- vcsim: Add vapi support for CategoryID and TagID fields

Fixes #3706

### DIFF
--- a/cli/tags/category/create.go
+++ b/cli/tags/category/create.go
@@ -1,18 +1,6 @@
-/*
-Copyright (c) 2018 VMware, Inc. All Rights Reserved.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


prziborowski (approved):
Thank you very much for addressing this!

PR #3726, opened 2025-03-18
govc: fix integer type conversion

tenthirtyam:
## Description

Added a check to panic if a value exceeds `math.MaxInt32` in the `Interval` method to prevent out-of-range errors.

### DIFF
--- a/cli/metric/performance.go
+++ b/cli/metric/performance.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math"
 	"strconv"
 
 	"github.com/vmware/govmomi/cli/flags"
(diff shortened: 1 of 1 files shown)


cli/metric/performance.go
```diff
@@ -88,7 +89,12 @@ func (f *PerformanceFlag) Interval(val int32) int32 {
 			if err != nil {
 				panic(err)
 			}
-			interval = int32(n)
+
+			if n > math.MaxInt32 {
+				panic(fmt.Errorf("value out of range for int32: %d", n))
```
dougm, line 94: Since we're parsing user input here.. we should change `Interval()` to return `(int32, error)`, rather than the existing panic() and adding another. But this hasn't been raised as an issue before either, find to keep as-is.

dougm (approved):
Thanks @tenthirtyam

Issue #3728, opened 2025-03-18
Support using pull upload mode when importing a OVA


luisdavim:
The `ovftool` has a flag (`--pullUploadMode`) than when the argument for the OVA path is a URL allows skipping the local download to upload and the download is done directly by the ESXi host, it would be nice if this library would allow the same.

dougm:
That would be nice, similar to govc library.import `-pull`. Hadn't seen this before, but looks like we'd need to use the [pullFromUrls](https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U3/html/ReferenceGuides/vim.HttpNfcLease.html#pullFromUrls) method

PR #3758, opened 2025-04-08
govc: Add datastore.cp disk format and adapter options

### LINKED ISSUE #3748: vmfktools Copy API too strict.
I Want to copy a thin provisioned disk on a datastore to a new disk. The new disk is always thick provisioned. I want the new disk to be thin provisioned too.
Right now converting a disk to thin from thick is only available with VMotion if we do not have ssh access to the esxi hosts. This makes development process horrible

I want to be able to tell the govc or govmomi cli what type of disk it should create.

Currently the datastore.disk.create has no option to create a disk from another disk or file. Add the feature to datastore.disk.create so it can also copy disks.

**Additional context**
This line should be modified:
https://github.com/vmware/govmomi/blob/d62c8e932b6dc24e4a895f2db21fd7e5fc043930/object/datastore_file_manager.go#L127
So instead of static nill parameter for type for the vmdk creation, propagate the parameter with a default value (to keep other calls intact) and:
https://github.com/vmware/govmomi/blob/d62c8e932b6dc24e4a895f2db21fd7e5fc043930/cli/datastore/cp.go#L72
here also open up the parameter for the datastore.cp cli and set the default parameter inside the arguments.

This feature was already present under a different cli command (datastore.disk.cp) #999 although this seems like it was deleted.

dougm:
Same '-a' and '-d' flags as datastore.disk.create

Fixes #3748

api: Add optional spec param to DatastoreFileManager.Copy

### DIFF
--- a/object/datastore_file_manager.go
+++ b/object/datastore_file_manager.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/vmware/govmomi/vim25/progress"
 	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
 )
 
 // DatastoreFileManager combines FileManager and VirtualDiskManager to manage files on a Datastore
@@ -115,16 +116,21 @@ func (m *DatastoreFileManager) CopyFile(ctx context.Context, src string, dst str
 }
 
 // Copy dispatches to the appropriate FileManager or VirtualDiskManager Copy method based on file name extension
-func (m *DatastoreFileManager) Copy(ctx context.Context, src string, dst string) error {
+func (m *DatastoreFileManager) Copy(ctx context.Context, src string, dst string, spec ...types.BaseVirtualDiskSpec) error {
 	srcp := m.Path(src)
 	dstp := m.Path(dst)
 
 	f := m.FileManager.CopyDatastoreFile
 
 	if srcp.IsVMDK() {
+		var dstSpec types.BaseVirtualDiskSpec
+		if len(spec) != 0 {
+			dstSpec = spec[0]
+		}
+
 		// types.VirtualDiskSpec=nil as it is not implemented by vCenter
 		f = func(ctx context.Context, src string, srcDC *Datacenter, dst string, dstDC *Datacenter, force bool) (*Task, error) {
-			return m.VirtualDiskManager.CopyVirtualDisk(ctx, src, srcDC, dst, dstDC, nil, force)
+			return m.VirtualDiskManager.CopyVirtualDisk(ctx, src, srcDC, dst, dstDC, dstSpec, force)
 		}
 	}
 
--- a/cli/datastore/cp.go
+++ b/cli/datastore/cp.go
@@ -12,6 +12,7 @@ import (
 	"github.com/vmware/govmomi/cli"
 	"github.com/vmware/govmomi/cli/flags"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
 )
 
 type cp struct {
@@ -30,6 +31,7 @@ type target struct {
 
 	kind  bool
 	force bool
+	spec  types.VirtualDiskSpec
 }
 
 func (cmd *target) FileManager() (*object.DatastoreFileManager, error) {
@@ -71,6 +73,8 @@ func (cmd *target) Register(ctx context.Context, f *flag.FlagSet) {
 
 	f.BoolVar(&cmd.kind, "t", true, "Use file type to choose disk or file manager")
 	f.BoolVar(&cmd.force, "f", false, "If true, overwrite any identically named file at the destination")
+	f.StringVar(&cmd.spec.AdapterType, "a", string(types.VirtualDiskAdapterTypeLsiLogic), "Disk adapter")
+	f.StringVar(&cmd.spec.DiskType, "d", string(types.VirtualDiskTypeThin), "Disk format")
 }
 
 func (cmd *target) Process(ctx context.Context) error {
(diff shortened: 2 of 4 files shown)


object/datastore_file_manager.go
```diff
@@ -115,16 +116,21 @@ func (m *DatastoreFileManager) CopyFile(ctx context.Context, src string, dst str
...
 	if srcp.IsVMDK() {
+		var destSpec types.BaseVirtualDiskSpec
+		if len(spec) != 0 {
+			destSpec = spec[0]
+		}
+
 		// types.VirtualDiskSpec=nil as it is not implemented by vCenter
 		f = func(ctx context.Context, src string, srcDC *Datacenter, dst string, dstDC *Datacenter, force bool) (*Task, error) {
-			return m.VirtualDiskManager.CopyVirtualDisk(ctx, src, srcDC, dst, dstDC, nil, force)
+			return m.VirtualDiskManager.CopyVirtualDisk(ctx, src, srcDC, dst, dstDC, destSpec, force)
```
tenthirtyam, line 133: Would it be better to use `dstSpec` for uniformity?

tenthirtyam (dismissed):
@dougm I added a comment for a minor nit, but otherwise LGTM.

object/datastore_file_manager.go
dougm: Yes, thanks, I'll fix that

tenthirtyam (approved):
LGTM!

object/datastore_file_manager.go
```diff
@@ -115,16 +116,21 @@ func (m *DatastoreFileManager) CopyFile(ctx context.Context, src string, dst str
...
 
 	f := m.FileManager.CopyDatastoreFile
 
 	if srcp.IsVMDK() {
+		var dstSpec types.BaseVirtualDiskSpec
+		if len(spec) != 0 {
+			dstSpec = spec[0]
+		}
+
 		// types.VirtualDiskSpec=nil as it is not implemented by vCenter
```
dougm, line 131: Just noticed my own comment here from ~7 years ago. And I'm not seeing a difference now with the spec in place, it seems to be ignored. The target disk is always the same format as the source, regardless of specifying a different `-d` (type). For example copying from `thin` to `preallocated` results in `thin`:
```console
% govc datastore.mkdir disks

% govc datastore.disk.create -size 16G -d thin disks/thin-disk-src.vmdk
Creating [sharedVmfs-0] disks/thin-disk-src.vmdk...OK

% govc datastore.cp -d preallocated disks/thin-disk-src.vmdk disks/thick-disk-dst.vmdk
Copying [sharedVmfs-0] disks/thin-disk-src.vmdk to [sharedVmfs-0] disks/thick-disk-dst.vmdk...OK

% govc datastore.disk.info disks/thick-disk-dst.vmdk
Name:      disks/thick-disk-dst.vmdk
  Type:    thin
```
I'll hold off merging for now, will take a closer look at some point too. cc @aklakina

aklakina: Are you sure that is the case?

For example in my use case, I upload a .vmdk to vsphere and then I use datastore.cp to copy it with auto type detection. The disk is created with starwind v2v as file type vsphere_ws_growable, which should be the thin provisioned. This is then uploaded to vsphere esxi host. (I know that esxi is not at all the same as a vsphere workstation, but because of security purposes, the machine running the starwind can not connect directly to esxi host, so it can not use the vsphere_esxi_growable output type.)

So the disk uploaded is thin provisioned (confirmed by size) but the result of cp is eager zeroed. Could the issue be somewhere else?

I haven't tried the changes yet. Will only have time at the start of next week.

dougm: I've only tested source disk from `datastore.disk.create`, not tried starwind. Ok, let us know how it goes when you are able to test. I'll still try to take a closer look too.

aklakina: I've ran the test and can confirm that
`govc datastore.cp <source> <dest>`
creates an eager zeroed disk on this branch too.

dougm: > This is then uploaded to vsphere esxi host
How was it uploaded? scp or ?
Another option to try: https://github.com/vmware/govmomi/blob/main/govc/USAGE.md#importvmdk
This goes through the conversion process, from desktop (streamOptimized) format to thin. After that, you can use `datastore.cp` and target disks will be thin format.

aklakina: It was converted to vsphere workspace growable format and then I used the datastore.upload command.

Sadly the import.vmdk command needs direct access to the esxi hosts, which would result in some security valnurabilities in our system, so that is not an option.

dougm: We could make import.vmdk work with Content Library, then upload would be to vCenter rather than ESX. When a vmdk is imported to CL, it will be converted, so datastore.cp target should also be thin.

Or if we added an option to import.{vmdk,ovf} to [pullFromUrls](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.HttpNfcLease.html#pullFromUrls) , then vCenter would pull from the provided URL, rather than client uploading to ESX. Planning to do that for #3728

aklakina: That is a neat feature. Sadly it would complicate our workflow probably too much so I would prefer the first option.

aklakina: > We could make import.vmdk work with Content Library, then upload would be to vCenter rather than ESX. When a vmdk is imported to CL, it will be converted, so datastore.cp target should also be thin.
> 
> Or if we added an option to import.{vmdk,ovf} to [pullFromUrls](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.HttpNfcLease.html#pullFromUrls) , then vCenter would pull from the provided URL, rather than client uploading to ESX. Planning to do that for #3728

Hi @dougm 

Any news on any of these features? I would mostly be interested in the first option, making import.vmdk work with content library. Do you have a timeline for this?

PR #3767, opened 2025-04-15
api: Rewrite lookup service URLs on DNS error

dougm:
If a vCenter's hostname is not in DNS (outside of VC's /etc/hosts), then lookup service URLs may fail to resolve. In this case, rewrite the URLs to use the same Host used for the vim25.Client's connection.

### DIFF
--- a/lookup/client.go
+++ b/lookup/client.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"log"
 	"net/url"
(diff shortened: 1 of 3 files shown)


Dei405 (approved):

lookup/client.go
```diff
@@ -59,22 +64,38 @@ func NewClient(ctx context.Context, c *vim25.Client) (*Client, error) {
...
+			return nil, err
+		}
+
+		client.ServiceContent = res.Returnval
+
+		return client, nil
+	}
 
-	return client, nil
+	panic("not reached")
```
Dei405, line 98: Thanks!  @dougm 

It might be helpful to include additional context in the panic message, even though this code path is not expected to be executed under normal circumstances:

panic(fmt.Sprintf(`ssoadmin.NewClient failed:
 - Initial attempt to use derived host failed
 - Second attempt with vim25 client host (%s) also failed
 - Final path tried: %s`, c.URL().Host, path.String()))

dougm: I removed the panic and changed the logic so all errors are returned

sidharthsurana (approved):

PR #3784, opened 2025-05-15
vcsim: Use OptionManager to configure session timeout duration

dougm:
Use of SessionIdleTimeout package var can race.
Changing session timeout via OptionManager:
- Avoids races
- Works with clients other than Go
- Same method used by real vCenter

BREAKING: var `simulator.SessionIdleTimeout` has been removed.
Use `sim25.SetSessionTimeout` instead.

Example against real vCenter:
```console
% govc option.set config.vmacore.soap.sessionTimeout 42
```

Setting is persisted here:
```console
# grep sessionTimeout -B1 -A1 /etc/vmware-vpx/vpxd.cfg
    <soap>
      <sessionTimeout>42</sessionTimeout>
    </soap>
```

### DIFF
--- a/session/keepalive/example_test.go
+++ b/session/keepalive/example_test.go
@@ -10,9 +10,10 @@ import (
 	"log"
 	"time"
 
-	"github.com/vmware/govmomi/session"
+	"github.com/vmware/govmomi"
 	"github.com/vmware/govmomi/session/keepalive"
 	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/simulator/sim25"
 	"github.com/vmware/govmomi/vapi/rest"
 	"github.com/vmware/govmomi/vim25"
 )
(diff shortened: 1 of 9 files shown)


session/manager.go
```diff
@@ -74,6 +74,30 @@ func (sm *Manager) SetLocale(ctx context.Context, locale string) error {
 	return err
 }
 
+// SetSessionTimeout changes vCenter's default session timeout (in minutes).
+// Changes HostAgent's session timeout when connected to an ESX host.
+// Note that vcsim accepts a time.Duration.String(),
+// but vCenter+ESX only accept an integer string, in minutes.
+func (sm *Manager) SetSessionTimeout(ctx context.Context, minutes string) error {
```
mikekim, line 81: I'm wondering if this is the proper interface. I understand that internally config uses strings for minutes but I'm wondering if perhaps `SetSessionTimeout` should take `Duration` and internally convert it to `string`?

session/keepalive/example_test.go
```diff
@@ -23,14 +24,21 @@ var (
...
-	simulator.SessionIdleTimeout = sessionIdleTimeout
-}
-
 func ExampleHandlerSOAP() {
-	simulator.Run(func(ctx context.Context, c *vim25.Client) error {
-		// No need for initial Login() here as simulator.Run already has
-		m := session.NewManager(c)
+	simulator.Run(func(ctx context.Context, vc *vim25.Client) error {
+		// Using the authenticated vc client, timeout will apply to new sessions.
+		session.NewManager(vc).SetSessionTimeout(ctx, sessionIdleTimeout.String())
```
mikekim, line 30: IIUC `sessionIdleTimeout` is `250ms` and will be represented that way when converted to `string`, yet the comment for `SetSessionTimeout` says that it expects the value in minutes? Based on that comment for `SetSessionTimeout` I'd expect the value passed in here to be `1` or `0.001`.

mikekim (approved):

session/manager.go
dougm: I considered doing that, problem is real VC can't parse fractions, so we'd have to round up to the nearest whole second. Don't really want to add a client-side check there to see if we're talking to vcsim. Thinking I'll move this helper to the simulator package and change it to use `time.Duration`, making it clear in the doc it's only meant for use with vcsim. It'll still be a handy 1-liner for tests and example for porting to other languages using vcsim. What do you think?

```console
% govc option.set config.vmacore.soap.sessionTimeout .250 # against real VC
govc: ServerFaultCode: A general system error occurred: Config Value not an integer
```

mikekim: Sounds good to me. I like the idea of having this function specialized for vcsim.

dougm: ok, done. Created a new `simulator/sim25` package, have some other helpers in mind that could live there.

ridaz (approved):
LGTM, Thanks.

tenthirtyam (approved):

PR #3787, opened 2025-05-21
vcsim: Add object.VirtualMachine.CreateSnapshotEx support

sreyasn:
## Description

Add CreateSnapshotEx method to object.VirtualMachine Api
vcsim: Add object.VirtualMachine.CreateSnapshotEx support
## How Has This Been Tested?
-unit tests

### DIFF
--- a/object/virtual_machine.go
+++ b/object/virtual_machine.go
@@ -649,7 +649,7 @@ func (v VirtualMachine) CreateSnapshot(ctx context.Context, name string, descrip
 }
 
 // CreateSnapshotEx creates a new snapshot of a virtual machine.
-func (v VirtualMachine) CreateSnapshotEx(ctx context.Context, name string, description string, memory bool, quiesceSpec *types.VirtualMachineGuestQuiesceSpec) (*Task, error) {
+func (v VirtualMachine) CreateSnapshotEx(ctx context.Context, name string, description string, memory bool, quiesceSpec types.BaseVirtualMachineGuestQuiesceSpec) (*Task, error) {
 	req := types.CreateSnapshotEx_Task{
 		This:        v.Reference(),
 		Name:        name,
--- a/object/virtual_machine.go
+++ b/object/virtual_machine.go
@@ -648,6 +648,24 @@ func (v VirtualMachine) CreateSnapshot(ctx context.Context, name string, descrip
 	return NewTask(v.c, res.Returnval), nil
 }
 
+// CreateSnapshotEx creates a new snapshot of a virtual machine.
+func (v VirtualMachine) CreateSnapshotEx(ctx context.Context, name string, description string, memory bool, quiesceSpec *types.VirtualMachineGuestQuiesceSpec) (*Task, error) {
+	req := types.CreateSnapshotEx_Task{
+		This:        v.Reference(),
+		Name:        name,
+		Description: description,
+		Memory:      memory,
+		QuiesceSpec: quiesceSpec,
+	}
+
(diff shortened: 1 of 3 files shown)


simulator/virtual_machine.go
```diff
@@ -2747,6 +2747,72 @@ func (vm *VirtualMachine) CreateSnapshotTask(ctx *Context, req *types.CreateSnap
 	}
 }
 
+func (vm *VirtualMachine) CreateSnapshotExTask(ctx *Context, req *types.CreateSnapshotEx_Task) soap.HasFault {
```
dougm, line 2750: Can we change the existing `CreateSnapshotTask` to call this method?
We do this for example with `RetrieveProperties` calling `RetrievePropertiesEx`:
https://github.com/vmware/govmomi/blob/49947e5bbd7b83e91cf89dc2ba80daae72dddfff/simulator/property_collector.go#L715

object/virtual_machine.go
```diff
@@ -648,6 +648,24 @@ func (v VirtualMachine) CreateSnapshot(ctx context.Context, name string, descrip
 	return NewTask(v.c, res.Returnval), nil
 }
 
+// CreateSnapshotEx creates a new snapshot of a virtual machine.
+func (v VirtualMachine) CreateSnapshotEx(ctx context.Context, name string, description string, memory bool, quiesceSpec *types.VirtualMachineGuestQuiesceSpec) (*Task, error) {
```
dougm, line 652: `quiesceSpec` should be `types.BaseVirtualMachineGuestQuiesceSpec` so [VirtualMachineWindowsQuiesceSpec](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.WindowsQuiesceSpec.html) can also be passed here.

Tempting to change the existing `CreateSnapshot` to call `CreateSnapshotEx`, and make the breaking change to the `quiesceSpec` param (bool -> types.BaseVirtualMachineGuestQuiesceSpec). I think govmomi's use of `CreateSnapshot` goes back to vSphere 5.0, `CreateSnapshotEx` was added in 6.5 and we're not supporting vSphere < 7.0 in govmomi anymore.

simulator/virtual_machine_test.go
```diff
@@ -1820,6 +1820,149 @@ func TestVmSnapshot(t *testing.T) {
...
+	}
+
+	s := m.Service.NewServer()
+	defer s.Close()
+
+	c, err := govmomi.NewClient(ctx, s.URL, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
dougm, line 1840: Instead of the boilerplate above, you can use the new `Test` helper:
```go
Test(func(ctx context.Context, c *vim25.Client) {
    ...
}, ESX())
```

Example: https://github.com/vmware/govmomi/blob/main/simulator/folder_test.go#L633

object/virtual_machine.go
sreyasn: Thanks Doug. I've made the change to types.BaseVirtualMachineGuestQuiesceSpec. Good catch on the WindowsQuiesceSpec.

As for changing the existing CreateSnapshot to call CreateSnapshotEx. From looking at our VMODL docs, CreateSnapshot wasn't deprecated until 8.0 GA.

dougm: ok thanks, let's leave existing CreateSnapshot as-is then

dougm (approved):
lgtm, thanks @sreyasn !

PR #3806, opened 2025-06-22
skip CNS transaction tests when env CNS_RUN_TRANSACTION_TESTS is not set

divyenpatel:
## Description

FSS for CNS transaction API change is not enabled and API is under development.
We need an option to skip running Transaction related tests.
when env `CNS_RUN_TRANSACTION_TESTS` is not set to `true`, tests related CNS transaction API are skipped.

This PR also updates `CnsVolumeCreateSpec` and `CnsSnapshotCreateSpec` by changing `VolumeId` and `SnapshotId` to pointer types. This change is necessary to ensure these fields can be omitted when not provided, enabling correct serialization behavior and allowing the API to function as expected when VolumeId or SnapshotId is not supplied.

## How Has This Been Tested?
Verified executing tests without setting env CNS_RUN_TRANSACTION_TESTS to true.

### DIFF
--- a/cns/types/types.go
+++ b/cns/types/types.go
@@ -79,7 +79,7 @@ type CnsVolumeCreateSpec struct {
 	types.DynamicData
 	Name                 string                                `xml:"name" json:"name"`
 	VolumeType           string                                `xml:"volumeType" json:"volumeType"`
-	VolumeId             CnsVolumeId                           `xml:"volumeId,omitempty" json:"volumeId"`
+	VolumeId             *CnsVolumeId                          `xml:"volumeId,omitempty" json:"volumeId"`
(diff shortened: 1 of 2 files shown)


cns/types/types.go
```diff
@@ -725,9 +725,9 @@ type CnsCreateSnapshotsResponse struct {
 type CnsSnapshotCreateSpec struct {
 	types.DynamicData
 
-	VolumeId    CnsVolumeId   `xml:"volumeId" json:"volumeId"`
-	Description string        `xml:"description" json:"description"`
-	SnapshotId  CnsSnapshotId `xml:"snapshotId,omitempty" json:"snapshotId"`
+	VolumeId    CnsVolumeId    `xml:"volumeId" json:"volumeId"`
```
skogta, line 728: There seems to be some extra space that got added here.

PR #3824, opened 2025-07-23
Add tests for types of meta fault

hickeng:
## Description

This adds compound tests for whether a fault is a transient error. Variants of this test show up in various locations
in client code so pull into the library. This also doubles as a reference for what transient error conditions exist as
not all of the client tests detail all of these error types.

Additionally this adds a test for whether an InvalidPowerState error is due to the VM already being in the target state.
Checking this is a common pattern when trying to map desired state to imperative API.
This is specifically for PowerOff as I don't yet have a need for the others and didn't want to make the test generic to
the point that it takes a target powerstate as an argument, both to keep callers concise and to avoid a caller needing
to extract possible values from the enum.

## How Has This Been Tested?

New unit test, all unit tests passed

## Guidelines


akutz:
Hey @hickeng,

You need to ensure one file is properly `go fmt`'d, and also, please use the `--sign-off` flag on your git commit. This is not about signing a commit, but signing off. It is confusingly named and different. See https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff.

hickeng:
> Hey @hickeng,
> 
> You need to ensure one file is properly `go fmt`'d, and also, please use the `--sign-off` flag on your git commit. This is not about signing a commit, but signing off. It is confusingly named and different. See https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff.

Interesting - I have vscode set to run goimports on save so clearly I've got a tool version mismatch or something odd going on with language server.
Added signoff.

PR #3850, opened 2025-09-02
Add PodVMInfo type and method

rohith05:
## Description

This patch adds support to PodVMInfo VMODL APIs under HostRuntimeInfo which includes hasPodVM and podVMOverheadInfo properties. In addition this patch also adds UpdatePodVMProperty method under HostSystem. The PodVMInfo property and method will be updated by ESXi host and will be consumed by DRS to determine PodVM placement. The change also includes a unit test that tests the setting of PodVMInfo property.

Note: PodVMOverheadInfo that is currently exposed as part of HostCapability will be removed
as a follow-up change, once existing consumers have moved away from it.

## How Has This Been Tested?
- make test:
  === RUN   TestPodVMInfo
  --- PASS: TestPodVMInfo (0.69s)
- Verified UpdatePodVMProperty() method from ESXi:
  --- PASS: TestSetOverheadVMODLProperty (0.20s)
  PASS

### DIFF
--- a/object/host_system.go
+++ b/object/host_system.go
@@ -148,20 +148,17 @@ func (h HostSystem) UpdatePodVMProperty(ctx context.Context, propertyPath string
 		PropertyPath: propertyPath,
 	}
 
-	if propertyPath == "podVMOverheadInfo" {
+	switch propertyPath {
+	case "podVMOverheadInfo":
 		req.Property = podVMInfo.PodVMOverheadInfo
-	} else if propertyPath == "hasPodVM" {
+	case "hasPodVM":
 		req.Property = podVMInfo.HasPodVM
-	} else if propertyPath == "podVMInfo" {
+	case "podVMInfo":
 		req.Property = podVMInfo
-	} else {
+	default:
 		return fmt.Errorf("unsupported propertyPath: %s", propertyPath)
 	}
 
 	_, err := methods.UpdatePodVMProperty(ctx, h.c, &req)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
--- a/object/host_system.go
+++ b/object/host_system.go
@@ -140,3 +140,28 @@ func (h HostSystem) ExitMaintenanceMode(ctx context.Context, timeout int32) (*Ta
 
 	return NewTask(h.c, res.Returnval), nil
 }
+
+func (h HostSystem) UpdatePodVMProperty(ctx context.Context, propertyPath string,
+	podVMInfo types.HostRuntimeInfoPodVMInfo) error {
+	req := types.UpdatePodVMProperty{
+		This:         h.Reference(),
+		PropertyPath: propertyPath,
+	}
+
+	if propertyPath == "podVMOverheadInfo" {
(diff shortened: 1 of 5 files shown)


object/host_system.go
```diff
@@ -140,3 +140,28 @@ func (h HostSystem) ExitMaintenanceMode(ctx context.Context, timeout int32) (*Ta
...
 }
+
+func (h HostSystem) UpdatePodVMProperty(ctx context.Context, propertyPath string,
+	podVMInfo types.HostRuntimeInfoPodVMInfo) error {
+	req := types.UpdatePodVMProperty{
+		This:         h.Reference(),
+		PropertyPath: propertyPath,
+	}
+
+	if propertyPath == "podVMOverheadInfo" {
```
akutz, line 151: This seems like a good use for a switch statement instead of so many if/elses.

object/host_system.go
```diff
@@ -140,3 +140,28 @@ func (h HostSystem) ExitMaintenanceMode(ctx context.Context, timeout int32) (*Ta
...
+		req.Property = podVMInfo.PodVMOverheadInfo
+	} else if propertyPath == "hasPodVM" {
+		req.Property = podVMInfo.HasPodVM
+	} else if propertyPath == "podVMInfo" {
+		req.Property = podVMInfo
+	} else {
+		return fmt.Errorf("unsupported propertyPath: %s", propertyPath)
+	}
+
+	_, err := methods.UpdatePodVMProperty(ctx, h.c, &req)
```
akutz, line 161: Rewrite as:

```golang
_, err := methods.UpdatePodVMProperty(ctx, h.c, &req)
return err
```

vim25/types/types.go
```diff
@@ -44786,6 +44786,8 @@ type HostRuntimeInfo struct {
 	PartialMaintenanceMode []HostPartialMaintenanceModeRuntimeInfo `xml:"partialMaintenanceMode,omitempty" json:"partialMaintenanceMode,omitempty" vim:"8.0.3.0"`
 	// Host persistent state encryption information.
 	StateEncryption *HostRuntimeInfoStateEncryptionInfo `xml:"stateEncryption,omitempty" json:"stateEncryption,omitempty" vim:"7.0.3.0"`
+	// PodVM related info for a host
+	PodVMInfo *HostRuntimeInfoPodVMInfo `xml:"podVMInfo,omitempty" json:"podVMInfo,omitempty" vim:"9.0.0.0"`
```
akutz, line 44790: Hmm, you have updated a generated type. I'm not sure we can support this.

akutz (changes requested):
@rohith05 ,

I'm not sure we can support this change in the public repo as you've updated a generated type that will be wiped once the types are regenerated. You may need to make this change in the internal GoVmomi mirror.

vim25/types/types.go
rohith05: This type is already part of HostRuntime VMODL (on main/9.1). Does that mean when HostRuntime is regenerated this property (PodVMInfo) will be re-added ?

akutz (approved):

PR #3903, opened 2025-11-16
Add property-based filtering capabilities and skipCount functionality to injection rules.

hickeng:
# Summary
Adds two enhancements to the fault injection rules:
1. Property Filter Support - inclusion and exclusion callback filters, with
    convenience functions for expressing AND or OR matches using property.Match syntax
3. Skip Count Functionality - skip the first N matches before injecting faults

# Details
- Property filters are evaluated after ObjectName matching but before probability checks
- Skip count works in the same way as maxCount, and is reset together with it
- Skip and max counts are incremented only if the rule makes it through the filters

Eg.
```
rule := &FaultInjectionRule{
    MethodName: "PowerOnVM_Task",
    InclusionPropertyFilter: FaultFilterAll(
        "config.name == DC0*",
        "runtime.powerState == poweredOff"
    ),
    ExclusionPropertyFilter: FaultFilterAny("config.name == *-test"),
    SkipCount: 2, // Skip first 2 matches
    MaxCount: 3,  // Then inject 3 times
    Probability: 1.0,
    Enabled: true,
```

# Testing

Added passing unit tests for all new functionality

### DIFF
--- a/simulator/fault_injection.go
+++ b/simulator/fault_injection.go
@@ -259,7 +259,7 @@ func (f *FaultInjector) GetRules() []*FaultInjectionRule {
 }
 
 // ShouldInjectFault determines if a fault should be injected for the given method call
-func (f *FaultInjector) ShouldInjectFault(method *Method, objectName string, handler mo.Reference) *FaultInjectionRule {
+func (f *FaultInjector) ShouldInjectFault(method *Method, handler mo.Reference) *FaultInjectionRule {
 	f.mu.RLock()
 	defer f.mu.RUnlock()
 
@@ -284,8 +284,15 @@ func (f *FaultInjector) ShouldInjectFault(method *Method, objectName string, han
 		}
 
 		// Check object name match (order 1)
-		if rule.ObjectName != "*" && rule.ObjectName != objectName {
-			continue
+		if rule.ObjectName != "*" {
+			var objectName string
+			if entity, ok := handler.(mo.Entity); ok {
+				objectName = entityName(entity)
+			}
+
+			if rule.ObjectName != objectName {
+				continue
+			}
 		}
 
 		// Check inclusion property filter (order 2)
--- a/simulator/simulator.go
+++ b/simulator/simulator.go
@@ -250,13 +250,8 @@ func (s *Service) call(ctx *Context, method *Method) soap.HasFault {
 	}
 
 	// We have a valid call. Check for fault injection first
-	var objectName string
-	if entity, ok := handler.(mo.Entity); ok {
-		objectName = entityName(entity)
-	}
-
 	if s.faultInjector != nil {
-		if rule := s.faultInjector.ShouldInjectFault(method, objectName, handler); rule != nil {
+		if rule := s.faultInjector.ShouldInjectFault(method, handler); rule != nil {
 			return s.faultInjector.CreateFault(rule, method)
 		}
 	}


simulator/fault_injection.go
```diff
@@ -30,6 +33,32 @@ type FaultInjectionRule struct {
 	// Use "*" to match all object names
 	ObjectName string
 
+	// InclusionPropertyFilter is an array of property filters to match the object.
+	// ObjectName above is a convenience as it's such a common filter, but
+	// could be expressed here. All filters in the array must match (AND).
+	// Supported operators:
```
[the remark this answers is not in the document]
dougm: I think it would be good to support expressing both, the issue I linked to was about adding !=
See also the govc find examples: https://github.com/vmware/govmomi/blob/main/govc/USAGE.md#find
`govc find . -type m -runtime.powerState poweredOn`

But if we're going to add support for `!=`, we should also consider other operators.
See the FCD query operators that can be used with `govc disk.ls`:
https://github.com/vmware/govmomi/blob/08dfb41f9f31c0614e1f928748475f8ae8832166/cli/disk/ls.go#L107-L115
(which were borrowed from PTQL: https://github.com/hyperic/sigar/blob/master/src/sigar_ptql.c#L203)

So maybe we extend property.Match to support this set of operators, which cover most use-cases. Reads well (imho) and shell friendly, for example:
```
govc find . -type m -runtime.powerState.ne poweredOn # VMs in state NOT poweredOn
govc find . -type h -hardware.cpuInfo.numCpuCores.ge 16 # >= 16
```

simulator/fault_injection.go
```diff
@@ -320,5 +451,80 @@ func (f *FaultInjector) ResetStats() {
...
+		}
+
+		propertyPath := strings.TrimSpace(parts[0])
+		expectedValue := strings.TrimSpace(parts[1])
+		if propertyPath == "" {
+			panic(fmt.Sprintf("empty property path in filter %q - expected format: \"property.path == value\"", filterStr))
+		}
+		// nil expected values are ok - that's how we expect to match empty properties
+
+		// Convert property path to camelCase (property collector format)
```
dougm: Why not just take camelCase as input, instead of converting here? We take camelCase as input elsewhere.

simulator/fault_injection.go
```diff
@@ -7,13 +7,95 @@ package simulator
...
+// See FaultFilterAll for details.
+func FaultFilterAny(filterStrs ...string) FaultFilterCallback {
+	// For exclusion matches, we need to handle each filter separately
+	// since multiple matches can target the same property path with different values
+	matches := make([]property.Match, 0, len(filterStrs))
+	allPaths := make(map[string]bool)
+	for _, filterStr := range filterStrs {
+		match := parsePropertyFilters([]string{filterStr})
+		if len(match) > 0 {
+			matches = append(matches, match)
```
dougm: You allocated (make) matches on line 62, so this can be `matches[i] = match`.

simulator/fault_injection.go
hickeng: It's easy to convert and just one less format thing for a user to worry about.
I'm happy to change it to require camelCase as the input form or merge as is. Preference?

It's not blocking yet - my other work is stacked on top of my hickeng branch, but that's some way off merging and this only blocks the merge, not the dev/test.

dougm (approved):

PR #3904, opened 2025-11-19
api: Fix OVF defVal & ConfigSpec vAppConfig

akutz:
## Description

This patch addresses some bugs in the way OVFs are unmarshalled as well as transformed into ConfigSpec objects. For some reason the original OVF unmarshalling code ignored the fact that properties have a "value" attribute and not just sub-elements. In fact, the latter is not even in the DMTF OVF spec. There was also a bug in how the default value and value were treated as the same, when they are not.

Closes: `NA`

## How Has This Been Tested?

Added unit tests and enhanced existing ones. Just run `go test -v -count=1 ./ovf`

## Guidelines

### DIFF
--- a/ovf/envelope.go
+++ b/ovf/envelope.go
@@ -111,14 +111,15 @@ type Property struct {
 	Type             string  `xml:"type,attr" json:"type,omitempty"`
 	Qualifiers       *string `xml:"qualifiers,attr" json:"qualifiers,omitempty"`
 	UserConfigurable *bool   `xml:"userConfigurable,attr" json:"userConfigurable,omitempty"`
-	Default          *string `xml:"value,attr" json:"default,omitempty"`
+	Default          *string `xml:"defaultValue,attr" json:"defaultValue,omitempty"`
 	Password         *bool   `xml:"password,attr" json:"password,omitempty"`
 	Configuration    *string `xml:"configuration,attr" json:"configuration,omitempty"`
+	Value            *string `xml:"value,attr" json:"value,omitempty"`
 
 	Label       *string `xml:"Label" json:"label,omitempty"`
 	Description *string `xml:"Description" json:"description,omitempty"`
 
-	Values []PropertyConfigurationValue `xml:"Value" json:"value,omitempty"`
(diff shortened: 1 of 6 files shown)


ovf/envelope.go
```diff
@@ -111,14 +111,15 @@ type Property struct {
...
+	Default          *string `xml:"defaultValue,attr" json:"defaultValue,omitempty"`
 	Password         *bool   `xml:"password,attr" json:"password,omitempty"`
 	Configuration    *string `xml:"configuration,attr" json:"configuration,omitempty"`
+	Value            *string `xml:"value,attr" json:"value,omitempty"`
 
 	Label       *string `xml:"Label" json:"label,omitempty"`
 	Description *string `xml:"Description" json:"description,omitempty"`
 
-	Values []PropertyConfigurationValue `xml:"Value" json:"value,omitempty"`
+	Values []PropertyConfigurationValue `xml:"Value" json:"values,omitempty"`
```
aruneshpa: Not your change, but I am wondering why the casing of the XML tag is weird between `Values` and `Value`.

akutz: It *is* my change. because I had to change it. Apparently this is valid in an OVF according to this code's original author, Pieter:

```xml
<Property ovf:key="my-key" ovf:type="string" ovf:userConfigurable="true" ovf:value="my-value">
  <Value value="my-value-for-hello-config" configuration="hello" />
</Property>
```

Even though I see no mention of this in the DMTF OVF spec at https://www.dmtf.org/sites/default/files/standards/documents/DSP0243_2.1.1.pdf.

Anyway, because of the above, because we've never actually parsed `value` correctly from an OVF, I had to add the field to the `PropertyInfo` type:

```go
Value *string `xml:"value,attr" json:"value,omitempty"`
```

Because of this, Go won't let me use the JSON `value` tag name for both the new field and existing `Values` tag, so I elected to change the JSON tag for the plural field to `values`.

aruneshpa: Got it.  Okay.  TIL about Values in OVF XML.  I was only aware of prop key and value.  Thanks for the context.

PR #3909, opened 2025-11-21
feat: add bindings for enableOnZones and enableOnComputeCluster

spacegospod:
## Description

Adding bindings for the following endpoints, structures included

https://developer.broadcom.com/xapis/vsphere-automation-api/latest/api/vcenter/namespace-management/supervisors__action=enable_on_zones/post/

https://developer.broadcom.com/xapis/vsphere-automation-api/latest/api/vcenter/namespace-management/supervisors/cluster__action=enable_on_compute_cluster/post/

Closes: #(issue-number)

## How Has This Been Tested?

Used these bindings in a local deployment of `vmware/terraform-provider-vsphere` to trigger both APIs

## Guidelines

### DIFF
--- a/vapi/namespace/namespace.go
+++ b/vapi/namespace/namespace.go
@@ -414,7 +414,7 @@ type CloudNativeFileVolume struct {
 func (c *Manager) EnableOnZones(ctx context.Context, spec *EnableOnZonesSpec) (string, error) {
 	var response string
 	url := c.Resource(path.Join(internal.SupervisorsPath)).WithParam("action", "enable_on_zones")
-	err := c.Do(ctx, url.Request(http.MethodPost, spec), response)
+	err := c.Do(ctx, url.Request(http.MethodPost, spec), &response)
 	return response, err
 }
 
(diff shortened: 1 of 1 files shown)


spacegospod:
> > LGTM
> 
> Thanks, so far I've manually tested `EnableOnComputeCluster` and I had to correct the bindings with `a829b5f7f56f8ce385fd36df0c03df73436a7467` I may have to make more corrections after testing `EnableOnZones`

Okay @tenthirtyam , I verified that `EnableOnZones` works too. I had to make 1 more fix to the bindings, they should be good-to-go now

spacegospod:
> @spacegospod - may I suggest updating comments for the existing `EnableClusterSpec` and `SizingHint` to align with the same `developer.broadcom.com` reference you included in this change? That might make this better - otherwise LGTM! 🚀

That's a good idea but there are 42 references to "vmware.com" in `namespace.go` alone, it doesn't end with `EnableClusterSpec` and `SizingHint`.
The URLs on the new developer portal are not a 1-to-1 mapping of the ones that used to be on vmware.com. What I mean to say is that each URL needs to be handled manually and I really want to avoid that

PR #3936, opened 2025-12-15
govc: add `device.sata.add`

### LINKED ISSUE #3780: Add govc command to add SATA controller
PR https://github.com/vmware/govmomi/pull/3458 added support for creating a SATA controller in the `vm.create` command. However, an independent command to add a SATA controller later was not added.

By analogy with `device.scsi.add` for adding additional SCSI controllers to an existing VM, there should be a `device.sata.add` command for adding SATA controllers to existing VMs.

Even better, imo, would be a more generic command like `device.controller.add` to allow adding any type of supported controller, specified with command options, rather than allowing 'device.\<type\>' commands to proliferate with every new technology to come out in the future.

I came across this missing command while trying to automate VM creation, using a hand-made VM as a model. The vSphere GUI adds a SATA controller to new VMs by default; `govc` does not. The result is that otherwise identically-configured VMs have a different number of PCI devices attached, which in turn changes the names of network interfaces in Linux VMs that generate NIC names based on their PCI address. This can then have effects on anything that was expecting a particular interface name, such as scripts run during guest customization or cloud-init configuration, or any config management solution that may run later on.

tenthirtyam:
### Description

- Added 'device.sata.add' command to add SATA controllers to virtual machines.
- Registered 'device.sata.add' in the 'govc' CLI entry point.
- Updated documentation and test suite to cover the new command.

Closes: #3780

### How Has This Been Tested?

`make govc-test` and manually.

```shell
govmomi on  govc/add-device-sata-add [!?] via 🐹 v1.25.5 
➜ vm="test-sata-$(date +%s)"
govc vm.create -on=false \
  -pool '/DC0/host/DC0_H0/Resources' \
  -net 'VM Network' \
  "$vm"

govmomi on  govc/add-device-sata-add [!?] via 🐹 v1.25.5 
➜ echo "== before =="
govc device.ls -vm "$vm" | egrep -i -- 'ahci|sata' || true

== before ==

govmomi on  govc/add-device-sata-add [!?] via 🐹 v1.25.5 
➜ id=$(govc device.sata.add -vm "$vm")
echo "added: $id"
govc device.info -vm "$vm" "$id"
added: ahci-204
Name:       ahci-204
  Type:     VirtualAHCIController
  Label:    ahci-204
  Summary:  ahci-204
  Key:      204
  Devices:  

govmomi on  govc/add-device-sata-add [!?] via 🐹 v1.25.5 
➜ echo "== after =="
govc device.ls -vm "$vm" | egrep -i -- 'ahci|sata' || true
== after ==
ahci-204      VirtualAHCIController      ahci-204
```

### DIFF
--- a/cli/device/sata/add.go
+++ b/cli/device/sata/add.go
@@ -1,0 +1,78 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package sata
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/vmware/govmomi/cli"
+	"github.com/vmware/govmomi/cli/flags"
+)
+
+type add struct {
+	*flags.VirtualMachineFlag
+}
+
+func init() {
+	cli.Register("device.sata.add", &add{})
+}
+
+func (cmd *add) Register(ctx context.Context, f *flag.FlagSet) {
+	cmd.VirtualMachineFlag, ctx = flags.NewVirtualMachineFlag(ctx)
+	cmd.VirtualMachineFlag.Register(ctx, f)
+}
+
+func (cmd *add) Description() string {
+	return `Add SATA controller to VM.
+
+Examples:
+  govc device.sata.add -vm $vm
+  govc device.info -vm $vm sata-*`
+}
+
+func (cmd *add) Process(ctx context.Context) error {
+	return cmd.VirtualMachineFlag.Process(ctx)
+}
+
+func (cmd *add) Run(ctx context.Context, f *flag.FlagSet) error {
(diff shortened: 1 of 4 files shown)


cli/device/sata/add.go
```diff
@@ -0,0 +1,81 @@
...
+Examples:
+  govc device.sata.add -vm $vm
+  govc device.info -vm $vm sata-*`
+}
+
+func (cmd *add) Process(ctx context.Context) error {
+	if err := cmd.VirtualMachineFlag.Process(ctx); err != nil {
+		return err
+	}
+	return nil
```
spacegospod:
```suggestion
	return cmd.VirtualMachineFlag.Process(ctx)
```

spacegospod (dismissed):
LGTM

dilyar85 (approved):

PR #3937, opened 2025-12-23
vcsim: Consider default devs from VM spec

akutz:
## Description

This patch updates the way vC Sim creates a new VM. Previously the VM's default devices were always added, regardless of whether or not the incoming ConfigSpec specified any of the default device info.

This patch instructs vC Sim to consider the incoming ConfigSpec and check if it contains one or more of the default devices, using them from the ConfigSpec if it does.

Closes: `NA`

## How Has This Been Tested?

The GitHub actions. Also, @fabriziopandini will test this with the Cluster API Provider for vSphere to see if it fixes his issue.

## Guidelines

### DIFF
--- a/simulator/esx/virtual_device.go
+++ b/simulator/esx/virtual_device.go
@@ -6,193 +6,156 @@ package esx
 
 import "github.com/vmware/govmomi/vim25/types"
 
+const (
+	VirtualMachineDefaultDevicePCIControllerKey = int32(100)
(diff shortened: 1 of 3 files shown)


bryanv (approved):

tenthirtyam (dismissed):

fabriziopandini:
Thank you @akutz, @bryanv and @tenthirtyam

PR #3942, opened 2026-01-02
Use *string for backingObjectId

prziborowski:
## Description

BackingObjectId is an optional value which means there is a distinction between unset and empty value.
The serialization process treats empty value as unset, which means it is changing the content between serialize/deserialize. Hostd differientiates between these 2 values and if it notices it changes will reject modifications when the VM is powered on, even though our intent was not to modify the content.

This change uses *string value to distinguish all the states. Clients that care about this value will have to do similar changes as I made to the tests.

Closes: #(issue-number)

## How Has This Been Tested?

I deployed this change in our environment where a vm.Reconfigure was consistently failing. With the change the operation succeeded.

## Guidelines


prziborowski:
Marking this as draft as I may need to do a similar thing for more properties.
I just tested with vSAN (which has a non-empty string of backingObjectId) and hostd is still blocking the reconfigure, meaning there are more properties that it thinks want to be changed. Unfortunately this environment isn't logged that, so I need a new environment, so likely won't have an update until tomorrow.

Issue #3952, opened 2026-02-03
AddLibraryUsage API returns incorrect output format


ridaz:
**Describe the bug**

The [AddLibraryUsage](https://github.com/vmware/govmomi/blob/main/vapi/library/library_usage.go#L66) API has response set to the below struct
```
var res struct {
		Value string `json:"value,omitempty"`
	} 

```
while the internal Content Library Usage API returns a string.
Hence, the Add Library Usage fails with the below error
```
err: json: cannot unmarshal string into Go value of type struct { Value string "json:\"value,omitempty\"" }
```

**To Reproduce**

1. Create a content library and attach it to the TKG service on the supervisor.
2. Corresponding ClusterContentLibrary is created.
3. It is attempted to add the Usage to library which then fails with the above error.

**Expected behavior**
The [AddLibraryUsage](https://github.com/vmware/govmomi/blob/main/vapi/library/library_usage.go#L66) should return a string instead of a struct object.

**Affected version**
The API was introduced as a part of [PR](https://github.com/vmware/govmomi/pull/3861)

**Screenshots/Debug Output**
Have attached it to internal JIRA.

**Additional context**

PR #3953, opened 2026-02-03
Fix AddLibraryUsage to return expected output format.

ridaz:
## Description

This change will fix the The [AddLibraryUsage](https://github.com/vmware/govmomi/blob/main/vapi/library/library_usage.go#L66) API to return "string" instead of a struct object, which is expected in the internal API.

The [AddLibraryUsage](https://github.com/vmware/govmomi/blob/main/vapi/library/library_usage.go#L66) API has response set to the below struct

var res struct {
		Value string `json:"value,omitempty"`
	} 

while the internal Content Library Usage API returns a string.
Hence, the Add Library Usage fails with the below error

```
err: json: cannot unmarshal string into Go value of type struct { Value string "json:\"value,omitempty\"" }
```

Closes: #3952 

## How Has This Been Tested?

The unit tests are utilising the simulated code.
Tested manually on a testbed by 
- Create a new Content Library and attach it to the supervisor.
- Content Library usage is set successfully.

## Guidelines

### DIFF
--- a/vapi/library/library_usage.go
+++ b/vapi/library/library_usage.go
@@ -68,8 +68,6 @@ func (c *Manager) AddLibraryUsage(ctx context.Context, libraryID string, addUsag
 		WithSubpath(libraryID).
 		WithSubpath(internal.LibraryUsages)
 
-	var res struct {
-		Value string `json:"value,omitempty"`
-	}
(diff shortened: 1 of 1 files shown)


ericvmw (approved):
LGTM, thanks for the fix!

PR #3972, opened 2026-03-02
api: Add SetRootCAsFromPEM method

prziborowski:
Add SetRootCAsFromPEM to allow passing in-memory PEM-encoded certificates instead of requiring them to be on disk. This is useful for read-only filesystems where CA contents are already available in memory.

## Description

Add SetRootCAsFromPEM to allow passing in-memory PEM-encoded
certificates instead of requiring them to be on disk. This is
useful for read-only filesystems where CA contents are already
available in memory.

## How Has This Been Tested?

Tested with a container doing:
```
	if len(config.CAContents) != 0 {
		if err = soapClient.SetRootCAsFromPEM([][]byte{config.CAContents}); err != nil {
			return fmt.Errorf("failed to set root CA %s: %w", config.CAFilePath, err)
		}
	}
```
and was able to connect to vCenter properly.

## Guidelines

### DIFF
--- a/vim25/soap/client_test.go
+++ b/vim25/soap/client_test.go
@@ -146,6 +146,74 @@ func setCAsOnClient(cas string) error {
 	return client.SetRootCAs(cas)
 }
 
+func TestValidRootCAsFromPEM(t *testing.T) {
+	pem, err := os.ReadFile("fixtures/valid-cert.pem")
+	if err != nil {
+		t.Fatalf("Failed to read valid cert: %v", err)
+	}
+
+	err = setCAsFromPEMOnClient([][]byte{pem})
+	if err != nil {
+		t.Fatalf("Err should not have occurred: %#v", err)
+	}
+}
+
+func TestInvalidRootCAsFromPEM(t *testing.T) {
+	cert := []byte(`-----BEGIN CERTIFICATE-----\n`)
+	err := setCAsFromPEMOnClient([][]byte{cert})
+
+	_, ok := err.(errInvalidCACertificate)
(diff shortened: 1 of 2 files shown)


vim25/soap/client_test.go
```diff
@@ -146,6 +146,74 @@ func setCAsOnClient(cas string) error {
...
+	_, ok := err.(errInvalidCACertificate)
+	if !ok {
+		t.Fatalf("Expected errInvalidCACertificate to occur with mixed certs")
+	}
+}
+
+func TestEmptyRootCAsFromPEM(t *testing.T) {
+	err := setCAsFromPEMOnClient([][]byte{})
+	if err != nil {
+		t.Fatalf("Empty slice should not cause error: %#v", err)
```
hickeng, line 201: I'm on the fence about this statement - while it's technically not an error it's likely indicative of a problem in the calling code; I cannot think of a good reason to be passing in an empty slice for a noop.

hickeng (approved):
Up to @prziborowski whether to change validation of empty slice or not.

vim25/soap/client_test.go
prziborowski: I can't think of a reason someone would want to do that.
But throwing an error in SetRootCAsFromPEM would mean I need to special case an empty slice scenario.
Or just remove this test case and then not consider it as a "supported" scenario.

hickeng: I'm happy to leave it as is - just wanted to flag it for your consideration.

PR #3976, opened 2026-03-06
Add vcsim support for container backed VMs able to run nested containers

hickeng:
Add RUN.nestedContainers ExtraConfig option to enable running Kubernetes
or other container workloads inside container-backed VMs. When enabled,
the container is created with kind-style flags:

- --cgroupns=private (instead of host) for proper cgroup v2 delegation
- --security-opt seccomp=unconfined for systemd/containerd syscalls
- --security-opt apparmor=unconfined to disable AppArmor restrictions
- --tmpfs /tmp,/run for systemd compatibility
- --volume /var for persistent containerd/kubelet data
- --volume /lib/modules:/lib/modules:ro for iptables/networking
- --device /dev/fuse for fuse-overlayfs snapshotter

In support of the above, add PropertyDiff and Checkpoint functions to generate PropertyChange
arrays by comparing two states of a managed object. This enables granular property change
notifications when modifying simulator objects.
This was necessitated by waiting for IP assignment, and Summary.Guest.IpAddress and similar
nested fields were not generating PropertyChange entries.

The typical usage pattern is:
```
  checkpoint := Checkpoint(vm)
  // ... make changes to vm ...
  changes := PropertyDiff(checkpoint, vm)
  ctx.Update(vm, changes)
```


hickeng:
Look at https://github.com/vmware/govmomi/blob/main/simulator/session_manager.go#L444 to address the race detector issue.

PR #4013, opened 2026-05-12
simulator: add PropertyDiff, Context helpers, and AutoUpdate pattern

hickeng:
## Description

Add PropertyDiff capability for granular property change tracking and .

* `PropertyDiff` – diffs old vs current managed-object state and returns
   []types.PropertyChange with correct Op values (Assign/Add/Remove). 
   Handles nested structs, embedded fields, and slices.
* `ctx.Checkpoint(obj)` – take a locked snapshot (deep copy) of obj's mo
   state; returns mo.Reference.
* `ctx.UpdateDiff(old, obj)`  – diff old vs current obj (under lock) and publish
   via ctx.Update.
* ctx.AutoUpdate(obj, mutateFn)  – acquire the lock once, snapshot, call
   mutateFn under the lock, diff, and publish. This helps avoid data races
   with concurrent updates via VMOMI API.

## How Has This Been Tested?

Unit and sim testing:
* simple/nested/slice field changes
* Add/Remove transitions
* embedded wrapper types (TestCheckpoint)
* the PropertyCollector integration (TestPropertyDiff_WithSimulator, TestContext_AutoUpdate)
* concurrent-access safety (TestPropertyDiff_ConcurrentAccess)

### DIFF
--- a/simulator/property_diff.go
+++ b/simulator/property_diff.go
@@ -1,0 +1,115 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package simulator
+
+import (
+	"reflect"
+
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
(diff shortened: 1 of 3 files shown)


dougm (approved):
Thanks @hickeng , this looks great!

PR #4014, opened 2026-05-12
simulator: improve podman compatibility for container-backed VMs

hickeng:
## Description

Add helpers and options that let vcsim container-backed VMs work correctly
under rootless podman-docker:

* commandError(cmd, err, stderr) – consistent error logging that includes the
   command's stderr so failures are easier to diagnose.
* RUN.network ExtraConfig – explicit container network name; needed for
   rootless podman where the default bridge does not provide cross-container
   IP connectivity.
* RUN.mountdmi ExtraConfig        – error hint when /dev/mem mount fails
  (rootless podman permission issue).
* syncNetworkConfigToVMGuestProperties – now calls ctx.Update for every
   modified property and retries IP resolution after container start to handle
   delayed IP assignment.
* watchContainer callback – updated to accept a *Context so callers can trigger
   property change notifications.
* test/helper.go – adds IsRootlessPodman() and ContainerNetworkFromSpec()
   helpers; Example_runContainer in feature_test.go uses them so the curl probe
   joins the same bridge network as the nginx VM.

## How Has This Been Tested?

Unit and sim testing

### DIFF
--- a/simulator/container.go
+++ b/simulator/container.go
@@ -35,9 +35,25 @@ const (
 )
 
 func init() {
-	if sh, err := exec.LookPath("bash"); err != nil {
+	if sh, err := exec.LookPath("bash"); err == nil && sh != "" {
 		shell = sh
 	}
+}
+
+// commandError logs and returns an error from a command execution failure, including stderr if available.
+func commandError(operation string, args []string, err error) error {
+	stderr := ""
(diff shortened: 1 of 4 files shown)


dougm (approved):
lgtm @hickeng . Should we look at changing the github workflow to use podman at some point? cc @tenthirtyam

PR #4016, opened 2026-05-18
api: add ConcurrentAccess to IsTransientError

### LINKED ISSUE #4017: api: add ConcurrentAccess to IsTransientError
## Problem

\`ConcurrentAccess\` is a vSphere fault thrown when an operation fails because another concurrent operation modified the datastructure. Clients that use \`IsTransientError\` as the canonical retry-check must currently add this fault type themselves, which defeats the purpose of the function (see #3824).

## Proposed Solution

Add \`types.ConcurrentAccess{}\` to the \`IsTransientError\` check in \`fault/meta_types.go\`.

Like \`TaskInProgress\`, the contention causing \`ConcurrentAccess\` is transient and the operation is expected to succeed on retry once the competing operation completes. vSphere HA itself treats \`ConcurrentAccess\` this way — when the fault is encountered during VM failover placement, HA sets an internal timeout and retries placement rather than failing immediately.

zhangheliu:
## Description

`ConcurrentAccess` is thrown when an operation fails because another concurrent operation modified the datastructure. Like `TaskInProgress` (already in the list), the contention is transient and the operation is expected to succeed on retry once the competing operation completes.

vSphere HA itself treats `ConcurrentAccess` this way — when the fault is encountered during VM failover placement, HA sets an internal timeout and retries placement rather than failing immediately.

This change ensures clients using `IsTransientError` as the canonical retry-check do not need to add this fault type themselves, which was the stated goal of the function (see #3824).

Closes: #4017

## How Has This Been Tested?

Added `TestIsTransientErrorConcurrentAccess` to `fault/meta_types_test.go` verifying both the positive case (`ConcurrentAccess` → `true`) and a negative case (`SystemError` → `false`).

## Guidelines

### DIFF
--- a/fault/meta_types.go
+++ b/fault/meta_types.go
@@ -45,5 +45,6 @@ func IsTransientError(err any) bool {
 		Is(err, &types.NetworkDisruptedAndConfigRolledBack{}) ||
 		Is(err, &types.VAppTaskInProgress{}) ||
 		Is(err, &types.FailToLockFaultToleranceVMs{}) ||
-		Is(err, &types.HostCommunication{})
+		Is(err, &types.HostCommunication{}) ||
+		Is(err, &types.ConcurrentAccess{})
 }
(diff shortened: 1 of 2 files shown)


dougm (approved):
Thanks @zhangheliu

PR #4021, opened 2026-05-21
govmomi: Add HostDistributedVirtualSwitchManager MO

chaitanyabandi:
## Description

This change adds the HostDistributedVirtualSwitchManager managed object into govmomi for the internal object implementation being added in [1] to use the manager to fetch the list of distributed virtual switches on the host.

This is needed for ESXi infravisord to fetch the list of distributed virtual switches on a host and get their port information for the CRXs that it manages.

1. https://github-vcf.devops.broadcom.net/vcf/govmomi-internal/pull/44

## How Has This Been Tested?

- Consumed this logic in infravisord and able to use the added FetchPortState and DistributedVirtualSwitches APIs.

### DIFF
--- a/vim25/mo/registry.go
+++ b/vim25/mo/registry.go
@@ -39,3 +39,16 @@ func (m TransitGateway) Reference() types.ManagedObjectReference {
 func init() {
 	t["TransitGateway"] = reflect.TypeOf((*TransitGateway)(nil)).Elem()
 }
+
+type HostDistributedVirtualSwitchManager struct {
+	Self                     types.ManagedObjectReference `json:"self"`
+	DistributedVirtualSwitch []string                     `json:"distributedVirtualSwitch"`
+}
+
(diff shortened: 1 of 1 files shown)


vim25/mo/mo.go
```diff
@@ -1854,3 +1854,16 @@ func (m VsanUpgradeSystem) Reference() types.ManagedObjectReference {
 func init() {
 	t["VsanUpgradeSystem"] = reflect.TypeOf((*VsanUpgradeSystem)(nil)).Elem()
 }
+
+type HostDistributedVirtualSwitchManager struct {
```
dougm: mo.go is generated and will be overwritten next time the generated code is updated. Moving the code here will avoid that: https://github.com/vmware/govmomi/blob/d12aa89c22af008b80ad44e3aeeec54822895d8f/vim25/mo/registry.go#L15

dougm (approved):

PR #4028, opened 2026-06-01
fix case-sensitive api call - enable supervisor cluster

dstoffel:
## Description

Structure VcenterNamespaceManagementNetworksEdgesNSXConfig is case sensitive.
Using edge_cluster_id instead of edge_cluster_ID is returning the error:

~~~
{"messages":[{"args":[],"default_message":"The nsxEdgeCluster field in NCPClusterNetworkInfo is required.","localized":"The nsxEdgeCluster field in NCPClusterNetworkInfo is required.","id":"vcenter.wcp.cluster.empty.nsx_edge_cluster"}]}
~~~

Changing the string to edge_cluster_ID

## How Has This Been Tested?

Tested by using terraform-provider-vsphere (vsphere_supervisor_v2)

### DIFF
--- a/vapi/namespace/namespace.go
+++ b/vapi/namespace/namespace.go
@@ -213,7 +213,7 @@ type HAProxy struct {
 // https://developer.broadcom.com/xapis/vsphere-automation-api/latest/data-structures/Vcenter%20NamespaceManagement%20Networks%20Edges%20NSXConfig
 // Since 8.0.0.1
 type EdgeNSX struct {
-	EdgeClusterID                *string    `json:"edge_cluster_id,omitempty"`
+	EdgeClusterID                *string    `json:"edge_cluster_ID,omitempty"`
 	DefaultIngressTLSCertificate *string    `json:"default_ingress_tls_certificate,omitempty"`
 	RoutingMode                  *string    `json:"routing_mode,omitempty"`
 	EgressIPRanges               *[]IPRange `json:"egress_ip_ranges,omitempty"`


vapi/namespace/namespace.go
```diff
@@ -213,7 +213,7 @@ type HAProxy struct {
 // https://developer.broadcom.com/xapis/vsphere-automation-api/latest/data-structures/Vcenter%20NamespaceManagement%20Networks%20Edges%20NSXConfig
 // Since 8.0.0.1
 type EdgeNSX struct {
-	EdgeClusterID                *string    `json:"edge_cluster_id,omitempty"`
+	EdgeClusterID                *string    `json:"edge_cluster_ID,omitempty"`
```
dougm, line 216: Will let @spacegospod confirm, but looking at an internal usage of this API and the [docs](https://developer.broadcom.com/xapis/vsphere-automation-api/latest/data-structures/Vcenter%20NamespaceManagement%20Networks%20Edges%20NSXConfig/), seems these also need upper-case:
- egress_IP_ranges 
- default_ingress_TLS_certificate

I don't know if this is a model or doc gen issue, but in the doc fields are listed in all lower-case "egress_ip_ranges", but the description includes the upper-case chars:
> Vcenter NamespaceManagement Networks Edges NSXConfig.egress_IP_ranges

spacegospod: I checked all 3 with the internal API specification - `edge_cluster_ID` and `egress_IP_ranges` need the upper-case fix but `default_ingress_TLS_certificate` is only mentioned in lower-case

dougm: Thanks @spacegospod . I'll merge this change, if @dstoffel wants to follow up with the `egress_IP_ranges` change that would be great.

dougm (approved):

PR #4032, opened 2026-06-03
vapi/library: tolerate spaces around '=' in ReadManifest

### LINKED ISSUE #3893: `library.ReadManifest` fails to parse the manifest file
**To Reproduce**
Call the [library.ReadManifest](https://github.com/vmware/govmomi/blob/02e26faac0813f47d493d400c948b810b7cd16e1/vapi/library/library_item_updatesession_file.go#L160) function to parse a manifest that looks like

```
SHA1(wazuh-4.13.1-disk-1.vmdk) = 7c1af154f974a4dd448dc527f1f4216491973039
SHA1(wazuh-4.13.1.ovf) = d861e788a9f65975b6c51686b721792985db88c7
```

The result is an empty list. The problem is that the existing code expects a file that looks like

```
SHA1(wazuh-4.13.1-disk-1.vmdk)= 7c1af154f974a4dd448dc527f1f4216491973039
SHA1(wazuh-4.13.1.ovf)= d861e788a9f65975b6c51686b721792985db88c7
```

I haven't found out if there is a specification for the `*.mf` file, but a parser shouldn't fail because of a `) = ` instead of `)= `.

**Expected behavior**
The function should be able to parse the file, even if there is a blank where it is not expected.

**Affected version**
v0.52.0

**Screenshots/Debug Output**
n/a

**Additional context**
n/a

mvanhorn:
## Summary

`library.ReadManifest` parsed each OVF manifest (`.mf`) line by splitting on the literal `")="`. Manifests written with whitespace around the equals sign (`") = "`) therefore produced a single-element slice, every line was skipped, and the function returned an empty checksum map.

This change anchors the parse on the closing paren of the algorithm group and tolerates optional whitespace around the `=`, so both `SHA1(name.vmdk)= <hash>` and `SHA1(name.vmdk) = <hash>` parse identically.

## Why this matters

Issue #3893 reports that real-world manifests (e.g. from `wazuh` OVAs) use the spaced `) = ` form, which has no strict `.mf` spec forbidding it. The old parser silently dropped every entry for such manifests, yielding an empty map with no error, so downstream import/verification lost all checksums. The new parse also handles file names containing `)` or `=` (anchored on the last `)` before the separator), which the prior `")="` split could mishandle.

## Testing

Rewrote `TestReadManifest` into a table-driven test covering:
- compact `)=` form (regression guard)
- spaced `) = ` form (the bug)
- mixed multi-line manifest (compact + spaced)
- `SHA256` algorithm name preserved
- file name containing parentheses
- file name containing `=`
- blank and malformed (no `)`) lines skipped without error

`go test ./vapi/library/...`, `go vet ./vapi/library/...`, and `gofmt -l` all pass.

Fixes #3893

AI was used for assistance.

### DIFF
--- a/vapi/library/library_item_updatesession_file.go
+++ b/vapi/library/library_item_updatesession_file.go
@@ -163,17 +163,26 @@ func ReadManifest(m io.Reader) (map[string]*Checksum, error) {
 
 	scanner := bufio.NewScanner(m)
 	for scanner.Scan() {
-		line := strings.SplitN(scanner.Text(), ")=", 2)
-		if len(line) != 2 {
+		// Manifest lines look like "ALG(file name)= digest". Anchor on the
+		// last ")" (the digest separator's closing paren) so that "(" or ")"
+		// or "=" inside the file name don't confuse the parse, then require an
+		// "=" after it. Whitespace around the "=" (e.g. ") = ") is tolerated.
+		line := scanner.Text()
+		rparen := strings.LastIndex(line, ")")
+		if rparen < 0 {
 			continue
 		}
-		name := strings.SplitN(line[0], "(", 2)
+		value := strings.TrimSpace(line[rparen+1:])
+		if !strings.HasPrefix(value, "=") {
+			continue
+		}
+		name := strings.SplitN(line[:rparen], "(", 2)
 		if len(name) != 2 {
 			continue
 		}
 		sum := &Checksum{
 			Algorithm: strings.TrimSpace(name[0]),
-			Checksum:  strings.TrimSpace(line[1]),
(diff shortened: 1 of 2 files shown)


dougm (approved):
Thanks @mvanhorn , lgtm

mvanhorn:
Glad this is in, thanks @dougm. Tolerating spaces around the equals sign in ReadManifest makes OVF manifest parsing more forgiving.

mvanhorn:
Appreciate the forgiving parser, @dougm. Tolerating spaces around the '=' in ReadManifest makes the OVF manifest handling a lot more robust.

PR #4046, opened 2026-06-16
vcsim: include VM extra config when cloning a VM

nickcaballero:
## Description

Include extra configuration from the VM cloning spec in the resulting VM.


dougm (approved):
Cool, thanks again @nickcaballero

PR #4051, opened 2026-06-21
Make VirtualDevice.NumaNode a pointer to allow node 0

### LINKED ISSUE #4049: [BUG] VirtualDevice.NumaNode: int32 + xml:omitempty makes NUMA node 0 unassignable and cleared state ambiguous
**Describe the bug**

`VirtualDevice.NumaNode` is declared as `int32` with `xml:"numaNode,omitempty"` in `vim25/types/types.go`. Go's XML encoder omits zero-valued `omitempty` fields, so writing `NumaNode = 0` into a `VirtualDeviceConfigSpec` is silently dropped from the SOAP payload — vSphere never receives the update. This causes two problems:

1. **NUMA node 0 cannot be assigned** — vSphere supports node 0 as a valid device NUMA assignment (nodes are zero-indexed), but it is unreachable from Go.
2. **Cleared vs. never-set state is ambiguous** — After clearing a NUMA assignment via `-1`, vSphere omits `numaNode` from `ConfigInfo`. govmomi reads the absent field back as `0` (int32 zero value), indistinguishable from a NIC that never had an assignment.

**To Reproduce**

1. Find a powered-off VM with at least one ethernet NIC and a vNUMA topology configured.
2. Use govmomi to issue `ReconfigVM_Task` with a `VirtualDeviceConfigSpec` (Edit) that sets `NumaNode = 1` on the NIC.
3. Read back `ConfigInfo.Hardware.Device` — `NumaNode` is `1` as expected.
4. Now issue another `ReconfigVM_Task` with `NumaNode = 0`.
5. Read back `ConfigInfo.Hardware.Device` — `NumaNode` is still `1`. The write of `0` was a no-op.

**Expected behavior**

Writing `NumaNode = 0` should assign the device to NUMA node 0, consistent with vSphere's wire protocol where 0 is a valid NUMA node. Reading back after a clear (`-1`) should return a distinguishable "unset" value (`nil` / absent) rather than `0`.

**Affected version**

Reproduced with `github.com/vmware/govmomi v0.48.1` against vCenter 8.x. The `int32` declaration has been present across all recent versions.

```go
// vim25/types/types.go
NumaNode int32 `xml:"numaNode,omitempty"`
```

**Screenshots/Debug Output**

Tested with a standalone Go program using govmomi:

| `ConfigSpec NumaNode` written | SOAP field present? | vSphere result | Go reads back |
|---|---|---|---|
| `1` | yes | stores 1 | `1` |
| `0` | **no** (omitted by omitempty) | no-op | `1` (unchanged) |
| `-1` | yes | clears affinity; field absent in ConfigInfo | `0` |

**Additional context**

Proposed fix — change to a pointer so zero is representable and absent becomes `nil`:

```go
// Before
NumaNode int32 `xml:"numaNode,omitempty"`

// After
NumaNode *int32 `xml:"numaNode,omitempty"`
```

This mirrors `UnitNumber *int32` in the same struct, which already uses a pointer because 0 is a valid unit number.

Current workaround: write `-1` to clear (non-zero → not omitted → vSphere interprets negative as "no affinity"). Assigning NUMA node 0 from Go is not possible until fixed.

Ref: govmomi `types.go` comment on `NumaNode`: *"A negative number means there is no affinity for the device. A positive number is a vNUMA node. An unset value of numaNode is status-quo during Reconfigure time."*

mvanhorn:
## Description

`VirtualDevice.NumaNode` was generated as `int32` with an `xml:"numaNode,omitempty"` tag. Go's XML encoder drops zero-valued `omitempty` fields, so writing `NumaNode = 0` into a `VirtualDeviceConfigSpec` was silently omitted from the SOAP payload and never reached vSphere. This made NUMA node 0 (a valid zero-indexed assignment) unreachable from Go, and made a cleared assignment read back as `0`, indistinguishable from a device that was never assigned.

This change makes `NumaNode` a `*int32` with an `xml:"numaNode"` tag (no `omitempty`) and `json:"numaNode,omitempty"`, mirroring the existing `CoresPerNumaNode *int32` pattern. A non-nil `*int32(0)` now marshals `<numaNode>0</numaNode>` (node 0 is assignable), and a nil pointer is omitted entirely (never-set / cleared state is distinguishable on read-back).

The field comment already documents this semantic: "An unset value of numaNode is status-quo during Reconfigure time."

Changes:
- `gen/vim_wsdl.rb`: add `"NumaNode"` to the optional-int pointer override list in `pointer_type?` (the same list that already contains `CoresPerNumaNode`, `IpPoolId`, `Reservation`, etc.), so the pointer shape survives the next WSDL regeneration.
- `vim25/types/types.go`: the regenerated `VirtualDevice.NumaNode` field (`*int32`, XML tag without `omitempty`).
- `cli/gpu/vm/info.go`: the only internal consumer of the field; updated to carry the pointer through `gpu.vm.info` output so an unassigned device prints `unset` instead of an ambiguous `0`.

The generator runs via `make generate-types` inside a container against VMware's proprietary WSDL/SDK bundle and regenerates every package while bumping the VIM version, which is not reproducible in a standard contributor environment. The single `types.go` field was therefore edited by hand to exactly match the shape the generator emits for the entries already in the `pointer_type?` list (e.g. `CoresPerNumaNode`); the generator entry is the source-of-truth change and was added so a full regeneration produces the same result.

Maintainer @dougm confirmed the fix on 2026-06-19, pointing at the generator's pointer-field override list in `gen/vim_wsdl.rb` as the place to add `NumaNode`.

Closes: #4049

## How Has This Been Tested?

- `go build ./vim25/types/... ./cli/gpu/...` and `go vet ./vim25/... ./cli/gpu/...` pass cleanly.
- `go test ./vim25/types/...` passes (XML and JSON round-trip fixtures).
- Verified marshalling behavior: a device with `NumaNode = types.NewInt32(0)` now emits `<numaNode>0</numaNode>`; a device with `NumaNode = nil` omits the element entirely.

## Guidelines

Fixes #4049

### DIFF
--- a/cli/gpu/vm/info.go
+++ b/cli/gpu/vm/info.go
@@ -58,7 +58,7 @@ type gpuInfo struct {
 	Index    int    `json:"index"`
 	Label    string `json:"label"`
 	Summary  string `json:"summary"`
-	NumaNode int32  `json:"numaNode"`
+	NumaNode *int32 `json:"numaNode"`
 	Key      int32  `json:"key"`
 }
 
@@ -72,7 +72,11 @@ func (r *infoResult) Write(w io.Writer) error {
 		fmt.Fprintf(tw, "GPU %d:\n", gpu.Index)
 		fmt.Fprintf(tw, "  Label: %s\n", gpu.Label)
 		fmt.Fprintf(tw, "  Summary: %s\n", gpu.Summary)
-		fmt.Fprintf(tw, "  Numa Node: %d\n", gpu.NumaNode)
+		if gpu.NumaNode != nil {
+			fmt.Fprintf(tw, "  Numa Node: %d\n", *gpu.NumaNode)
+		} else {
+			fmt.Fprintf(tw, "  Numa Node: unset\n")
+		}
 		fmt.Fprintf(tw, "  Key: %d\n", gpu.Key)
 	}
 	return tw.Flush()
--- a/gen/vim_wsdl.rb
+++ b/gen/vim_wsdl.rb
@@ -252,7 +252,7 @@ class Simple
 
   def pointer_type?
     ["UnitNumber"].include?(var_name) or
-      optional? && ["CoresPerNumaNode", "IpPoolId", "OwnerId", "GroupId", "MaxWaitSeconds", "Reservation", "Limit", "OverheadLimit", "ResourceReductionToToleratePercent"].include?(var_name)
+      optional? && ["CoresPerNumaNode", "IpPoolId", "OwnerId", "GroupId", "MaxWaitSeconds", "NumaNode", "Reservation", "Limit", "OverheadLimit", "ResourceReductionToToleratePercent"].include?(var_name)
   end
 
   def var_type
--- a/vim25/types/types.go
+++ b/vim25/types/types.go
@@ -85943,7 +85943,7 @@ type VirtualDevice struct {
 	// An unset value of numaNode is status-quo during Reconfigure time.
 	// If numaNode is unset during ConfigInfo, then it means there is no
 	// affinity for the device.
-	NumaNode int32 `xml:"numaNode,omitempty" json:"numaNode,omitempty" vim:"8.0.0.1"`
+	NumaNode *int32 `xml:"numaNode" json:"numaNode,omitempty" vim:"8.0.0.1"`
 	// Information about device group device is part of.
 	//
 	// Devices in the device group cannot be added/removed individually,


dougm (approved):
Thanks @mvanhorn !

hpannem:
Thanks for fixing this. While this is the correct fix, we may need to release note the API change if there are clients already using it. @dougm please review.

dougm:
Yeah, should have documented with [BREAKING:](https://github.com/vmware/govmomi/blob/main/CONTRIBUTING.md#example-5---document-breaking-api-changes) in the git commit.

mvanhorn:
Thank you @dougm. Making VirtualDevice.NumaNode a pointer so node 0 survives the omitempty XML encoder is a subtle but important fix.

PR #4052, opened 2026-06-23
vcsim: power on and customize VM when cloning

nickcaballero:
## Description

Power on and customize VMs cloned in vcsim.

## How Has This Been Tested?

- Create a VM in vcsim
- Clone the VM with `powerOn=true` and some customizations
  - If the customization fails, the error should be reported in the result of the clone task


dougm (approved):
Nice, thanks @nickcaballero

PR #4060, opened 2026-06-29
Adds govc commads for checking interoperability of components

phiremath:
## Description

Use vcenter paths /lcm/rest/cis/tasks/ and /lcm/rest/vcenter/lcm/interop to submit interop check request and get inerop report. These API endpoints support basic auth only, hence we only use net/http client. Added mirror implementations of interop task response content objects. These commands are marked as unreleased.
Use GOVC_SHOW_UNRELEASED=true to use them.

The two commands are as described below:

% GOVC_SHOW_UNRELEASED=true govc lcm.interop.check -h Usage: govc lcm.interop.check [OPTIONS]

Submit an LCM interoperability validation task.

Prints the task ID on success. Use lcm.interop.result to retrieve the report, or pass -wait to block until completion and print it immediately.

Examples:
  govc lcm.interop.check -version 8.0.3
  govc lcm.interop.check -product com.vmware.vcenter.supervisor -version 8.0.3 -identifier my-instance
  govc lcm.interop.check -version 8.0.3 -wait

govc lcm.interop.report -h
Usage: govc lcm.interop.report [OPTIONS] TASK_ID

Fetch and display the interoperability report for a task.

The task must have been created by lcm.interop.check. If the task is not yet complete, the current status is printed and the command exits with an error.

Examples:
  govc lcm.interop.report <task-id>
  govc lcm.interop.report -json <task-id

Closes: #(issue-number)

## How Has This Been Tested?

Unit and integration testing.

### DIFF
--- a/vapi/lcm/interop.go
+++ b/vapi/lcm/interop.go
@@ -1,0 +1,284 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package lcm
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/vmware/govmomi/vapi/cis/tasks"
+)
+
+const (
+	interopTasksPath = "/lcm/rest/cis/tasks/"
+	interopPath      = "/lcm/rest/vcenter/lcm/interop"
+)
+
+// InteropComponent specifies the component to validate interoperability for.
+type InteropComponent struct {
+	ProductName string
+	Version     string
+	Identifier  string // optional instance identifier
+}
+
+// InteropResult is the decoded payload returned by a completed interop task.
+type InteropResult struct {
+	Report    InteropReport         `json:"report"`
+	Issues    *InteropNotifications `json:"issues,omitempty"`
+	CsvReport *string               `json:"csv_report,omitempty"`
+}
+
+// InteropNotifications groups info, warning, and error notifications from the API.
+type InteropNotifications struct {
+	Info     []InteropNotification `json:"info,omitempty"`
+	Warnings []InteropNotification `json:"warnings,omitempty"`
+	Errors   []InteropNotification `json:"errors,omitempty"`
+}
+
+// InteropNotification is a single notification entry returned alongside the report.
+type InteropNotification struct {
+	Id         string                     `json:"id"`
+	Time       *time.Time                 `json:"time,omitempty"`
+	Message    InteropLocalizableMessage  `json:"message"`
+	Resolution *InteropLocalizableMessage `json:"resolution,omitempty"`
+}
+
+// InteropLocalizableMessage carries a human-readable message from the API.
+type InteropLocalizableMessage struct {
+	Id             string  `json:"id"`
+	DefaultMessage string  `json:"default_message"`
+	Localized      *string `json:"localized,omitempty"`
+}
+
+// InteropReport is the top-level report container.
+type InteropReport struct {
+	DateCreated time.Time          `json:"date_created"`
+	Rows        []InteropReportRow `json:"rows"`
+}
+
+// InteropReportRow is one row of the interoperability report.
+type InteropReportRow struct {
+	TargetComponent      InteropTargetComponent       `json:"target_component"`
+	Summary              InteropSummary               `json:"summary"`
+	AssociatedComponents []InteropAssociatedComponent `json:"associated_components"`
+}
+
+// InteropTargetComponent describes the component being validated.
+type InteropTargetComponent struct {
+	Identifier *string            `json:"identifier,omitempty"`
+	Product    InteropProductInfo `json:"product"`
+	Version    string             `json:"version"`
+	Name       *string            `json:"name,omitempty"`
+}
+
+// InteropProductInfo holds product metadata.
+type InteropProductInfo struct {
+	DisplayName string `json:"display_name"`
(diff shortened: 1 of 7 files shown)


vapi/lcm/interop.go
```diff
@@ -0,0 +1,284 @@
...
+	rawURL := m.lcmURL(interopTasksPath+taskID, nil)
+	var info tasks.Info
+	if err := m.do(ctx, http.MethodGet, rawURL, nil, &info); err != nil {
+		return nil, err
+	}
+	return &info, nil
+}
+
+// WaitForCompletion polls GetTask until the task reaches a terminal state.
+func (m *Manager) WaitForCompletion(ctx context.Context, taskID string) (*tasks.Info, error) {
```
sidharthsurana, line 242: you should be able to use `tasks.NewManager(m.Client).WaitForCompletion(ctx, taskId)`

vapi/lcm/interop.go
```diff
@@ -0,0 +1,284 @@
...
+	rawURL := m.lcmURL(interopPath, url.Values{"vmw-task": {"true"}})
+	var taskID string
+	if err := m.do(ctx, http.MethodPost, rawURL, body, &taskID); err != nil {
+		return "", err
+	}
+	return taskID, nil
+}
+
+// GetTask fetches the current state of an interop task.
+func (m *Manager) GetTask(ctx context.Context, taskID string) (*tasks.Info, error) {
```
sidharthsurana, line 232: you should be able to use `tasks.NewManager(m.Client).Get(ctx, taskId)`

phiremath: For this use-case, the path does not start with "/api", hence it gets changed which can lead to 404. https://github.com/vmware/govmomi/blob/main/vapi/rest/client.go#L109.

vapi/lcm/interop.go
```diff
@@ -0,0 +1,284 @@
...
+
+// WithPollingInterval overrides the interval between WaitForCompletion polls.
+func (m *Manager) WithPollingInterval(d time.Duration) *Manager {
+	m.pollingInterval = d
+	return m
+}
+
+// NewInsecureManager is a convenience constructor that creates its own
+// http.Client with TLS verification disabled.
+func NewInsecureManager(baseURL, username, password string) *Manager {
```
fdawg4l: Why not use url.URL for `baseURL` vs a string.  Types are always preferred over `string`s.

vapi/lcm/interop.go
```diff
@@ -0,0 +1,284 @@
...
+	return NewManager(baseURL, username, password, &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
+		},
```
fdawg4l, line 171: This is very problematic.  We shouldn't skip tls verification.

Also, why can't we use [this](https://github.com/vmware/govmomi/blob/main/vapi/rest/client.go#L28) instead?

vapi/lcm/interop.go
```diff
@@ -0,0 +1,284 @@
...
+
+func (m *Manager) lcmURL(path string, params url.Values) string {
+	u, _ := url.Parse(m.baseURL + path)
+	if len(params) > 0 {
+		u.RawQuery = params.Encode()
+	}
+	return u.String()
+}
+
+func (m *Manager) do(ctx context.Context, method, rawURL string, reqBody, resBody any) error {
```
fdawg4l, line 182: I'm not sure we need to do any of this if we use the rest.Client.  

I'm actually surprised this works.  The vapi json includes DynamicData typed fileds which the json unmarshaller does not like.

vapi/lcm/interop.go
```diff
@@ -0,0 +1,284 @@
...
+	req.Header.Set("Accept", "application/json")
+	req.SetBasicAuth(m.username, m.password)
+
+	resp, err := m.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
```
fdawg4l, line 203: The body is an io.Reader.  You could decode the reader using the [Decoder](https://pkg.go.dev/encoding/json#NewDecoder) and avoid copying the body to memory only to unmarshall it.

vapi/lcm/interop.go
phiremath: The interop APIs support basic-auth and do not need sessionId/cookie of any kind. Hence rolled a new Manager/Client.

vapi/lcm/interop.go
phiremath: We do need raw body for constructing error (non 2xx responses), hence went with ReadAll + Unmarshall.
We do not expect the body to be very large for our use-case (checking interop of only one component).

vapi/lcm/interop.go
phiremath: Interop APIs starts with /lcm/* and rest.Client is meant for /rest/* or /api/* (ref: https://github.com/vmware/govmomi/blob/main/vapi/rest/client.go#L109, https://github.com/phiremath/govmomi/blob/add-interop-api-client/vapi/lcm/interop.go#L22)
Interop json responses do not use discriminator fields, just plain JSON with snake case.
Interop also has JSON-RPC endpoints which we are not using here.

vapi/lcm/interop.go
phiremath: govc support insecure connections. Its has a flag to skip verification (GOVC_INSECURE), this was meant as a convenience constructor for that use case (ref: https://github.com/phiremath/govmomi/blob/add-interop-api-client/cli/lcm/interop/common.go#L31) and currently only used for testing.

vapi/lcm/interop.go
phiremath: Sure, makes sense to url.URL.

PR #4064, opened 2026-07-01
api(cns): add ActiveClusters to CnsVolumePolicyReconfigSpec

Joshzxj:
## Description

Adds an optional `ActiveClusters` field to `CnsVolumePolicyReconfigSpec`.

`CnsReconfigVolumePolicy` can auto-relocate a volume to a compatible datastore
when the target storage profile differs from the volume's current profile, but
only when the caller supplies the set of active `ClusterComputeResource`s to
relocate within — the same input already carried by
`CnsVolumeCreateSpec.ActiveClusters`.

This change adds the field to `CnsVolumePolicyReconfigSpec`, mirroring the
create spec's type and XML serialization (`omitempty` + `typeattr`), so a policy
reconfigure can request a relocate. Without it, `ReconfigPolicy` can only perform
an in-place profile change. The field is optional and backward compatible:
existing callers that omit it serialize the same request as before.

## How Has This Been Tested?

`go test -race ./cns/types/...` and `make go-test` pass.

A new unit test (`TestCnsVolumePolicyReconfigSpecActiveClusters`) verifies that
`ActiveClusters` serializes into the `ReconfigPolicy` request XML when set, and
is omitted (via `omitempty`) when nil — i.e. an in-place reconfigure.

## Guidelines

### DIFF
--- a/cns/types/activeclusters_test.go
+++ b/cns/types/activeclusters_test.go
@@ -1,0 +1,52 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/govmomi/vim25/xml"
+)
+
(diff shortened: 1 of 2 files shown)


dougm (approved):
lgtm

PR #4065, opened 2026-07-03
fix: remove ovf test files with non-ascii names

dougm:
Causes go module checksum mismatch on certain systems.


HakanSunay (commented):
Lgtm, cant approve anymore

PR #4075, opened 2026-07-14
Use AutoUpdate for container-backed VM

hickeng:
## Description

This shifts container backed sim VMs to using AutoUpdate to populate PropertyCollector updates for network info.

It also:
* corrects the reflection based diff to equate nil and empty slices
* allows use of an existing result from container inspect to be passed into the network info diff, both for efficiency (only one inspect call) and result consistency.
* refactors setup of containers to support curl probes into a helper
* corrects docstrings where they're imprecise or incorrect.
* adds tests for the above

## How Has This Been Tested?

vcsim testing

### DIFF
--- a/simulator/container.go
+++ b/simulator/container.go
@@ -704,7 +704,7 @@ func (c *container) updated() {
 // returns:
 //
 //	err - uninitializedContainer error - if c.id is empty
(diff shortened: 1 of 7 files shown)


dougm (approved):
Cool, lgtm

tenthirtyam (approved):

PR #4076, opened 2026-07-14
Add CRUD ops for LookupService entries

hickeng:
## Description

Adds the Create, Get, Delete methods for LookupService entries

## How Has This Been Tested?

Tested via use in a topic branch of a different repo that makes use of lookup service.

### DIFF
--- a/lookup/simulator/simulator.go
+++ b/lookup/simulator/simulator.go
@@ -73,6 +73,7 @@ type ServiceRegistration struct {
 	Info []types.LookupServiceRegistrationInfo
 
 	register sync.Once
+	mu       sync.RWMutex
 }
 
 func (s *ServiceRegistration) GetSiteId(_ *types.GetSiteId) soap.HasFault {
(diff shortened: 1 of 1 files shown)


lookup/simulator/simulator.go
```diff
@@ -73,6 +73,7 @@ type ServiceRegistration struct {
 	Info []types.LookupServiceRegistrationInfo
 
 	register sync.Once
+	mu       sync.RWMutex
```
dougm: Is there an issue with the default locking mechanism? If so, please describe and opt-out of auto-locking by either embedding `sync.RWMutex` or:

https://github.com/vmware/govmomi/blob/2520160009b1a61094dbd1ef0c02357ca40f7ecf/simulator/registry.go#L713

dougm (approved):
lgtm

lookup/simulator/simulator.go
hickeng: @dougm resolved - I also ran concurrency tests to make sure it was getting properly protected by the auto lock, but figured they didn't really add anything useful for the long term so didn't include them in the PR so as not to further expand the test suite time.

PR #4077, opened 2026-07-14
Add dependency injection for container-backing

hickeng:
## Description

This is added so that we can attach container backing to sim VMs that are created by product code, without requiring the product code to understand the simulation and the special extraConfig keys.
The intended usage pattern is that test code registers `ContainerImageEntry`'s that will be applied during the sim-testing of unmodified product code that's creating VMs.

This is an extremely basic implementation initially, intended to support deploying a VM via ovftool.
I'm expecting a followup to support  vmoperator creating VMs originating from ContentLibrary which will likely need to expand this to include Content Library IDs and/or source URLs.

## How Has This Been Tested?

Tested using a WIP topic branch in another repo.


dougm (approved):
Nice!

tenthirtyam (approved):

PR #4083, opened 2026-08-14
refactor: modernize for loops to use range

tenthirtyam:
## Description

Tech Debt: Replace classic C-style loops with Go's modern range-over-integer syntax introduced in Go 1.22.


dougm (approved):
Nice!

PR #4085, opened 2026-08-17
fix: include AudienceRestriction in SAML assertion C14N

### LINKED ISSUE #4082: sts: AudienceRestriction is dropped from assertion C14N, breaking ActAs token requests
**Describe the bug**
`saml2:Conditions` may contain a `saml2:AudienceRestriction` element, which scopes an
assertion to specific relying parties. `sts/internal` does not model this element, so
`Assertion.C14N()` silently drops it.

This breaks signing. `RequestSecurityToken.String()` embeds the **original** assertion in
the request body, while the signature digest is computed over
`RequestSecurityToken.C14N()`. Any element missing from the canonical form produces a
digest the STS cannot reproduce from the body it received, so the request is rejected with
`Signature is invalid`.

In practice this affects `Issue` with `TokenRequest.ActAs` (delegated / Holder-of-Key
tokens) and `Renew`, whenever the presented token carries an audience.

**To Reproduce**
Deterministic, no vCenter required — save as `sts/internal/repro_test.go`:

```go
package internal

import (
      "testing"

      "github.com/vmware/govmomi/vim25/xml"
)

func TestAudienceRestrictionDropped(t *testing.T) {
      in := `<saml2:Conditions xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion" ` +
              `NotBefore="2026-01-01T00:00:00.000Z" NotOnOrAfter="2026-01-01T00:05:00.000Z">` +
              `<saml2:ProxyRestriction Count="10"></saml2:ProxyRestriction>` +
              `<saml2:AudienceRestriction><saml2:Audience>https://example.com/saml</saml2:Audience></saml2:AudienceRestriction>`
              `</saml2:Conditions>`

      var c Conditions
      if err := xml.Unmarshal([]byte(in), &c); err != nil {
              t.Fatal(err)
      }
      t.Logf(" in: %s", in)
      t.Logf("out: %s", c.C14N())
}
```

1. `go test ./sts/internal/ -run TestAudienceRestrictionDropped -v`
1. The `AudienceRestriction` present in the input is absent from the C14N output.

Against a live STS:

1. Obtain a bearer SAML token whose `Conditions` ction` — for
   example a token requested with `wst:Participants`.
1. Call `sts.Client.Issue` with `TokenRequest{ActAn>,
   Certificate: <solution user cert>}`.
1. The STS rejects the request with `Signature is

Note: govmomi cannot currently request an audience
`internal.RequestSecurityToken` has no `Participants` field — so the audience-bearing
token has to be acquired through another client.

**Expected behavior**
`Assertion.C14N()` preserves `AudienceRestriction`, so the signed digest matches what the
STS computes from the received body, and `ActAs` / `Renew` requests succeed.

**Affected version**
main and older releases

**Screenshots/Debug Output**
NA

**Additional context**
NA

DimpleRajaVamsi:
## Description

`saml2:Conditions` may contain a `saml2:AudienceRestriction` element, which
scopes an assertion to specific relying parties. `sts/internal` did not model
it, so `Assertion.C14N()` dropped it.

That matters because the two sides of the signature are built differently:
`RequestSecurityToken.String()` embeds the *original* assertion in the request
body, while the digest is computed over `RequestSecurityToken.C14N()`. Any
element missing from the canonical form therefore produces a digest the STS
cannot reproduce from the body it received, and the request is rejected with
"Signature is invalid".

In practice this breaks `Issue` with `TokenRequest.ActAs` (delegated /
Holder-of-Key tokens) and `Renew`, whenever the presented token carries an
audience — for example a token that was requested with `wst:Participants`.

This adds `AudienceRestriction` and `Audience` types plus a
`Conditions.AudienceRestriction` field, and canonicalizes them in
`Conditions.C14N()`.

Closes: #4082 

## How Has This Been Tested?

- Unit tested
- Tested against VC SAML that has `AudienceRestriction`
- Tested against VC SAML that doesn't has `AudienceRestriction`

## Guidelines

### DIFF
--- a/sts/internal/types.go
+++ b/sts/internal/types.go
@@ -368,12 +368,15 @@ func init() {
 	types.Add("rsa:RenewRestrictionType", reflect.TypeOf((*RenewRestriction)(nil)).Elem())
 }
 
+// Conditions field order must match that of an issued token, as C14N preserves document order.
+// The STS appends AudienceRestriction after the Condition elements, hence its position here.
 type Conditions struct {
-	XMLName          xml.Name
-	NotBefore        string            `xml:",attr"`
-	NotOnOrAfter     string            `xml:",attr"`
-	ProxyRestriction *ProxyRestriction `xml:",omitempty"`
-	Condition        []BaseCondition   `xml:",omitempty"`
+	XMLName             xml.Name
+	NotBefore           string                `xml:",attr"`
+	NotOnOrAfter        string                `xml:",attr"`
+	ProxyRestriction    *ProxyRestriction     `xml:",omitempty"`
+	Condition           []BaseCondition       `xml:",omitempty"`
+	AudienceRestriction []AudienceRestriction `xml:",omitempty"`
 }
 
 func (c *Conditions) C14N() string {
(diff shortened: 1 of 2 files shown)


dougm (approved):
Thanks @DimpleRajaVamsi , lgtm

PR #4092, opened 2026-08-18
Vcsim nested containers and guestrpc

hickeng:
## Summary

Extends vcsim's container-backed VMs (`RUN.container`) so guest-side tooling that expects a real VMware guest environment — cloud-init's `DataSourceVMware`, `vmware-rpctool`, custom init processes reading/writing `guestinfo` — works **unmodified** against a container instead of needing to be VM-agnostic or mocked out.

Three pieces:

1. **`RUN.vmci`**: a per-VM GuestRPC server (Unix socket, bind-mounted into the container) that implements the same RPCI surface as real `vmtoolsd`: `info-get`/`info-set` against the VM's `ExtraConfig`, reflected as `PropertyChange`s. This makes `guestinfo` genuinely **bidirectional** — today, container-backed VMs can only push guestinfo in at creation time; this lets the guest write back status (readiness, config, whatever a workload needs) and have it show up as a normal `ExtraConfig`/property change on the VM object.
2. **`govmomi/toolbox`**: two new `Channel` implementations (Unix socket, AF_VSOCK with DataMap framing — the real `vmtoolsd` wire format) plus a small CLI mode so the `toolbox` binary itself can act as a drop-in `vmware-rpctool` or answer `vmtoolsd --cmd` one-shot invocations. This is generic GuestRPC tooling with no vcsim dependency; the simulator wires it in as a convenience (auto-injected at `/usr/bin/vmware-rpctool` when `RUN.vmci=true`).
3. **Nested-container improvements**: refinements to `RUN.nestedContainers` (the kind-style flag set for running Kubernetes/other container workloads inside the container-backed VM) so it composes cleanly with the GuestRPC socket bind-mount — in particular, the socket *directory* is mounted rather than the socket file, since a file bind-mount requires the target to already exist inside a fresh `--tmpfs /run`, which it doesn't.

This is aimed at anyone using vcsim to simulate a fleet of guests that need to look enough like real VMs for their own tooling to work without a real hypervisor — Kubernetes-node simulation, cloud-init-driven provisioning tests, or anything else that currently has to special-case "running under vcsim."

## What's included

- `simulator/guestrpc_server.go` — the per-VM GuestRPC server, its wire-format auto-detection (a lightweight LE-framed protocol vs. the DataMap framing real `vmtoolsd` uses), and concurrency handling.
- `simulator/vmci_artifacts_linux.go` (+ `_other.go` stub) — builds and caches the `toolbox` binary for auto-injection.
- `toolbox/unix_channel.go`, `toolbox/vsock_channel_linux.go` (+`_other.go`), `toolbox/datamap.go`, `toolbox/noop_channel.go` — the new channels and framing.
- `toolbox/toolbox/main.go` — `vmware-rpctool` and `vmtoolsd --cmd` compatible CLI modes, transport selection.
- Tests: in-process server tests (`guestrpc_test.go`), container-integration tests gated on Docker/Podman (`container_guestrpc_test.go`, `vsock_container_test.go`), and two test fixtures under `simulator/testdata/` (a systemd-init image, a small vsock test binary).

## Known limitations / follow-up

- `info-set` currently accepts any `ExtraConfig` key, not just `guestinfo.*` — real VMX restricts guest writes to that namespace. Worth tightening before this is relied on with untrusted guest images.
- The GuestRPC artifact build (`vmci_artifacts_linux.go`) writes into `simulator/bin/` relative to the source file, which won't resolve when `govmomi` is consumed as a module dependency from the read-only module cache; it also hardcodes `GOARCH=amd64`. A test-fixture build (Makefile target, `go:embed`, or a `TestMain` step) instead of a power-on-time `go build` would sidestep both.
- An AF_VSOCK seccomp-intercept transport was prototyped separately and is not part of this change; the Unix-socket transport (`RUN.vmci=true`) covers the common cases, but a caller that specifically needs a real `socket(AF_VSOCK)` call to succeed inside the container would still need it.

I've posted a more detailed review (findings + what already works) as a review comment on this PR, since some of it is more useful there than in the PR body.

## Testing done

- `go build ./...`, `go vet ./simulator/... ./toolbox/...` — clean.
- `go test ./toolbox/...` — pass.
- `go test ./simulator/...` (full package, unfiltered) — pass.
- Docker-gated container tests pass where a runtime is available (`TestVMCI_ToolboxBinary_GuestInfoRoundTrip`) and skip cleanly otherwise (the two tests needing a locally-built test image skip with a message rather than failing).

### DIFF
--- a/.github/workflows/govmomi-go-tests.yaml
+++ b/.github/workflows/govmomi-go-tests.yaml
@@ -42,6 +42,14 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-${{ matrix.go-version }}-
 
+      - name: Build VMCI systemd-init test image
+        # Non-fatal by design (see the Makefile target's comment): a runner without
+        # docker, or a build failure, must not fail the job. continue-on-error is
+        # belt-and-suspenders on top of that; Run Tests below triggers the same
+        # target again via go-test's prerequisite, so this step is for CI visibility.
+        continue-on-error: true
+        run: make vmci-systemd-init-image
+
       - name: Run Tests
         env:
           TEST_COUNT: 1
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,11 @@ dist/
 # Ignore editor temp files
 *~
 .vscode/
+.cursor/
+
+# Test binaries
+*.test
+
+# vmci simulation artifacts (built on-demand by simulator tests via go build)
+/vmci-guest
+/simulator/bin/
--- a/Makefile
+++ b/Makefile
@@ -143,8 +143,22 @@ ifeq (-timeout,$(findstring -timeout,$(TEST_OPTS)))
 $(error Use TEST_TIMEOUT to override this option)
 endif
 
+VMCI_SYSTEMD_INIT_IMAGE ?= docker.io/library/vmci-systemd-init:latest
+
+# Always attempts a fresh build rather than checking for an existing image first, so a
+# Dockerfile change can never be shadowed by a stale image the way a stale toolbox binary
+# once was (see vcsim spec D-97-20). Never fails the build: TestVMCI_SystemdInit_GuestInfoRoundTrip
+# and TestVMCI_NestedContainers_GuestInfoRoundTrip skip themselves (t.Skipf) when the image is
+# absent, so a system without docker/podman, or any build failure (no network, etc.), must not
+# break `make test`.
+.PHONY: vmci-systemd-init-image
+vmci-systemd-init-image: ## Build the vmci-systemd-init image for simulator VMCI/nestedContainers tests
+	-@command -v docker >/dev/null 2>&1 && \
+		docker build -t $(VMCI_SYSTEMD_INIT_IMAGE) ./simulator/testdata/vmci-systemd-init/ || \
+		echo "vmci-systemd-init-image: skipping (docker/podman unavailable, or build failed); dependent simulator tests will skip themselves"
+
 .PHONY: go-test
-go-test: ## Runs go unit tests with race detector enabled
+go-test: vmci-systemd-init-image ## Runs go unit tests with race detector enabled
 	GORACE=$(GORACE) CGO_ENABLED=1 $(GO) test \
   -count $(TEST_COUNT) \
   -race \
--- a/simulator/container.go
+++ b/simulator/container.go
@@ -29,6 +29,28 @@ var (
 	eventWatch eventWatcher
 )
 
+// shellQuote wraps s in POSIX single quotes, escaping any embedded single
+// quotes with the '\” sequence.  Use this before appending container image
+// names or command args to the bash -c string in create().
+func shellQuote(s string) string {
+	return "'" + strings.ReplaceAll(s, "'", `'\''`) + "'"
+}
+
+// filterSeccompUnconfined removes the "--security-opt seccomp=unconfined" pair
+// from run (in-place, preserving order).  Used when a custom seccomp profile
+// replaces the unconfined setting added by nestedContainers mode.
+func filterSeccompUnconfined(run []string) []string {
+	out := run[:0]
+	for i := 0; i < len(run); i++ {
+		if run[i] == "--security-opt" && i+1 < len(run) && run[i+1] == "seccomp=unconfined" {
+			i++ // skip both tokens
+			continue
+		}
+		out = append(out, run[i])
+	}
+	return out
+}
+
 const (
 	deleteWithContainer = "lifecycle=container"
 	createdByVcsim      = "createdBy=vcsim"
@@ -40,12 +62,22 @@ func init() {
 	}
 }
 
+// exitErrStderr returns the trimmed stderr captured by an *exec.ExitError, or
+// "" if err is not an *exec.ExitError (or captured no stderr). Shared by
+// every call site in this file that needs to inspect a failed command's
+// stderr, whether to format an error (commandError) or to decide on a
+// specific failure reason (e.g. "already exists", "already in use").
+func exitErrStderr(err error) string {
+	var exitErr *exec.ExitError
+	if !errors.As(err, &exitErr) {
+		return ""
+	}
+	return strings.TrimSpace(string(exitErr.Stderr))
+}
+
 // commandError logs and returns an error from a command execution failure, including stderr if available.
 func commandError(operation string, args []string, err error) error {
-	stderr := ""
-	if xerr, ok := err.(*exec.ExitError); ok {
-		stderr = strings.TrimSpace(string(xerr.Stderr))
-	}
+	stderr := exitErrStderr(err)
 	var cmdErr error
 	if stderr != "" {
 		cmdErr = fmt.Errorf("%s %v failed: %s: %s", operation, args, err, stderr)
@@ -220,9 +252,17 @@ func createVolume(volumeName string, labels []string, files []tarEntry) (uid str
 		cmd := exec.Command("docker", run...)
 		out, err := cmd.Output()
 		if err != nil {
-			return "", commandError("volume create", cmd.Args, err)
+			// Podman (unlike Docker) rejects volume create when the volume already
+			// exists; Docker is idempotent here.  Named volumes can be implicitly
+			// created by "docker create -v name:/path" before we reach this explicit
+			// create, so treat "already exists" as success and continue to populate.
+			if !strings.Contains(exitErrStderr(err), "already exists") {
+				return "", commandError("volume create", cmd.Args, err)
+			}
+			uid = name // volume already exists; provided name is the uid
+		} else {
+			uid = strings.TrimSpace(string(out))
 		}
-		uid = strings.TrimSpace(string(out))
 
 		if name == "" {
 			name = uid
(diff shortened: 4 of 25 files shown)


hickeng:
I ran a structured self-review against this change before opening it, distilling a standard from
what the code says it's for (not just what it does) and then checking the code against that
standard. Posting it here since it's more useful attached to the PR than folded into the
description. Findings are ordered by severity; conformance passes are included too, since a review
that only lists violations doesn't show its own standard.

## Standard used

Distilled from the change itself and its own doc comments:

- Guest→host `guestinfo` read/write over a transport available in an unprivileged container.
- Writes are reflected in the VM's `ExtraConfig` and generate property-change notifications.
- Unmodified guest consumers work against it: cloud-init's `DataSourceVMware` prefix/exit contract,
  `vmtoolsd --cmd` semantics.
- Transport selection degrades predictably and never silently targets the wrong endpoint.
- Per-VM isolation; concurrent guest RPCs are safe against the simulator's object model.
- Sockets, goroutines, and temp directories are released on stop, remove, and abnormal exit.
- No silent dependence on the host toolchain or architecture — works when `govmomi` is consumed as
  a module dependency, not just built in-tree.
- Guest-controlled input can't alter host-side simulator behaviour beyond guestinfo semantics.
- Test-only artifacts are distinguishable from production binaries.

## Findings

### 1. Daemon mode picks AF_VSOCK over the working Unix socket, on a mistaken premise

`toolbox/toolbox/main.go`'s `selectChannels()` deliberately excludes `$VMX_RPC_SOCK`, with the
comment: *"the Unix socket server closes after each request, so a persistent daemon channel would
break on the second exchange."*

That isn't what the server does. `simulator/guestrpc_server.go`'s `serveConnLEWithConn` loops until
a read error and never closes per request; `serveConnDataMap` only returns early when the *client*
sets `fastClose`. So daemon mode falls through to `IsVsockAvailable()` → `VsockChannel` →
`connect(cid=2, port=976)` — the real hypervisor CID, not vcsim. I confirmed `socket(AF_VSOCK, ...)`
succeeds on a plain Linux host with no vsock module loaded, so this isn't gated by an environment
check that would catch it.

Suggest: prefer the Unix socket in `selectChannels`, mirroring the order `openRPCChannel` already
uses for the one-shot path.

### 2. The toolbox-artifact build can't work when govmomi is a module dependency, and hardcodes amd64

`simulator/vmci_artifacts_linux.go` resolves its output directory via `runtime.Caller(0)` to
`simulator/bin/`. The `os.MkdirTemp` fallback only fires when `runtime.Caller` returns `!ok` — not
when the subsequent `os.MkdirAll` fails. In the read-only Go module cache, `MkdirAll` always fails,
so the build error is set, the returned path is empty, and `/usr/bin/vmware-rpctool` injection is
silently skipped. That's the common case for anyone using vcsim from `go get`/`go mod` rather than a
checkout of this repo.

Separately, the build hardcodes `GOOS=linux GOARCH=amd64`; on an arm64 host the injected binary
fails at container exec time rather than at build time.

Suggest: build into a temp or user cache directory instead of a path relative to the source file,
and use `runtime.GOARCH`/the container's target platform instead of a hardcoded arch.

### 3. `info-set` accepts any `ExtraConfig` key, including `RUN.*`

`GuestRPCServer.infoSet` writes any key into the VM's `ExtraConfig`. Real VMX restricts guest
`info-set` to the `guestinfo.*` namespace. The socket directory here is bind-mounted **read-write**
into a container that may run an arbitrary image, so guest code can set `RUN.volume.*`, `RUN.vmci`,
`RUN.nestedContainers` — keys vcsim itself interprets as host-side container directives on the next
start. That's both a fidelity gap against real VMX and a guest-influences-host-config path that's
probably not intended for anything beyond a fully-trusted test image.

Suggest: restrict `info-set` to the prefixes real VMX permits.

### 4. `guestRPC` can leak on the `stop()` error path

`stop()` returns early when `svm.c.stop(ctx)` errors, before reaching `svm.guestRPC.Stop()`.
`remove()` stops the GuestRPC server *before* `c.remove()`. A container-stop failure leaves the
listener, its goroutines, and the socket alive, and the two teardown paths are asymmetric with no
stated reason for the difference.

Suggest: stop the GuestRPC server unconditionally (e.g. via `defer`) and make both paths consistent.

### 5. The stale-socket sweep in `TestMain` doesn't match anything

`main_test.go`'s `TestMain` globs `filepath.Join(os.TempDir(), "vcsim-rpc-*.sock")`. The actual path
is `os.TempDir()/vcsim-rpc-<uid>/rpc.sock` — the socket lives one directory level deeper than the
glob checks, and Go's `filepath.Match` doesn't let `*` cross a path separator (I verified this
against both the socket path and its parent directory). The sweep is currently dead code, and
leftover `vcsim-rpc-*` directories accumulate under `TempDir` across crashed test runs.

Suggest: glob `vcsim-rpc-*` (the directories) and `os.RemoveAll` each match.

### 6. A panic in `infoGet`/`infoSet` would leave the VM's object lock held forever

`Registry.WithLock` calls `f()` without a `defer` around the unlock, so a panic inside `f()` skips
releasing the lock. `ObjectLock.Release` only clears `heldBy` when its internal refcount reaches
zero, so a panic here leaves it permanently non-zero. `GuestRPCServer.serveConn` recovers any panic
from its dispatch (reasonably, to keep one bad connection from taking down the server), which means
a panic in `infoGet`/`infoSet` — which do real work under this lock, `Checkpoint`/`PropertyDiff`
reflection over guest-supplied key names — would silently wedge every future operation on that VM
from any other caller, with no error surfaced anywhere.

The missing `defer` in `WithLock` predates this change; this is the first caller I can see that
drives it from network input reachable from inside a container, with a `recover()` sitting right
above it that would hide the resulting wedge rather than surface it.

Suggest: add the `defer` in `WithLock` (or acquire/release with `defer` locally in this call site),
and consider not swallowing the panic silently — even a log line naming the VM would turn a silent
hang into a diagnosable one.

### 7. The GuestRPC artifact is auto-injected under the real binary's name, with no way to detect it

`VmciToolboxBinaryPath()` builds and injects the `toolbox` binary as `/usr/bin/vmware-rpctool` — the
same name and path a real installation would use. There's currently no way for downstream tooling
(an image-build pipeline, a compliance scan) to tell this apart from the genuine binary by name,
version string, or embedded marker.

Suggest: give the built artifact something identifying — a build tag baked into a `--version`
output, an embedded string, anything a caller who cares can grep or exec-and-check for. Costs
little and makes "don't ship this in a production image" checkable rather than aspirational.

### 8. `go build` is skipped whenever a cached binary exists, and the comment overstates what that buys

`vmci_artifacts_linux.go` skips `go build` entirely once a cached binary exists at the target path.
The comment says this "benefit[s] from Go's incremental build," but the build isn't run at all in
that case — a source change to the toolbox package would silently keep using a stale cached binary
until `simulator/bin/` is deleted by hand. `sync.Once` also means a transient build failure is
cached as a permanent one for the life of the process.

Suggest: always invoke `go build` (Go's own build cache already makes a no-op rebuild cheap) rather
than gating on `os.Stat`.

### 9. The seccomp-profile plumbing in `container.go` is currently unreachable

`SeccompProfile` is hardcoded to `""` at the one call site (`container_virtual_machine.go`), so the
`if opts.SeccompProfile != ""` branch in `container.go`, and `filterSeccompUnconfined`, are dead
code as of this change. Not a functional problem, but worth either wiring up or removing so the
surface area matches what's actually exercised.

### 10. `IsVsockAvailable`'s doc comment describes a mechanism that isn't present

The comment says "the seccomp intercept handles the `socket()` call, so this returns true inside
containers with `RUN.vmci=true`." There's no seccomp intercept in this change — `IsVsockAvailable`
just calls `socket(AF_VSOCK, ...)` directly, which is exactly why finding 1 above is possible: the
comment reads as if a real interception layer is protecting daemon-mode selection, and it isn't.

### 11. Unix socket path length

`sun_path` is limited to roughly 108 bytes on Linux. `os.TempDir()` + `vcsim-rpc-` + a 36-character
UUID + `/rpc.sock` fits comfortably under `/tmp`, but a long `$TMPDIR` (not unusual in CI) could
push it over the limit into an opaque `bind` failure rather than a clear error.

### 12. Nits

`.gitignore` is missing a trailing newline.

## What already holds up

- **The `rpcMu` mutex in `GuestRPCServer` is correct, and the reasoning in its comment checks out.**
  I traced this against `simulator/internal/object_lock.go`: `ObjectLock` is reference-counted
  keyed on the caller's `*Context` pointer, so two goroutines sharing the same `*Context` both
  satisfy the re-entry check and proceed without taking the underlying mutex. Without the explicit
  `rpcMu`, two concurrent `info-set` calls from the same server would genuinely race on
  `ExtraConfig`. The refcount bookkeeping is itself protected, so it can't release early either.
  Subtle, and correctly reasoned about.
- **Protocol auto-detection between the two wire framings is sound** for realistic payloads: the
  big-endian/little-endian disambiguation against the size threshold holds, a zero-length LE frame
  correctly falls into the LE branch, and a malformed DataMap packet always consumes the declared
  length so the parser can't spin.
- **Unmodified cloud-init's transport-selection logic is correctly reverse-engineered**, not just
  assumed: the reasoning that `is_container()` always makes `is_vmware_platform()` return false in a
  container (so the envvar seed is the only transport cloud-init will use here) is exactly why the
  approach works without patching cloud-init.
- **Toolchain build failure is non-fatal and logged** — the container still starts even if the
  `vmware-rpctool` build fails.
- **Build-tag pairs (`_linux.go`/`_other.go`) keep an identical exported surface and fail closed**
  with clear errors on unsupported platforms rather than silently no-opping.

## A cheaper alternative worth considering

The runtime `go build` in `vmci_artifacts_linux.go` exists to get one binary into the container.
Producing it as a build/test fixture instead — a Makefile target, a `go:embed`ded prebuilt static
binary, or a step in `TestMain` — would remove findings 2 and 8 outright, drop the runtime
dependency on a Go toolchain being present, and make giving the artifact an identifying marker
(finding 7) trivial to add at build time instead of after the fact.
</content>

PR #4094, opened 2026-08-19
refactor: use `slices.Contains` for membership checks

tenthirtyam:
## Description

Replace manual loops that searched string slices with `slices.Contains`. This simplifies the code, improves readability, and standardizes slice membership checks across these components.


dougm (approved):
Nice cleanups!

PR #4096, opened 2026-08-19
refactor: use `WaitGroup.Go`

tenthirtyam:
## Description

In Go 1.25 and later, `WaitGroup` goroutine pattern can use `WaitGroup.Go`. The `(*sync.WaitGroup).Go(func())` combines these steps in one call. This keeps the goroutine launch and the `WaitGroup `accounting together and reduces boilerplate code.


dougm (approved):
Good stuff

PR #4099, opened 2026-08-19
refactor: remove redundant else branches

tenthirtyam:
## Description

Simplified control flow by removing `else` blocks that followed early `return`/halt paths. This aligns with idiomatic Go style, reduces nesting, and keeps behavior unchanged.


dougm (approved):
lgtm

PR #4102, opened 2026-08-19
refactor: replace conditional assignment with `max` or `min`

tenthirtyam:
## Description

Using `max` or `min` makes the intent explicit and reduces branching code.

### DIFF
--- a/simulator/model.go
+++ b/simulator/model.go
@@ -978,10 +978,7 @@ func RunContainer(ctx context.Context, c *vim25.Client, vm mo.Reference, args st
 
 // delay sleeps according to DelayConfig. If no delay specified, returns immediately.
 func (dc *DelayConfig) delay(method string) {
-	d := 0
-	if dc.Delay > 0 {
-		d = dc.Delay
-	}
+	d := max(dc.Delay, 0)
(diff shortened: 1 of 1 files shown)


vim25/json/decode.go
```diff
@@ -546,10 +546,7 @@ func (d *decodeState) array(v reflect.Value) error {
 		if v.Kind() == reflect.Slice {
 			// Grow slice if necessary
 			if i >= v.Cap() {
-				newcap := v.Cap() + v.Cap()/2
```
dougm, line 549: Keep in mind that `vim25/json` is forked from upstream golang, as is `vim25/xml`. In general, it's probably better to leave these packages as-is wrt style/lint, as it will avoid potential future conflicts when we rebase. I didn't go through all this history, but for example this hunk of the code doesn't exist on [main](https://github.com/golang/go/blob/master/src/encoding/json/decode.go#L508) afaict.

tenthirtyam: Updated to exclude `vim25/json` for the time being.

https://github.com/vmware/govmomi/commit/e82eb81d32eb56fa05c24971218fc394567cce7d

dougm (approved):

PR #4109, opened 2026-08-20
simulator: fix PropertyCollector/ListView deadlocks under concurrency

slimm609:
## Description
lock-ordering bug deadlock the simulator's PropertyCollector under concurrent load (e.g. a CSI driver watching CNS tasks through a ListView while another goroutine mutates it):

1 A PropertyFilter that traverses a ListView acquires two object locks in opposite orders on two code paths: * WaitForUpdates{,Ex} -> PropertyFilter.update takes the filter lock, then PropertyFilter.collect traverses the ListView and takes the ListView lock (filter -> ListView). * ModifyListView takes the ListView lock at dispatch, then Registry.Update -> applyHandlers -> PropertyFilter.UpdateObject takes the filter lock (ListView -> filter). UpdateObject only needs to flag that the filter must re-sync; it does not need the filter object lock to do so, and taking it there is what creates the second lock-order. Make PropertyFilter.sync an atomic.Bool: UpdateObject stores it lock-free, and update swaps it before taking the filter lock for the refs reset + collect. That removes the ListView -> filter edge, leaving the single filter -> ListView order everywhere.

## How Has This Been Tested?

observed as ~100 goroutines blocked on a single ObjectLock, with task completions never delivered to the watcher.

Added test coverage: 
Adds TestWaitForUpdatesConcurrentCancel and
TestListViewModifyWaitForUpdatesLockOrder, both of which deadlock without these fixes.

tested in fork branch for several weeks with hundreds of iterations.

## Guidelines

### DIFF
--- a/simulator/simulator.go
+++ b/simulator/simulator.go
@@ -266,9 +266,21 @@ func (s *Service) call(ctx *Context, method *Method) soap.HasFault {
 		args = append(args, reflect.ValueOf(ctx))
 	}
 	args = append(args, reflect.ValueOf(method.Body))
-	ctx.Map.WithLock(ctx, handler, func() {
+
+	// WaitForUpdates{,Ex} block until an update or cancel, and
+	// CancelWaitForUpdates must interrupt such a blocked call. They
+	// synchronize internally via pc.mu and lock each collected object as
+	// needed, so they must not be serialized under the per-object dispatch
+	// lock: holding it across a blocking wait deadlocks the cancel (a method
+	// on the same PropertyCollector) against the waiter.
+	switch method.Name {
+	case "WaitForUpdates", "WaitForUpdatesEx", "CancelWaitForUpdates":
 		res = m.Call(args)
-	})
+	default:
+		ctx.Map.WithLock(ctx, handler, func() {
+			res = m.Call(args)
+		})
+	}
 
 	return res[0].Interface().(soap.HasFault)
 }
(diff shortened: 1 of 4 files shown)


simulator/simulator.go
```diff
@@ -266,9 +266,21 @@ func (s *Service) call(ctx *Context, method *Method) soap.HasFault {
...
 	args = append(args, reflect.ValueOf(method.Body))
-	ctx.Map.WithLock(ctx, handler, func() {
+
+	// WaitForUpdates{,Ex} block until an update or cancel, and
+	// CancelWaitForUpdates must interrupt such a blocked call. They
+	// synchronize internally via pc.mu and lock each collected object as
+	// needed, so they must not be serialized under the per-object dispatch
+	// lock: holding it across a blocking wait deadlocks the cancel (a method
+	// on the same PropertyCollector) against the waiter.
+	switch method.Name {
```
dougm: We shouldn't need this, as PropertyCollector already opts-out of the default per-object lock via: https://github.com/vmware/govmomi/blob/80c9cf53576b492b1cd1025a44483ef588040f57/simulator/property_collector.go#L31
https://github.com/vmware/govmomi/blob/80c9cf53576b492b1cd1025a44483ef588040f57/simulator/registry.go#L718
Unless there is a problem with this path, but I've not dug into it yet: https://github.com/vmware/govmomi/blob/80c9cf53576b492b1cd1025a44483ef588040f57/simulator/registry.go#L672-L676

slimm609: Thanks Doug,  Was able to confirm this change wasn't needed.

dougm (approved):
Thanks @slimm609 !

PR #4113, opened 2026-08-21
vcsim: guard SearchIndex.FindAllByIp against non-entity objects

dbhat-arkin:
## Description

FindAllByIp iterated every object in the registry and called
    asHostSystemMO/asVirtualMachineMO on each one without checking its
    reference type first. The registry also holds internal helper objects
    (ManagedMethodExecuter, DynamicTypeManager) that don't embed an mo type,
    lazily created the first time anything does an esxcli passthrough call
    against a host - getManagedObject panics on those rather than returning
    false, crashing the request. Guard with the same ref.Type check already
    used by Registry.All/AllReference before attempting the conversion.


dougm (approved):
Thanks @dbhat-arkin

PR #4114, opened 2026-08-21
vcsim: fix Datastore.path() fallback when only Info.Url is set and in load path

dbhat-arkin:
## Description

**Recorded inventory**
    vcsim: preserve a saved datastore Url on the -load path

    Same bug class as e56e8606, different trigger: Datastore.model() runs
    when loading a saved inventory (vcsim -load, typically produced by
    `govc object.save` against a real vCenter) and unconditionally rewrote
    Info.Url/Summary.Url to vcsim's own local temp dir, discarding whatever
    URL the datastore actually had -- including a real vCenter's genuine
    ds:///vmfs/volumes/<uuid>/ URL, which record/replay testing against a
    real vCenter's captured inventory would hit immediately.

    Fix: reuse the localPath/path() mechanism from e56e8606. For a loaded
    datastore whose saved Url already has the real-vCenter ds:// shape,
    back it locally with a fresh temp dir (ds.localPath) but leave the
    original, already-correct Url untouched instead of overwriting it. For
    a loaded datastore that was itself a raw local path (e.g. re-loading a
    previously-saved vcsim instance), synthesize a ds:///vmfs/volumes/<uuid>/
    value same as CreateLocalDatastore, since there's no real URL to
    preserve in that case.

    Verified: saved a running vcsim instance's own inventory (govc
    object.save), loaded it into a fresh instance -- datastore Url reports
    ds:///vmfs/volumes/<uuid>/ (not a raw temp path), and datastore.ls/
    datastore.info resolve without error against the freshly-created real
    backing directory. Full simulator test suite passes, including
    TestModelLoadAlignCounter.

**New Inventory**

    vcsim: report a real vCenter-shaped Url for local datastores

    CreateLocalDatastore set Info.Url/Summary.Url to vcsim's own raw backing
    directory (e.g. /tmp/govcsim-<pid>/DC0-LocalDS_0), instead of the
    ds:///vmfs/volumes/<uuid>/ style URL a real vCenter local/VMFS datastore
    reports. At least one real collector's SOAP client parses this URL and
    throws (IllegalArgumentException: Datastore url format is not correct)
    when it doesn't contain /vmfs/volumes, which silently kills that
    datastore's entire config payload every poll cycle -- and cascades
    downstream into the platform never resolving VM-to-datastore link
    objects for any datastore-attributed metric.

    Fix: keep the real directory for vcsim's own internal file resolution
    (new Datastore.localPath field, used via a new path() helper in place of
    Summary.Url in resolve()/RefreshDatastore()/ServeDatastore()), and only
    overwrite the externally-visible Info.Url/Summary.Url with a synthesized
    ds:///vmfs/volumes/<uuid>/ value after CreateLocalDatastore's add() (which
    needs the real path to os.Stat) succeeds. No other datastore type
    (NAS/vSAN/loaded) is affected -- they leave localPath unset and continue
    using Summary.Url directly, unchanged.

    Verified: local vcsim instance now reports Info.Url/Summary.Url as
    ds:///vmfs/volumes/<uuid>/; VM creation/vmx placement and datastore.ls
    still correctly resolve against the real backing directory; full
    simulator test suite passes.

### DIFF
--- a/govc/test/import.bats
+++ b/govc/test/import.bats
@@ -15,9 +15,10 @@ load test_helper
   run govc vm.destroy "$TTYLINUX_NAME"
   assert_success
 
-  # link ovf/ova to datastore so we can test with an http source
-  dir=$(govc datastore.info -json | jq -r .datastores[].info.url)
-  ln -s "$GOVC_IMAGES/$TTYLINUX_NAME"* "$dir"
+  # upload ovf/ova to datastore so we can test with an http source
+  for f in "$GOVC_IMAGES/$TTYLINUX_NAME"*; do
+    govc datastore.upload "$f" "$(basename "$f")"
+  done
 
   run govc import.spec "https://$(govc env GOVC_URL)/folder/$TTYLINUX_NAME.ovf"
   assert_success
--- a/govc/test/library.bats
+++ b/govc/test/library.bats
@@ -213,14 +213,16 @@ load test_helper
 
 @test "library.deploy" {
   vcsim_env
+  unset GOVC_HOST # else datastore.download tries via ESX
 
   run govc library.create my-content
   assert_success
   library_id="$output"
 
-  # link ovf/ova to datastore so we can test library.import with an http source
-  dir=$(govc datastore.info -json | jq -r .datastores[].info.url)
-  ln -s "$GOVC_IMAGES/$TTYLINUX_NAME."* "$dir"
+  # upload ovf/ova to datastore so we can test library.import with an http source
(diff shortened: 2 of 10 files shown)


dougm (commented):
Looks like this change is causing unit test failures and govc test failures, I've not looked close yet

dbhat-arkin:
> Looks like this change is causing unit test failures and govc test failures, I've not looked close yet

Thanks @dougm  for the review. I have fixed the test failures in the latest commit.  Kindly review and let me know the feedback / suggestions.

dougm (approved):
lgtm, thanks @dbhat-arkin !

PR #4115, opened 2026-08-21
vcsim: add -host-ip-base flag for deterministic host IPs

dbhat-arkin:
## Description

Implemented a flag / argument / switch while will enable the new inventory creation will have the IPs assigned to hosts .

    Model.HostIPBase (vcsim flag -host-ip-base) assigns each created
    HostSystem a deterministic management vnic IP, incrementing across both
    standalone and cluster host creation, instead of the template default
    (127.0.0.1) every host otherwise shares. Since these hosts go through
    the model's own addHost path, they also get real DVS membership and
    datastore mounting automatically - no separate post-startup scripting
    needed for any of that.

Example usage:
vcsim -host 4 -host-ip-base 10.20.8.0
this command will bring up 4 hosts and the hosts will have IPs 10.20.8.0 to 10.20.8.3 (sequential assignment). 

### DIFF
--- a/simulator/model.go
+++ b/simulator/model.go
@@ -7,9 +7,11 @@ package simulator
 import (
 	"context"
 	"crypto/tls"
+	"encoding/binary"
 	"fmt"
 	"log"
 	"math/rand"
+	"net"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


dougm (approved):
Cool, lgtm

PR #4119, opened 2026-08-27
Fix: retry vCenter's "Too many outstanding operations" fault

sh013053:
## Problem

`govc object.save` (and other long-running govc commands) can abort partway
through when working against large vCenter inventories:

Saving HostCertificateManager:ha-certificate-manager-29...ok
Saving HostVirtualNicManager:virtualNicManager-29...ok
Saving HostNetworkSystem:networkSystem-29...ok
govc: ServerFaultCode: A general system error occurred: Too many outstanding operations

This is vCenter's own `SystemError` fault, returned when its internal
operation queue is briefly saturated — common when traversing inventories
with thousands of objects (e.g. ~5K VMs). It's transient and normally
clears within a few seconds, but govc had no retry for it, so a single
blip anywhere in a multi-hour dump fails the entire run with no resume.

## Fix

- Add `vim25.RetryTooManyOutstandingOperations(delay)`, a `RetryFunc` that
  detects this specific fault (via the existing `fault.As` helper, matching
  on the `SystemError.Reason` text) and retries after `delay`.
- Wire it into govc's default `RoundTripper` (`cli/flags/client.go`)
  alongside the existing temporary-network-error retry: up to 5 attempts,
  5s apart (max ~20s additional wait), applied to every govc command.

## Test plan

- Added `TestRetryTooManyOutstandingOperations` in `vim25/retry_test.go`
  covering: recovery after a transient fault, giving up after exhausting
  attempts, and not retrying unrelated `SystemError` faults.
- `go build ./...`
- `go test ./vim25/... ./cli/flags/... ./fault/...`

### DIFF
--- a/cli/flags/client.go
+++ b/cli/flags/client.go
@@ -291,6 +291,11 @@ func (flag *ClientFlag) RoundTripper(c *soap.Client) soap.RoundTripper {
 	// This means a maximum of 3 attempts.
 	rt := vim25.Retry(c, vim25.RetryTemporaryNetworkError, 3)
 
+	// Retry when vCenter's operation queue is briefly saturated, e.g. when
+	// traversing large inventories. Backs off 5s between attempts, up to 5
+	// attempts (20s max additional wait).
+	rt = vim25.Retry(rt, vim25.RetryTooManyOutstandingOperations(5*time.Second), 5)
+
 	switch {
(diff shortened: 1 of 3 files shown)


dougm (approved):
Nice, thanks @sh013053 !

PR #4120, opened 2026-08-28
Fix/dvs concrete type

dbhat-arkin:
## Description

vcsim always registers its DVS as the abstract `DistributedVirtualSwitch` type. A real vCenter never does this — it always reports the concrete `VmwareDistributedVirtualSwitch` subtype. Any client that keys off an object's own reported type (rather than querying generically by the abstract type) never sees vcsim's DVS at all. That includes real collectors like vRNI, and it also happened to mask two existing bugs in govmomi's own client code — `object.ReferenceFromString`'s `dvs-NNN` shorthand and `govc find -type w` — which referenced the wrong (abstract) type name and so looked like they worked against vcsim while actually being broken against any real vCenter.

Fixing the type rename surfaced a second, related bug: vcsim never populated `DVSConfigInfo.ProductInfo`. That field is required and always sent by a real vCenter — its own API docs say the server fills it in if the client doesn't — but vcsim left it as an empty value whose `Vendor` sub-field gets dropped from the wire format entirely when unset. At least one real collector (vRNI) reads `config.productInfo.getVendor()` unconditionally and throws a `NullPointerException` on the resulting null, which silently dropped that DVS's config from every single poll cycle.

**This change:**
- Renames vcsim's DVS type to `VmwareDistributedVirtualSwitch`, matching real vCenter, while keeping property queries by the abstract type name working (real clients can still query either way).
- Fixes the two govmomi client bugs the rename uncovered, so both now resolve to the concrete type as they should against a real vCenter too.
- Defaults `DVSConfigInfo.ProductInfo.Vendor` (and the equivalent `DVSSummary.ProductInfo`) whenever the client leaves it empty — which is what `govc dvs.create` does by default — so a vcsim-created DVS always reports a valid vendor.

## How Has This Been Tested?

Added `TestDVSConcreteType` (checks the MOR type, and that queries by both the concrete and abstract type name still resolve it), `TestDVSDefaultProductInfo` (checks the default, an explicitly-supplied `ProductInfo` is preserved, and the empty-but-non-nil case `govc dvs.create` sends also defaults), `TestKindsAliasVmwareDVS`, and a new case in `TestReferenceFromString`. Full `simulator`, `object`, `cli/object`, and `govc` test suites pass.

Also validated end-to-end: built a small multi-cluster, multi-DVS vcsim topology, connected it to a real vRNI collector, and confirmed the DVS and its portgroups now show up correctly in vRNI's UI with no more `NullPointerException` in the collector logs — previously the DVS was invisible to vRNI entirely and its portgroups showed no DVS association.

### DIFF
--- a/cli/object/find.go
+++ b/cli/object/find.go
@@ -54,7 +54,7 @@ var alias = []struct {
 	{"p", "ResourcePool"},
 	{"r", "ComputeResource"},
 	{"s", "Datastore"},
-	{"w", "DistributedVirtualSwitch"},
+	{"w", "VmwareDistributedVirtualSwitch"},
 }
 
 func aliasHelp() string {
--- a/cli/object/find_test.go
+++ b/cli/object/find_test.go
@@ -1,0 +1,25 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package object
+
+import "testing"
+
+// TestKindsAliasVmwareDVS guards against "-type w" mapping back to the
+// abstract DistributedVirtualSwitch. That name resolves fine against a
+// generic hierarchy-aware type filter, but "govc find" applies it as an
+// exact string match against each traversed object's own reported type
+// (kinds.wanted), which is always the concrete VmwareDistributedVirtualSwitch
+// on a real vCenter -- so the abstract name here silently matched nothing.
+// It went unnoticed because vcsim used to (incorrectly) register its own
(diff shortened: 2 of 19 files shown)


tenthirtyam (changes requested):
CI is failing:

```
 Examples:
   govc find
@@ -4616,7 +4616,7 @@ The '-t' flag value can be a managed entity type or one of the following aliases
   p    ResourcePool
   r    ComputeResource
   s    Datastore
-  w    DistributedVirtualSwitch
+  w    VmwareDistributedVirtualSwitch
 
 Examples:
   govc ls -l '*'
Error: Run make doc to update USAGE.md
Error: Process completed with exit code 1.
```

PR #4121, opened 2026-08-28
vcsim: support structured guest IP config, gateway and DNS

dbhat-arkin:
## Description

vcsim's `SET.guest.ipAddress` backdoor only ever wrote to the legacy `GuestNicInfo.IpAddress` field, as a bare IP with no subnet info. Real API consumers — vRNI's collector, for one — don't read that legacy field for "IP Address" / "Network Address"; they read the structured `GuestNicInfo.IpConfig` field instead, which needs a prefix length that a bare IP can't carry. There was also no way at all to give a simulated VM a default gateway or DNS server. Net effect: anyone using vcsim to scale-test IP/network discovery had no way to make simulated VMs show up correctly in a real vCenter API consumer.

This change:
- Lets `SET.guest.ipAddress` accept CIDR notation (`10.0.0.42/24`), and populates `GuestNicInfo.IpConfig` (address, prefix length, origin, state) alongside the legacy field. A bare IP with no `/` still behaves exactly as before, so this isn't a breaking change.
- Adds two new backdoor keys, `SET.guest.defaultGateway` and `SET.guest.dnsServer`, which populate `GuestInfo.IpStack`'s route table and DNS config — the only place a real vCenter's API reports this data.
- Makes sure setting gateway and DNS in the same reconfigure call doesn't clobber each other, since they share one `GuestStackInfo` entry.

**Why this change?**
The application  where the vcsim was added as datasource, wasn't getting the IPs of the VM inventories in the vcsim.

## How Has This Been Tested?

Added a unit test covering CIDR parsing, `IpConfig` population, the default route, comma-separated DNS servers, and the gateway/DNS non-clobbering case, alongside the existing `TestApplyExtraConfig` test.

Also validated end-to-end against a live vcsim instance wired up to a real vRNI (VMware Aria Operations for Networks) collector: assigned IPs, gateways, and DNS servers to simulated VMs and confirmed vRNI's UI correctly showed IP Address, Network Address, and Default Gateway — fields that were previously blank.

## Guidelines

### DIFF
--- a/simulator/virtual_machine.go
+++ b/simulator/virtual_machine.go
@@ -542,6 +542,19 @@ func extraConfigKey(key string) string {
 	return key
 }
 
+// parseGuestIPAndPrefix accepts either a plain IP ("10.0.0.42") or CIDR
+// notation ("10.0.0.42/24") for the SET.guest.ipAddress backdoor. A plain IP
+// keeps its prior meaning (prefix length 0) for backwards compatibility;
+// CIDR notation additionally provides the prefix length needed to compute a
+// subnet, which the legacy plain-IP form cannot express.
+func parseGuestIPAndPrefix(addr string) (ip string, prefixLen int32) {
+	if i, ipNet, err := net.ParseCIDR(addr); err == nil {
+		ones, _ := ipNet.Mask.Size()
+		return i.String(), int32(ones)
+	}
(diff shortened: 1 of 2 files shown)


dougm (approved):
lgtm, thanks @dbhat-arkin

PR #4125, opened 2026-09-04
vcsim: fix content library storage_uris to match datastore url

aruneshpa:
## Description

PR #4114 ("vcsim: report a real vCenter-shaped Url for local datastores") changed Datastore.Summary.Url/Info.Url for local datastores to a synthesized "ds:///vmfs/volumes/<uuid>/" value, matching real vCenter, and #4114's follow-up commit ("vcsim: export Datastore.Path() and fix vapi/simulator's content library path") moved vapi/simulator's internal file-IO helper, libraryPath(), off of Info.Url and onto the newly exported Datastore.Path() (vcsim's real on-disk directory) so uploads/reads kept working against the real temp dir.

But vapi/simulator's libraryItemStorageByID(), which builds the client-facing storage_uris returned by ListLibraryItemStorage and GetLibraryItemStorage, also called libraryPath() to build those URIs. Before #4114, Info.Url/Summary.Url and libraryPath() both resolved to the same raw temp directory, so this was self-consistent by accident. After #4114 they diverge: storage_uris is still the raw temp path, while Summary.Url is the synthesized ds:// value.

vapi/library/finder.PathFinder.ResolveLibraryItemStorage resolves a storage_uri to a "[datastore] path" by stripping the datastore's Summary.Url prefix. Since the prefix no longer matches, the strip is a silent no-op (strings.TrimPrefix returns its input unchanged on a mismatch) and the resolved path leaks vcsim's raw temp directory, e.g. "[LocalDS_0] tmp/govcsim-.../DC0-LocalDS_0/contentlib-.../disk.vmdk" instead of "[LocalDS_0] contentlib-.../disk.vmdk". Downstream, a consumer that stat's or copies that path against the real datastore gets a "file does not exist" error, since the leaked path doesn't exist relative to the datastore root.

Fix libraryItemStorageByID to build storage_uris from the datastore's own Summary.Url instead of libraryPath()'s real on-disk directory, matching what a real vCenter reports and what
ResolveLibraryItemStorage expects to strip. Only the URL's Path field is joined (not the raw URL string), since path.Join would otherwise collapse the "ds://" scheme's double slash.

Add a regression test that uploads a real file to a library item and round-trips it through ListLibraryItemStorage and
ResolveLibraryItemStorage, asserting the resolved path never contains vcsim's internal temp directory.

## How Has This Been Tested?

- go test ./vapi/...: passes, including the new regression test; the new test fails on main with the leaked temp path shown above.
- go test ./simulator/...: two pre-existing failures (TestFileInfo, TestCreateVm) reproduce identically on unmodified main and are unrelated to this change.
Please describe any manual tests done to verify your changes.

### DIFF
--- a/vapi/library/finder/path_test.go
+++ b/vapi/library/finder/path_test.go
@@ -7,6 +7,10 @@ package finder_test
 import (
 	"context"
 	"fmt"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -20,6 +24,7 @@ import (
 	_ "github.com/vmware/govmomi/vapi/simulator"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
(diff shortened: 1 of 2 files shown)


dougm (approved):
lgtm, thanks @aruneshpa

tenthirtyam (changes requested):
@aruneshpa can you sign this off with your broadcom.com email to pass the CLA check?

tenthirtyam (approved):

